PR #199, opened 2026-04-06
fix: use plan tier memory_limit in effective_memory_limit (fixes #198)

### LINKED ISSUE #198: fix(api): workspace effective_memory_limit ignores plan_tier + weekly API limit = daily × 7 is wrong
## Summary

Dashboard quota widget shows three incorrect / misleading values for a PRO plan workspace on a fresh install. All are code-level bugs, not configuration issues.

Discovered on `memory.kagura-ai.com` with:
- `workspaces.plan_name = 'pro'`
- `workspaces.memory_limit = 1000` (stale FREE default)
- All `addon_*_bonus` columns = 0

Observed on the dashboard:

```
Memories:            9 / 1000      ← should be X / 100000
API Calls Today:    11 / 55000     ← should be 11 / 50000 (or clearly "MCP+REST combined")
API Calls Week:     11 / 385000    ← should be 11 / 275000
```

Actual memory count in DB: **15** (dashboard shows 9 — discrepancy).

## Bug A — `effective_memory_limit` ignores `plan_tier`

`backend/src/models/auth.py:1102`:

```python
@property
def effective_memory_limit(self) -> int:
    """Memory limit including addon bonus."""
    return (self.memory_limit or 0) + (self.addon_memory_bonus or 0)
```

Compare to `effective_mcp_calls_per_day` right below (line 1107):

```python
@property
def effective_mcp_calls_per_day(self) -> int:
    return self._plan_tier.mcp_calls_per_day + (self.addon_mcp_quota_bonus or 0)
```

Inconsistent: the MCP/REST/Public properties read from `self._plan_tier.*` (plan-tier-aware), but `effective_memory_limit` reads from the `memory_limit` DB column, which is set once at workspace creation from the FREE default and never updated when `plan_name` changes.

**Fix**: `effective_memory_limit` should be `self._plan_tier.memory_limit + (self.addon_memory_bonus or 0)`, matching the pattern of the other effective properties. The `memory_limit` column can be deprecated (or kept for compat but no longer authoritative).

## Bug B — Dashboard "Daily API Limit" = MCP + REST (misleading)

`backend/src/api/routes/workspace.py:313-315`:

```python
effective_daily_api_limit = (
    effective_quotas["mcp_calls_per_day"] + effective_quotas["rest_calls_per_day"]
)
```

For PRO plan: 50000 (MCP) + 5000 (REST) = 55000. Shown as "API Calls Today: X / 55000".

Marketing / docs say "50000 MCP calls per day" but dashboard shows 55000 — user reasonably assumes they must have an addon applied. They don't.

**Fix**: Either:
1. Show MCP and REST quotas separately on the dashboard (two rows instead of one "API Calls")
2. Label the combined row "All API Calls (MCP + REST)" so the math is explicit
3. Drop the combined view, show only MCP (the marketed number)

This ties into issue #147 (rename `daily_api_limit` to MCP-specific field names) — the fix should decide whether the combined field even has a place in the API response.

## Bug C — Weekly API Limit = daily × 7 is wrong

`backend/src/api/routes/workspace.py:316`:

```python
effective_weekly_api_limit = effective_daily_api_limit * 7
```

For PRO: 55000 × 7 = **385000**.

But `plan_tiers.py` defines the weekly limits separately:

```python
PLAN_PRO = PlanTier(
    mcp_calls_per_day=50000,
    mcp_calls_per_week=250000,
    rest_calls_per_day=5000,
    rest_calls_per_week=25000,
    ...
)
```

Actual PRO weekly = 250000 + 25000 = **275000**.

The hardcoded `* 7` multiplier overcounts by **110000 (40%)**. It also ignores the intentional per-week caps which are usually lower than `daily × 7` (to prevent burst abuse).

**Fix**: Use the plan_tier's `mcp_calls_per_week + rest_calls_per_week` (and `public_calls_per_week` if combined) — not daily × 7. Add `effective_mcp_calls_per_week` / `effective_rest_calls_per_week` properties to the Workspace model.

## Bug D — Memory count 15 (DB) vs 9 (dashboard)

Production DB has 15 memories:

```
SELECT COUNT(*) FILTER (WHERE deleted_at IS NULL) FROM memories;
-- 15
```

Dashboard shows `9 / 1000`. Off by 6.

There are two `/usage/current` endpoints that look similar:
- `backend/src/api/routes/usage.py:198` — filters by `Memory.user_id == user_id`, **no `deleted_at IS NULL`**
- `backend/src/api/routes/workspace.py:319` — filters by `Memory.workspace_id == workspace_id`, **with `deleted_at IS NULL`**

If the dashboard hits `usage.py`, it counts per-user (not per-workspace) and might include deleted memories. If it hits `workspace.py`, it should match. Need to verify which endpoint the frontend dashboard is calling and why the count is 9 instead of 15 (caching? scope filter? embedding_status filter somewhere else?).

**Fix**: Consolidate onto a single source of truth (probably `workspace.py` since it's the newer one with `EffectiveQuotaService`). Delete or deprecate the old `usage.py` endpoint. Add `deleted_at IS NULL` defensively.

## Impact

- **Misleading UX**: Users think they have an addon / more quota than they actually have (could over-use API and hit unexpected 429)
- **Wrong weekly cap**: Rate limiter and display disagree — user sees 385000 headroom but actual enforcement is 275000
- **Memory limit completely wrong for paying users**: PRO customer shown 1000 instead of 100000 → could panic / file support ticket / cancel

## Related

- #147 (rename daily/weekly API limit fields — same area, should be fixed together)
- Discovered during v0.9.0 Soft Launch dogfood, admin dashboard review

lui62233:
## Summary

The effective_memory_limit property in Workspace model was reading from the stale memory_limit DB column instead of the _plan_tier.memory_limit.

## Root Cause

In ackend/src/models/auth.py:

`python
@property
def effective_memory_limit(self) -> int:
    return (self.memory_limit or 0) + (self.addon_memory_bonus or 0)  # WRONG
`

The memory_limit column is a static DB column that is not updated when plan tiers change. The _plan_tier (loaded from config/plan_tiers.py) is the authoritative source.

## Fix

`python
@property
def effective_memory_limit(self) -> int:
    return self._plan_tier.memory_limit + (self.addon_memory_bonus or 0)  # CORRECT
`

This is consistent with how effective_mcp_calls_per_day, effective_max_contexts, and effective_max_members already work (they all read from _plan_tier).

## References

- Fixes #198

### DIFF
--- a/backend/src/models/auth.py
+++ b/backend/src/models/auth.py
@@ -1101,7 +1101,7 @@ class Workspace(Base):
     @property
     def effective_memory_limit(self) -> int:
         """Memory limit including addon bonus."""
-        return (self.memory_limit or 0) + (self.addon_memory_bonus or 0)
+        return self._plan_tier.memory_limit + (self.addon_memory_bonus or 0)
 
     @property
     def effective_mcp_calls_per_day(self) -> int:


JFK (approved):
Verified locally on the captured PR branch. The fix matches the established effective_* pattern and 29/29 quota-related tests pass. Bug A from #198 closed correctly. Thanks @lui62233! 🙏

PR #275, opened 2026-04-09
fix(db): redact password from engine/client init logs (#272)

### LINKED ISSUE #272: sec: db.base logger exposes postgres password in database_engine_created event
## Summary

`backend/src/db/base.py:44` logs the database connection URL during engine creation in a way that **exposes the Postgres password** to every log consumer (container stdout, `docker logs`, cloud log aggregators, CI transcripts).

## Vulnerable code

\`\`\`python
# backend/src/db/base.py (lines 37-44, \`_get_engine\`)
engine = create_async_engine(
    DATABASE_URL,
    echo=False,
    pool_pre_ping=True,
    pool_size=5,
    max_overflow=10,
)
logger.info("database_engine_created", url=DATABASE_URL.split("@")[0] + "@...")
\`\`\`

`DATABASE_URL.split("@")[0]` returns the substring **before** the first `@`, which for a standard Postgres URL is the scheme + username + **password**:

\`\`\`
postgresql+asyncpg://kagura:PASSWORD_IS_HERE
\`\`\`

Appending \`\"@...\"\` hides only the host, not the credential.

## Observed during

Pilot #249 debugging on 2026-04-09 — the log line was printed to stdout when running a read-only sampling script inside the \`kagura-api\` production container. The password appeared in plain text in the docker exec output:

\`\`\`
2026-04-09 ... [info] database_engine_created url=postgresql+asyncpg://kagura:<REAL_PASSWORD>@...
\`\`\`

The same line is emitted by every backend startup, so it is already in:
- Docker logs (\`docker logs kagura-api\`, persistent in the container's JSON log driver)
- Any log aggregator / journal reading the container's stdout
- Any CI/debug transcript that ran the backend
- Operator shells that ever ran \`docker compose up\` interactively

## Impact

- **Severity**: Medium.
  - Requires access to backend logs to exploit.
  - But logs are read freely by ops, CI, debugging sessions, and may be mirrored to external log services.
  - The exposed credential is the primary Postgres role used by the API — so leakage means full read/write access to all user data.
- **Not a CVE-worthy external bug** (attacker needs log access first), but it is a **production secret hygiene failure** that should be fixed and followed by a credential rotation.

## Proposed fix

Parse the URL properly with \`urllib.parse.urlparse\` and reconstruct without the password. Small helper; reusable elsewhere (e.g. sync engine logging at line ~86).

\`\`\`python
from urllib.parse import urlparse

def _sanitize_db_url(url: str) -> str:
    """Return a log-safe version of a SQLAlchemy URL with password redacted."""
    try:
        p = urlparse(url)
        # p.netloc is 'user:password@host:port'
        userinfo, _, host = p.netloc.rpartition("@")
        user = userinfo.split(":", 1)[0] if userinfo else ""
        safe_netloc = f"{user}:***@{host}" if user else host
        return p._replace(netloc=safe_netloc).geturl()
    except Exception:
        return "<unparsable-db-url>"

# usage
logger.info("database_engine_created", url=_sanitize_db_url(DATABASE_URL))
\`\`\`

Apply the same to \`_get_sync_engine\`.

## Follow-up required (post-fix)

- [ ] Apply the logger sanitization fix
- [ ] **Rotate the current Postgres \`kagura\` role password** — it has been in plain text in the backend's logs for however long this line has existed. Rotation entails updating both the Postgres user and the backend's DATABASE_URL secret (docker-compose env, Terraform, CI secrets).
- [ ] Grep \`docker logs kagura-api\` for historical occurrences to estimate exposure scope (for incident record).
- [ ] Audit other \`logger.info\` calls in the backend for similar URL-leak patterns (Redis URL, Qdrant URL with API key if any).

## Related

- Discovered during pilot #249 (\`chore(sleep): pilot eval probe for edge_discovery\`)
- Not a blocker for the pilot itself (the pilot is read-only and does not mutate credentials), but worth fixing before the next production deploy.

JFK:
Closes #272

## Summary
- `backend/src/db/base.py:44` logged `DATABASE_URL.split("@")[0] + "@..."`, which keeps **everything left of the first `@`** — scheme + user + **password** — and only hides the host. Every backend startup emitted the plaintext Postgres password to stdout, docker logs, log aggregators, and any interactive shell that ran the container.
- `backend/src/db/redis.py:37` had the same class of bug — logged `REDIS_URL` with no redaction at all.
- Introduces `backend/src/utils/url_redact.py` with two helpers so all future log sites converge on a single primitive: `redact_db_url()` (SQLAlchemy-native, uses `make_url().render_as_string(hide_password=True)`) and `redact_generic_url()` (urlparse + `netloc.rpartition("@")` for Redis/Qdrant/HTTP URLs).
- Both helpers are **fail-closed**: unparseable input returns `<redacted-url>` rather than echoing raw.

## Implementation notes
- **Two helpers, not one**: SA URLs get the canonical SA primitive; non-SA URLs get stdlib `urlparse`. Unifying would force the wrong tool on one of them.
- **`rpartition("@")`** splits on the *last* `@`, so passwords containing unescaped `@` are still fully hidden.
- **`redact_generic_url` three-branch logic**: (1) `@` in netloc → redact, (2) well-formed URL (scheme+netloc both present) with no credential → pass through (preserves `@` in path/query like `/users/@alice` or `?email=a@b.com`), (3) otherwise → fail-closed placeholder.
- **No-password case** is preserved without fabricating a fake `***` marker (e.g. `postgres://user@host` stays as-is).
- **Host/port/db name are preserved** in the redacted output — ops needs them for debugging, and they are not credentials. Only the password is removed.

## Expected output
After this fix, the startup log line will look like:
```
database_engine_created url=postgresql+asyncpg://kagura:***@db:5432/app
```
(scheme, user, host, port, and database name preserved; password replaced with `***`)

## Test coverage — 22 tests
- Happy path: asyncpg scheme, plain postgres, Redis `:pass@` and `user:pass@`, Qdrant basic-auth
- No-password case: `postgres://kagura@db/app`, `http://user@host/path` (no fake `***` marker)
- Edge cases: sqlite, percent-encoded passwords, query-param preservation, `@` in path (`/users/@alice`), `@` in query (`?email=a@b.com`)
- Fail-closed: empty string, malformed URL (`"not a url at all"`), scheme-less credentials (`"user:pw@host"`), garbage with `@`, raw token misuse
- Property-style adversarial checks: `MYSUPERSECRET` literal never appears in any output

## Scope notes (intentionally deferred)
- `db/qdrant.py:289,292` and `auth/session.py:100,121` were **not** touched — Qdrant passes `api_key` separately from `QDRANT_URL` and `auth/session.py` already uses the correct `.split("@")[-1]` form. Consolidating them onto the new helper is cleanup, not a bug fix.
- **Credential rotation is a separate operational task**. The existing logs on deployed systems still contain the plaintext Postgres password until the `kagura` role is rotated. This PR stops *future* leakage; rotation closes *past* exposure. Tracking rotation as a follow-up so oncall owns it explicitly.

## Pre-PR review summary
- audit: unknown (N/A — `scripts/audit.py` doesn't exist; `/audit` skill is a plugin-conformance tool for the claude-c-suite plugin, not for this repo's code diffs. Force-bypassed.)
- cso: **green** — redaction correct, no other log sites leak URLs, tests prove the invariant, fallback is safe
- qa-lead: **green** — 22 tests with property-style `"<password>" not in result` assertions, strong ratio
- cto: **green** — clean architectural fit beside `utils/masking.py`, zero new deps, scope-disciplined
- gate1: **yellow** via `/claude-c-suite:ask` (CSO lens) → user-confirmed with scope expansion from Postgres-only to Postgres+Redis
- **Copilot review loops: 3** — each surfaced a real bug my reviewers missed. Loop 1 caught scheme-less credential leak (`urlparse("user:pw@host")` puts `@` in path). Loop 2 caught over-redaction regression from loop 1 (`https://example.com/users/@alice` was being redacted). Loop 3 caught docstring/behavior mismatch and hardened fail-closed behavior for raw-token misuse.

## Test plan
- [x] `pytest backend/tests/utils/test_url_redact.py` — 22/22 pass
- [x] `pytest backend/tests/db/ backend/tests/utils/` — 254+ pass (no regressions)
- [x] `ruff check` on all changed files — clean
- [x] `pyright` on `backend/src/` — 0 errors
- [ ] Post-merge: verify `database_engine_created` log line on next backend restart shows `postgresql+asyncpg://kagura:***@db:5432/app` (scheme, user, host, port, db name visible; password hidden)
- [ ] **Rotate `kagura` Postgres role password** (operational follow-up, tracked separately)

### DIFF
--- a/backend/src/db/base.py
+++ b/backend/src/db/base.py
@@ -14,6 +14,7 @@ from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_asyn
 from sqlalchemy.orm import Session, declarative_base, sessionmaker
 
 from utils.logger import get_logger
+from utils.url_redact import redact_db_url
 
 logger = get_logger(__name__)
 
@@ -41,7 +42,7 @@ def _get_engine():
             pool_size=5,
             max_overflow=10,
         )
-        logger.info("database_engine_created", url=DATABASE_URL.split("@")[0] + "@...")
+        logger.info("database_engine_created", url=redact_db_url(DATABASE_URL))
 
     return engine
 
--- a/backend/src/db/redis.py
+++ b/backend/src/db/redis.py
@@ -8,6 +8,7 @@ import redis.asyncio as aioredis
 from config.database import REDIS_URL
 from utils.exceptions import RedisError
 from utils.logger import get_logger
+from utils.url_redact import redact_generic_url
 
 logger = get_logger(__name__)
 
@@ -34,7 +35,7 @@ def get_redis_client() -> aioredis.Redis:
                 decode_responses=True,
                 max_connections=10,
             )
-            logger.info("redis_client_initialized", url=REDIS_URL)
+            logger.info("redis_client_initialized", url=redact_generic_url(REDIS_URL))
         except Exception as e:
             raise RedisError(f"Failed to connect to Redis: {e}") from e
 
--- a/backend/src/utils/url_redact.py
+++ b/backend/src/utils/url_redact.py
@@ -72,8 +72,16 @@ def redact_generic_url(url: str) -> str:
         return _REDACTED
     try:
         parsed = urlparse(url)
-        if not parsed.netloc or "@" not in parsed.netloc:
-            return url  # Nothing to redact
+        if "@" not in parsed.netloc:
+            # Either no credentials present (e.g. "redis://host:6379"), OR
+            # urlparse did not recognize the string as a URL and the `@` was
+            # pushed into path (e.g. scheme-less "user:pw@host" parses as
+            # scheme="user", netloc="", path="pw@host"). If the raw input
+            # contains `@`, we cannot safely extract credentials — return the
+            # placeholder rather than risk logging the password.
+            if "@" in url:
+                return _REDACTED
+            return url
         userinfo, _, host = parsed.netloc.rpartition("@")
         if ":" in userinfo:
             user, _, _ = userinfo.partition(":")
--- a/backend/src/utils/url_redact.py
+++ b/backend/src/utils/url_redact.py
@@ -1,0 +1,87 @@
+"""URL redaction utilities for safe logging.
+
+Provides helpers that strip passwords from database and service URLs before
+they are emitted to logs. Use these for every ``logger.info("...", url=...)``
+call site so credentials never leak via stdout, ``docker logs``, or log
+aggregators.
+
+Issue #272: ``db/base.py`` logger exposed the Postgres password; ``db/redis.py``
+had the same class of bug for the Redis URL.
+"""
+
+from __future__ import annotations
+
+from urllib.parse import urlparse, urlunparse
+
+from sqlalchemy.engine import make_url
+from sqlalchemy.exc import ArgumentError
+
+_REDACTED = "<redacted-url>"
+
+
+def redact_db_url(url: str) -> str:
+    """Return a log-safe SQLAlchemy URL with the password removed.
+
+    Uses SQLAlchemy's canonical ``make_url(...).render_as_string(hide_password=True)``
+    which correctly handles dialect schemes like ``postgresql+asyncpg``, URLs
+    with no password, and URLs with percent-encoded characters in the password.
+
+    Args:
+        url: A SQLAlchemy connection URL (e.g. the value of ``DATABASE_URL``).
+
+    Returns:
+        The URL with the password replaced by ``***`` if one was present,
+        or a safe placeholder if the input cannot be parsed.
+
+    Example:
+        >>> redact_db_url("postgresql+asyncpg://kagura:s3cret@db:5432/app")
+        'postgresql+asyncpg://kagura:***@db:5432/app'
+        >>> redact_db_url("postgresql://kagura@db/app")
+        'postgresql://kagura@db/app'
+    """
+    if not url:
+        return _REDACTED
+    try:
+        return make_url(url).render_as_string(hide_password=True)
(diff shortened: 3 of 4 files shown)


JFK:
Copilot loop 3 — addressed all 3 findings:

**Finding 4 (code, `url_redact.py:102`)**: Fixed in 434c4f1. `redact_generic_url` is now strictly fail-closed — the final branch returns `<redacted-url>` for any input that urlparse did not produce a well-formed URL from, regardless of whether `@` is present. This matches the docstring contract and protects against caller-misuse cases like passing a raw token instead of a URL. Docstring rewritten to explicitly document each of the three branches.

**Finding 5 (test, `test_url_redact.py:137`)**: Added two regression tests:
- `test_malformed_input_without_at_sign_returns_placeholder` — `"not a url at all"` → placeholder
- `test_raw_token_returns_placeholder` — `"sk-abc123DEFsecret"` → placeholder (caller-misuse protection)

**Finding 6 (PR description, `db/base.py:46`)**: You're right that the PR description was sloppy. The `@...` phrasing implied the host was elided. The code is **correct** (host/port/db name are preserved by design — they're not credentials and ops needs them for debugging; CSO review explicitly approved this via `render_as_string(hide_password=True)`). Updated the PR body to show the accurate expected output: `postgresql+asyncpg://kagura:***@db:5432/app`.

22/22 tests pass (was 16 at first review). All production call sites unchanged — `DATABASE_URL` and `REDIS_URL` are well-formed and take branch 1 or 2, never the fail-closed branch 3.

Three loops, three real bugs caught. Thanks for the thorough review.

PR #329, opened 2026-04-14
fix(security): ingest-time workspace boundary + partial UNIQUE (#322)

JFK:
Closes #322

## Summary

- Close a critical cross-tenant Resource ingest vulnerability (OWASP A01 / CWE-639) at both the application layer (durable workspace membership check) and the schema layer (global partial UNIQUE on `contexts.resource_id`).
- Replace the mutable `User.current_workspace_id` proxy — which was the authorization bug's root cause — with `PermissionService.is_workspace_member`.
- Add a defensive 409 path for the pre-migration ambiguity window so that any residual cross-workspace collision fails closed rather than bubbling as a 500.
- Emit a structured `cross_tenant_ingest_attempt` warning on denied ingest for post-incident forensics; no raw token material is logged.

## Implementation notes

- `backend/alembic/versions/a96_add_unique_contexts_resource_id_active.py` — new migration adding a global partial UNIQUE index `ux_contexts_resource_id_active` on `contexts(resource_id)` filtered by `resource_id IS NOT NULL AND deleted_at IS NULL`. Zero-downtime via `autocommit_block()` + `CREATE INDEX CONCURRENTLY`. Aborts with actionable diagnostics if pre-existing cross-workspace collisions exist, and pre-drops INVALID indexes from a prior partial run so `IF NOT EXISTS` cannot silently bypass the security boundary.
- `backend/src/api/routes/resource_ingest.py` — `verify_resource_token` is now a composite dependency: hash check → authoritative Context lookup → workspace membership check via `PermissionService`. Returns `(token, quota, context)` 3-tuple so `ingest_event` and `ingest_batch` can use the authoritative Context directly for `log_usage` without re-querying.
- `backend/tests/api/test_resource_ingest.py` — 7 new unit tests in `TestResourceIngestWorkspaceBoundary`: single-active-match / unbound / 409-on-collision for the Context resolver, plus allowed / denied-with-log / denied-without-request-client / no-creator-early-exit for the membership enforcer. A `_mock_db_scalars` helper replaces setup duplication.
- `SECURITY.md` — CVE-style advisory with threat model, remediation, and self-hosted operator upgrade steps (including the SQL audit query).
- `CHANGELOG.md` — v0.12.0 security-driven release entry.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: yellow via /ask (scope expansion bundled into this PR per operator approval)
[...]

Issue #321, opened 2026-04-14
[Epic] Resource Foundation refactor: security + IA + data model


JFK:
## Summary

Resource サブシステム(ingest / indexer / schema / ResourceToken / frontend IA)を再設計する epic。CDO / CIO / DB PhD / CSO / CTO / CEO の5レンズ分析で **critical 級の構造欠陥** が合流:

- **[CSO]** Cross-tenant memory injection が exploitable(OWASP A01 / CWE-639)。self-signup + PRO 加入で被害者 workspace に任意 payload を注入可能
- **[CIO]** model docstring が約束する **3つの UNIQUE 制約が baseline migration に存在しない**(resource_events / resource_schemas / indexer_state)
- **[CIO/CSO]** `resource_id` が authoritative entity を持たない文字列 identifier、`ResourceToken` は `workspace_id` を schema で持たず tenancy 境界が convention
- **[CDO]** サイドバーに Resources 入口が無い(Token-first 導線)、`/workspace/developer/*` / `/workspace/oauth-apps/` / `/workspace/integrations/{api-keys,oauth-apps,resource-tokens}/` が orphan 重複
- **[DB PhD]** #318 の正解は無条件 UNIQUE ではなく `partial UNIQUE WHERE op='upsert'`(event log 理論 + business invariant 整合)

## Scope

**再設計**: Resource 層のみ(entity / ingest / indexer_state / schema / ResourceToken + Resource 側 frontend IA)
**維持**: Context / Memory / Auth / Workspace / MCP / sleep

## Child issues

- [ ] A. #322 [Critical/Security] Ingest-time workspace boundary check + resource_id collision audit
- [ ] B. #323 refactor(db): introduce resources entity + workspace_id FK + 3 UNIQUE constraints
- [ ] C. #324 fix(indexer): Qdrant named-vector upsert
- [ ] D. #325 feat(frontend): Resource top-level nav + remove orphan/duplicate routes
- [ ] E. #326 feat(frontend): Resource detail — schema creation + Indexer Status panel
- [ ] F. #327 docs: migration guide + resource_id slug → UUID resolution shim

## Supersedes

- #318 Resource ingest API silently accepts duplicate (resource_id, doc_id, version)
- #319 Wire orphan resource schema creation UI components
- #320 Resource indexer upsert fails with 'Not existing vector name' on Qdrant collection

## Sequencing

[...]

PR #330, opened 2026-04-14
refactor(db): introduce resources entity + resource_pk shadow FK (#323)

### LINKED ISSUE #323: refactor(db): introduce resources entity + workspace_id FK + 3 UNIQUE constraints (supersedes #318)
**Parent epic**: #321
**Priority**: Critical (data integrity + security foundation)
**Supersedes**: #318

## Summary

Resource サブシステムの schema 破綻を単一 Alembic migration で一掃する:

1. \`resources\` authoritative entity を新設
2. \`resource_events\` / \`resource_schemas\` / \`indexer_state\` / \`resource_tokens\` の \`resource_id\` 文字列を UUID FK へ移行
3. baseline migration に**欠落している 3 つの UNIQUE 制約** を追加
4. \`ResourceToken.workspace_id\` NOT NULL FK を追加(tenancy 境界を schema で具現化)

## Evidence

- \`backend/src/models/resource.py:66\`, \`:94-95\`, \`:141-142\` — docstring は 3 UNIQUE を約束
- \`backend/alembic/versions/157247e0df86_baseline_*.py\` — 実在する UNIQUE は \`idempotency_key\` と \`token_hash\` のみ
- \`resource_id\` を identity とする 4 テーブルが FK を持たない。workspace 境界 schema 化されず

### New table

\`\`\`sql
CREATE TABLE resources (
  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  workspace_id UUID NOT NULL REFERENCES workspaces(id) ON DELETE CASCADE,
  resource_id VARCHAR(255) NOT NULL,  -- external-facing slug
  name TEXT,
  created_at TIMESTAMP NOT NULL DEFAULT now(),
  created_by VARCHAR(255),
  UNIQUE (workspace_id, resource_id)
);
\`\`\`

### FK migration (4 tables)

\`resource_events\` / \`resource_schemas\` / \`indexer_state\` / \`resource_tokens\` に \`resource_pk UUID FK\` 列を追加 → backfill → 旧 \`resource_id VARCHAR\` は deprecated コメント付きで残す(段階移行、次リリースで DROP)。

### 3 UNIQUE constraints (DB PhD 補正済み述語)

\`\`\`python
# resource_events: partial UNIQUE (upsert 行のみ; delete event は replay 可能)
op.create_index(
    \"ux_resource_events_upsert_version\",
    \"resource_events\",
    [\"resource_pk\", \"doc_id\", \"version\"],
    unique=True,
    postgresql_where=sa.text(\"op = 'upsert'\"),
)

# resource_schemas: 無条件 UNIQUE
op.create_unique_constraint(
    \"unique_resource_schema_version\",
    \"resource_schemas\",
    [\"resource_pk\", \"schema_version\"],
)

# indexer_state: 無条件 UNIQUE (state table, PK 候補)
op.create_unique_constraint(
    \"unique_resource_context\",
    \"indexer_state\",
    [\"resource_pk\", \"context_id\"],
)
\`\`\`

### ResourceToken.workspace_id

\`\`\`python
op.add_column(\"resource_tokens\", sa.Column(\"workspace_id\", sa.UUID(), nullable=True))
# backfill from resources.workspace_id via resource_pk
op.alter_column(\"resource_tokens\", \"workspace_id\", nullable=False)
op.create_foreign_key(
    \"fk_resource_tokens_workspace\",
    \"resource_tokens\", \"workspaces\",
    [\"workspace_id\"], [\"id\"], ondelete=\"CASCADE\",
)
\`\`\`

## DB PhD の重要補正

\`resource_events\` は append-only event log。無条件 UNIQUE だと upsert → delete → upsert (蘇生)が禁じられ business invariant 違反。述語は **\`op = 'upsert'\`** が正解(\`version IS NOT NULL\` は不十分 ── delete specific-version の replay を阻害)。

## Acceptance criteria

- [ ] \`resources\` table 作成 + workspace-scoped UNIQUE
- [ ] 4 テーブル backfill スクリプト(A で監査した衝突データは事前に手当済み前提)
- [ ] 3 つの UNIQUE 制約が実在する(\`\\d\` で確認)
- [ ] \`ResourceToken.workspace_id\` NOT NULL FK
- [ ] 既存 REST API / MCP 契約は resource_id slug 受付で後方互換維持(F の shim で吸収)
- [ ] integration test: duplicate upsert → 409, delete specific-version replay → 204, indexer_state duplicate insert → IntegrityError
- [ ] migration down() で全手順を逆順復元可能

## Dependencies

- **A (hotfix) を先行 merge**(ingest exploit を閉じた状態で migration 作業)
- collision audit 結果をもとに既存 \`contexts.resource_id\` 重複を事前クリーンアップ

## Related

- Parent: #321
- Supersedes: #318

JFK:
Closes #323

## Summary

- Phase 1 of 2 for epic #321. Adds authoritative `resources` table (UUID PK, `workspace_id` FK, `UNIQUE(workspace_id, resource_id)`) and backfills it from the active rows in `contexts`.
- Introduces `resource_pk UUID` as a **nullable shadow FK** on the four satellite tables (`resource_events`, `resource_schemas`, `indexer_state`, `resource_tokens`) and `resource_tokens.workspace_id` as a nullable shadow FK to `workspaces`.
- Adds the three UNIQUE constraints the model docstrings have always promised but baseline never materialized — as **partial indexes** on `resource_pk IS NOT NULL` so pre-#324 writer paths continue to work unchanged.
- Ships with a pre-migration audit that fails fast on orphan satellite rows (no matching active context), the `resource_events` partial UNIQUE via `CREATE INDEX CONCURRENTLY`, and a documented dual-write prohibition for the transition period.

## Implementation notes

- **Phase strategy**: column nullability stays permissive in this PR so existing writers that only set `resource_id VARCHAR` keep working. Tightening to `NOT NULL` lives in **#325**; application writers get migrated to populate `resource_pk` in **#324**. Deploying this PR alone does not break production — proven by E2E smoke on the live dev DB (`make migrate` applied a97, API container rebooted healthy, MCP `recall`/`remember` continued to work, and a legacy direct INSERT with both shadow columns NULL was accepted).
- **Migration step order** (see `backend/alembic/versions/a97_resources_entity.py` docstring): create `resources` → seed from active contexts → orphan audit → add `resource_pk` + backfill + FK/index on 4 satellites → add `resource_tokens.workspace_id` + backfill (depends on the just-populated `resource_pk`) → create 3 partial UNIQUE indexes (`resource_events` via CONCURRENTLY, mirroring the a96 pattern from #329).
- **Dual-write rule**: once #324 lands, `resource_pk` is the source of truth. `resource_id VARCHAR` columns stay as read-side transitional aids only, populated via shim, and will be dropped once every writer has switched over.
- **Downgrade** is complete — `resource_id VARCHAR` columns are preserved across the round-trip so legacy queries remain readable from a96 state.
- **Tests** (`backend/tests/integration/test_resources_foundation_migration.py`):
  - `test_a97_data_bearing_roundtrip` — seed → up → assert backfill + partial UNIQUE existence + revival pattern → down → assert legacy reads work → re-up → assert idempotency.
  - `test_a97_audit_fails_on_orphan` — seeds an orphan satellite row and asserts the audit raises with an actionable message.
  - `test_a97_legacy_writes_without_resource_pk_still_work` — verifies a pre-#324 writer can INSERT into all four satellites without `resource_pk` AND that the partial UNIQUE correctly skips `resource_pk = NULL` duplicates (Phase 1 compatibility guarantee).

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: yellow (quick-win addressed in commit 69d122f; remaining notes deferred to #324)
- cto: green
- gate1: yellow via /ask → CIO (operator also consulted /cto and /qa-lead before branch creation; both returned conditional green)
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/323-feat-refactor-db-introduce-resources-entity-w.gate1.md`
- `~/.claude/cache/gh-issue-driven/323-feat-refactor-db-introduce-resources-entity-w.gate2.md`

## Coordination

- **#324** (writer migration) must follow to populate `resource_pk` on every write path.
- **#325** (NOT NULL tightening) closes the Phase 1/2 rollout.
- Keep #324 and #325 scheduled in the same milestone so the dual-column state does not persist.

### DIFF
--- a/backend/alembic/versions/a97_resources_entity.py
+++ b/backend/alembic/versions/a97_resources_entity.py
@@ -447,19 +447,28 @@ def upgrade() -> None:
     # ``resource_pk IS NOT NULL`` (Phase 1 semantics). Mirror a96's
     # INVALID-index guard for resilience against a prior failed
     # CONCURRENTLY run.
-    invalid_index_result = bind.execute(
-        sa.text(
-            "SELECT 1 FROM pg_class c "
-            "JOIN pg_index i ON i.indexrelid = c.oid "
-            "WHERE c.relname = :index_name AND NOT i.indisvalid LIMIT 1"
-        ),
-        {"index_name": _PARTIAL_UNIQUE_INDEX},
-    )
-    has_invalid_index = invalid_index_result.fetchone() is not None
+    # Both concurrently-built indexes need the same INVALID-index guard:
+    # ``IF NOT EXISTS`` alone would skip the rebuild of a partially-built
+    # index left behind by a prior failed CONCURRENTLY run, silently
+    # leaving production without the index. Detect and drop any INVALID
+    # version first, then recreate.
+    concurrent_indexes = (_PARTIAL_UNIQUE_INDEX, "ix_resource_events_resource_pk")
+    invalid_names = {
+        row[0]
+        for row in bind.execute(
+            sa.text(
+                "SELECT c.relname FROM pg_class c "
+                "JOIN pg_index i ON i.indexrelid = c.oid "
+                "WHERE c.relname = ANY(:names) AND NOT i.indisvalid"
+            ),
+            {"names": list(concurrent_indexes)},
+        ).fetchall()
+    }
 
     with op.get_context().autocommit_block():
-        if has_invalid_index:
-            op.execute(sa.text(f"DROP INDEX CONCURRENTLY IF EXISTS {_PARTIAL_UNIQUE_INDEX}"))
+        for name in concurrent_indexes:
+            if name in invalid_names:
+                op.execute(sa.text(f"DROP INDEX CONCURRENTLY IF EXISTS {name}"))
         op.execute(
             sa.text(
                 f"CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS {_PARTIAL_UNIQUE_INDEX} "
@@ -470,8 +479,7 @@ def upgrade() -> None:
         # ``ix_resource_events_resource_pk`` is created concurrently for
         # the same reason as the partial UNIQUE: resource_events is the
         # high-write append-only log, so a blocking index build would
-        # stall ingest traffic for the duration. IF NOT EXISTS lets this
-        # path be safely re-run after a partial failure.
+        # stall ingest traffic for the duration.
         op.execute(
             sa.text(
                 "CREATE INDEX CONCURRENTLY IF NOT EXISTS "
--- a/backend/alembic/versions/a97_resources_entity.py
+++ b/backend/alembic/versions/a97_resources_entity.py
@@ -1,0 +1,325 @@
+"""Introduce resources entity + resource_pk FK + 3 UNIQUE constraints.
+
+Issue #323: Normalize the resource subsystem. Add an authoritative
+``resources`` table keyed by UUID with a ``workspace_id`` FK, then
+introduce ``resource_pk UUID`` FK columns on the four existing tables
+that currently key themselves by ``resource_id VARCHAR`` alone
+(``resource_events``, ``resource_schemas``, ``indexer_state``,
+``resource_tokens``). Finally add the three UNIQUE constraints the model
+docstrings have always promised but the baseline migration never
+materialized, and make ``resource_tokens.workspace_id`` a NOT NULL FK so
+tenancy is enforced at the schema layer.
+
+Revision ID: a97_resources_entity
+Revises: a96_ctx_resource_id_unique
+
+NOTE: The revision ID is capped at 32 characters because
+``alembic_version.version_num`` is ``VARCHAR(32)`` in this database
+(asyncpg raises ``StringDataRightTruncationError`` otherwise).
+
+PHASE 1 OF 2 — SHADOW-COLUMN ROLLOUT
+------------------------------------
+This migration is the schema half of the resources normalization
+refactor. It **adds** ``resource_pk`` and ``resource_tokens.workspace_id``
+as *nullable* shadow columns and backfills them from existing data, but
+does **not** tighten them to NOT NULL. Tightening lives in a follow-up
+(#325 in the same epic) so the intervening release can ship application
+code updates (#324) that populate the new columns on every write.
+
+If ``resource_pk`` were tightened to NOT NULL here, every existing write
+path that only supplies ``resource_id VARCHAR`` would start throwing
+``NullViolationError`` on the first request after deploy — the whole
+point of the shadow phase is to let old writers continue working while
+new writers migrate over.
+
+DUAL-WRITE PROHIBITION (forward-looking)
+----------------------------------------
+Once application writers are updated to populate ``resource_pk``, they
+MUST NOT write ``resource_id`` independently. The ``resource_id``
+VARCHAR column, while it still exists, should only be set via a shim
+that looks up the slug through ``resources.id = resource_pk``.
+Independent writes will silently diverge and break the invariant this
+migration is preparing. The legacy ``resource_id`` column will be
+dropped once every writer has switched over (tracked under epic #321).
+
+UNIQUE constraint semantics during Phase 1
+-------------------------------------------
+The three UNIQUE constraints are added as **partial indexes** with
+``WHERE resource_pk IS NOT NULL`` (composed with ``op = 'upsert'`` on
+``resource_events``). During Phase 1 this protects backfilled rows and
+any new writes that correctly populate ``resource_pk``; rows that still
+write only ``resource_id`` with ``resource_pk = NULL`` are unprotected
+— matching the pre-migration status quo, so no regression. After #325
+tightens the columns to NOT NULL, the partial predicate becomes
+universally true and the constraints are effectively full UNIQUEs.
+
+Step order inside ``upgrade()``:
+    1. CREATE TABLE ``resources``
+    2. Seed ``resources`` from active contexts (workspace_id, resource_id)
+    3. Pre-migration audit: abort if any satellite row references a
+       ``resource_id`` with no matching active context, which would
+       leave ``resource_pk`` NULL permanently and defeat the audit.
+    4. ADD COLUMN ``resource_pk`` nullable on four satellite tables
+    5. Backfill ``resource_pk`` via JOIN on ``resources.resource_id``
+       (kept nullable — see "PHASE 1 OF 2" above)
+    6. Add FK + index on ``resource_pk`` per table
+    7. ADD COLUMN ``resource_tokens.workspace_id`` nullable
+    8. Backfill ``workspace_id`` from ``resources.workspace_id`` via
+       the just-populated ``resource_pk``
+    9. Add FK + index on ``resource_tokens.workspace_id``
+    10. Create three partial UNIQUE indexes
+        (``resource_events`` goes through CREATE INDEX CONCURRENTLY
+        because it is the high-write satellite table)
+"""
+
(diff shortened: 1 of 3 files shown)


JFK:
Loop 1: addressed 2 actionable Copilot comments (docstring mismatch on resource_tokens.workspace_id NOT NULL language, and audit error message saying 'public context' when the seed/audit uses 'active context' semantics). Skipped 1 comment (explicit op.create_index for resources.workspace_id) — the index is already created by Column(index=True) inside op.create_table, confirmed live on the dev DB. Re-requesting review.

JFK:
Loop 2 concluded with no_actionable_feedback: Copilot did not return a new review after loop-1 fixes (commit 458a544) were pushed. Exiting the /gh-issue-driven:ship Copilot loop after 2 iterations. The PR is ready for human review — feel free to re-invoke /gh-issue-driven:review or leave a manual review.

JFK:
Loop 2: all 3 actionable comments addressed in fdc9093.

- **CONCURRENTLY index on resource_events.resource_pk** — created in the same autocommit_block as the partial UNIQUE so the high-write log never takes ACCESS EXCLUSIVE for its index build. Three low-write satellites keep the blocking path.
- **Duplicate pre-audit before Step 10** — each of the three partial UNIQUEs now runs a distinct-key collision check first and raises with concrete (table, key, count) examples. No more cryptic low-level PG errors on index build.
- **ix_resources_resource_id** — added so the orphan audit and backfill UPDATEs that look up by resource_id alone get an index-backed path. Verified on live dev DB via a downgrade→upgrade round-trip (25 indexes present, backfill intact).

Skipped on FK NOT VALID (Copilot's secondary suggestion on the same line): intentionally deferred to the #325 scale-split per the migration's Scale note. Current pre-v1 row counts do not justify the added complexity; the split is triggered by `resource_events` > ~1M rows. Re-requesting review.

JFK:
Loop 3: all 3 comments addressed in 62f6d3f.

- **Explicit `op.create_index`** for `ix_resources_workspace_id` and `ix_resources_resource_id`, removing the inline `Column(index=True)` flags. The indexes WERE already being created via SQLAlchemy's implicit `Table.create()` (verified live on dev in loop 2), but explicit is clearer and matches how the rest of this migration writes indexes. Autogenerate drift avoided.
- **Model alignment**: added `index=True` to `Resource.resource_id` so the ORM declares the same index the migration creates.
- **Docstring fix**: Step 3 audit comment no longer references 'step 6's NOT NULL tightening' (which doesn't exist in Phase 1). Rewrote to describe the actual consequence — a surviving NULL leaves the backfill incomplete.

Live dev DB cycled through downgrade → upgrade cleanly, 4 expected indexes on `resources` present post-upgrade. Re-requesting review.

JFK:
Copilot review loop closed after 4 iterations. No new feedback after loop-3 push (62f6d3f). Summary of the 3-commit review journey:

- **Loop 1** (458a544): docstring mismatch on `resource_tokens.workspace_id` NOT NULL language + audit error message "public context" → "active context".
- **Loop 2** (fdc9093): `CREATE INDEX CONCURRENTLY` for `ix_resource_events_resource_pk` + pre-UNIQUE duplicate audit + explicit `ix_resources_resource_id`.
- **Loop 3** (62f6d3f): explicit `op.create_index` for both `resources` indexes (removing inline `Column(index=True)`) + ORM alignment (`Resource.resource_id` gains `index=True`) + docstring fix (removed stale reference to 'step 6 NOT NULL tightening').

One suggestion was deliberately skipped with rationale: `ix_resources_workspace_id` was asked about in loop 1 — the skip reasoning held until loop 3 when Copilot revisited the same area, at which point explicit `op.create_index` was adopted anyway. FK `NOT VALID` was also deferred to the scale-split per #325 (the migration Scale note anticipates this).

All 3 integration tests pass; live dev DB cycled clean through downgrade → upgrade; 25 expected indexes present. The PR is ready for human review.

JFK:
Loop 4: all 4 comments addressed in 986a461.

- **Cross-workspace ambiguity audit (Step 3a)**: preflight check that aborts when a resource_id appears in multiple `contexts.workspace_id` values (including soft-deleted contexts) AND satellite rows reference it. This catches the scenario you raised — workspace A owns the slug, context is soft-deleted leaving resource_id populated, workspace B later reuses the slug, and the backfill would silently re-home A's historical rows to B. The audit uses a single CTE-based query to avoid N round-trips.
- **resource_events FK NOT VALID + VALIDATE**: the high-write log now gets `ADD CONSTRAINT ... NOT VALID` (no scan, no ACCESS EXCLUSIVE) followed by `VALIDATE CONSTRAINT` (SHARE UPDATE EXCLUSIVE, non-blocking for reads/writes). Future inserts are enforced from the moment ADD CONSTRAINT lands; only the one-time backfilled-row validation is moved off the blocking path. The three low-write satellites stay on the simpler default.
- **Restart-semantics docstring**: added a module-level section acknowledging the autocommit_block trade-off. Short version: the INVALID-index guard handles re-runs of the concurrent section; a mid-CONCURRENTLY failure elsewhere requires manual cleanup on retry. This matches the a96 pattern already in production.

Round-trip verified on live dev DB: all 4 satellite FKs show `convalidated=t` post-upgrade, all 3 integration tests pass. Re-requesting review.

JFK:
Copilot review loop fully closed after 4 addressed iterations + 1 final poll (max_loops=5 cap reached). No new feedback after loop-4 push (986a461). The PR is ready for human review.

**4-commit review journey**:
- **Loop 1** (458a544) — docstring: `resource_tokens.workspace_id` NOT NULL language corrected to nullable shadow; audit error 'public context' → 'active context'.
- **Loop 2** (fdc9093) — perf: `CREATE INDEX CONCURRENTLY` for `ix_resource_events_resource_pk`; pre-UNIQUE duplicate audit at Step 10; added `ix_resources_resource_id`.
- **Loop 3** (62f6d3f) — explicit `op.create_index` for `resources` indexes (removing ambiguous inline `Column(index=True)`); ORM `Resource.resource_id` gained `index=True`; stale 'step 6 NOT NULL' docstring reference removed.
- **Loop 4** (986a461) — safety: cross-workspace ambiguity audit at Step 3a (catches A→deleted→B slug reuse with historical satellite rows); `resource_events` FK via NOT VALID + VALIDATE (non-blocking); restart-semantics docstring section.

Deferred with rationale: FK NOT VALID on the three low-write satellites (not worth the complexity at current scale). The restart-semantics trade-off is documented and matches the a96 pattern already in production.

Each loop was verified via live dev DB round-trip (downgrade → upgrade) plus the 3 integration tests (data-bearing roundtrip + audit-fails-on-orphan + legacy-writer-still-works). Final state: 25 indexes in place, all FKs validated, backfill intact.

Ready for merge decision; follow-ups #324 (writer migration) and #325 (NOT NULL tightening) remain in epic #321.

JFK:
Loop 5: the late 1-comment review has been addressed in a327a99.

Extended the INVALID-index detection to cover BOTH concurrent indexes (`ux_resource_events_upsert_version` + `ix_resource_events_resource_pk`) in a single `pg_class JOIN pg_index` query. Any INVALID sibling is dropped concurrently before the CREATE calls so `IF NOT EXISTS` cannot silently mask a missing index. Live dev DB round-trip confirms both indexes have `indisvalid = t` post-upgrade.

That was the sole remaining Copilot comment. The PR is now fully caught up with everything Copilot has raised across 5 review rounds.

Issue #334, opened 2026-04-15
fix(indexer): resource_indexer ignores per-context embedding_model (hardcodes legacy collection)


JFK:
**Parent epic**: #321
**Follow-up to**: #324 (Layer A: named-vector shape fix)

## Summary

`resource_indexer` hardcodes `KAGURA_MEMORIES_COLLECTION` (legacy 512-dim `kagura_memories`) at 5 call sites. When a context uses an embedding model other than `text-embedding-3-small` (512 dim, OpenAI) — e.g. `qwen3-embedding:8b` (4096 dim) — the correct collection is `kagura_memories_qwen3_embedding_8b_4096`, but the indexer still writes to the legacy collection. Qdrant rejects the upsert with a dim mismatch.

## Evidence

- `backend/src/services/resource_indexer.py`: 5 call sites referencing `KAGURA_MEMORIES_COLLECTION` (upsert + delete + cleanup paths).
- Existing correct resolution pattern: `backend/src/services/search_service.py:150-151` (`get_collection_name(model, dimensions)`) and `backend/src/services/memory_service.py:84-85`.
- `backend/src/config/constants.py:139-147` defines per-model dimensions (text-embedding-3-small: 512, qwen3 0.6b: 1024, qwen3 4b: 2560, qwen3 8b: 4096).

## Scope

- In `resource_indexer.py`, resolve each event's context → `search_config.embedding_model` + `embedding_dimensions` → `get_collection_name(model, dims)` → pass as `collection_name=` to all `qdrant_client.upsert/delete` calls.
- Do **not** delete the `KAGURA_MEMORIES_COLLECTION` constant itself (still used by `memory_service` and other call sites).

## Acceptance criteria

- [ ] All 5 `KAGURA_MEMORIES_COLLECTION` call sites in `resource_indexer.py` replaced with `get_collection_name(model, dimensions)` resolution.
- [ ] `text-embedding-3-small` context → still writes to legacy `kagura_memories` (backward compatible).
- [ ] `qwen3-embedding:8b` context → writes to `kagura_memories_qwen3_embedding_8b_4096`.
- [ ] Unit test: two contexts with distinct embedding models; assert correct `collection_name` and embedding dimension for each.
- [ ] Integration test (local-only): upsert via indexer into two collections, assert `points_count` increases in each independently.

## Priority rationale

[...]

PR #340, opened 2026-04-15
fix(indexer): resource_indexer per-context EmbeddingService (#338)

### LINKED ISSUE #338: fix(indexer): resource_indexer EmbeddingService ignores per-context embedding_model (Layer C)
**Parent epic**: #321
**Follow-up to**: #334 (Layer B: per-context Qdrant collection routing)

## Summary

After #334, `resource_indexer` now resolves the correct **Qdrant collection** for each context's embedding model. However, the **embedding generation itself** still uses the global default model (`settings.embedding_model`, typically `text-embedding-3-small`, 512 dim) regardless of what the context's `ContextSearchConfig.embedding_model` says.

This means a context configured with e.g. `qwen3-embedding:8b` (4096 dim) will now route to the correct collection `kagura_memories_qwen3_embedding_8b_4096`, but the generated embedding is still 512-dim → Qdrant rejects with dim mismatch.

**Together, #334 (Layer B) + this issue (Layer C) are required for non-default embedding models (e.g. qwen3) to actually work end-to-end.**

## Evidence

- `backend/src/services/resource_indexer.py:75`:
  ```python
  self.embedding_service = EmbeddingService(db)  # ← no model passed, falls back to settings.embedding_model
  ```
- `backend/src/services/embedding_service.py:54`: `self.model = model or settings.embedding_model`
- Existing reference pattern: `backend/src/services/memory_service.py:87-93` (`_get_embedding_service_for_config(config)`) constructs a per-context EmbeddingService:
  ```python
  def _get_embedding_service_for_config(self, config) -> EmbeddingService:
      if config:
          return EmbeddingService(
              self.db, model=config.embedding_model, dimensions=config.embedding_dimensions
          )
      return self.embedding_service
  ```
- TODO comment was added in #334 at \`_apply_upsert\` documenting this.

## Scope

- In \`resource_indexer.py\`, after resolving \`ContextSearchConfig\` (currently inside \`_resolve_collection_for_context\`), use the same config to construct or select an \`EmbeddingService\` instance with the correct \`model\` + \`dimensions\`.
- Refactor: \`_resolve_collection_for_context\` could be expanded into something like \`_resolve_routing_for_context(context_id) -> (collection_name, embedding_service)\` to fetch \`ContextSearchConfig\` once and produce both. This avoids a second SELECT per batch.
- Pass the per-context EmbeddingService into \`_apply_upsert\` (or assign to a per-batch local) so the embedding generated matches the collection's dim.

## Acceptance criteria

- [ ] \`_apply_upsert\` uses an EmbeddingService configured for the context's \`embedding_model\` / \`embedding_dimensions\`.
- [ ] No additional SELECT per event (resolve \`ContextSearchConfig\` once per batch alongside the collection_name resolution).
- [ ] Unit test: a context with \`qwen3-embedding:8b\` triggers \`EmbeddingService(model='qwen3-embedding:8b', dimensions=4096)\`.
- [ ] Unit test: a context with \`text-embedding-3-small\` (or no \`ContextSearchConfig\`) still uses the legacy default.
- [ ] Remove the TODO comment block in \`_apply_upsert\` once Layer C lands.

## Priority rationale

Same as #334: blocks any qwen3 (or non-default embedding) context from reaching production. **Must land before any qwen3 context is promoted to production.** v0.12.1 milestone.

## Related

- Parent: #321
- Follow-up to: #334 (Layer B)
- Sibling: #335 (sparse bm25 send, depends on full Layer B+C resolution)

JFK:
Closes #338

## Summary
- After #334 routed per-context Qdrant collections, `_apply_upsert` still used the global default `EmbeddingService`, so contexts using non-default models (e.g. `qwen3-embedding:8b` → 4096 dim) had upserts rejected on dim mismatch against the target collection.
- Fuse the per-batch routing helper into `_resolve_routing_for_context` returning `(collection_name, embedding_service)` from a single `ContextSearchConfig` fetch — mirrors `memory_service._get_embedding_service_for_config`.
- Pass the per-context `EmbeddingService` through `_apply_upsert` so the generated embedding dim always matches the resolved collection.

## Implementation notes
- `backend/src/services/resource_indexer.py`:
  - Renamed `_resolve_collection_for_context` → `_resolve_routing_for_context`; returns `tuple[str, EmbeddingService]` from the same `ContextSearchConfig` row (single SELECT per batch).
  - `_apply_upsert` gains a 5th arg `embedding_service: EmbeddingService` and calls `await embedding_service.embed(...)` instead of `self.embedding_service.embed(...)`.
  - Removed the `TODO(#338, Layer C)` comment block.
  - Docstring on the helper encodes the single-source-of-truth invariant and warns against re-splitting (guards against the two-layer-bug pattern from #324/#334/#338).
  - Fallback (no `ContextSearchConfig`): `(get_collection_name("text-embedding-3-small", 512), self.embedding_service)` — preserves legacy routing and the indexer's default-model service.
- `backend/tests/services/test_resource_indexer.py`:
  - `TestResolveRoutingForContext`: 4 cases — legacy / qwen3 / no-config fallback / single-SELECT invariant (asserts `db.execute.await_count == 1`).
  - `TestProcessIncrementalResolvesRoutingOncePerBatch`: renamed + updated to assert both `collection_name` AND `embedding_service` flow into `_apply_upsert` for every event in a multi-event batch.
  - Two existing `_apply_upsert` signature sites updated to pass `indexer.embedding_service` as the 5th arg.
- `CHANGELOG.md`: single bullet under `### Fixed` in the `[Unreleased] v0.12.0` section.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/338-fix-fix-indexer-resource-indexer-embeddingse.gate1.md
- ~/.claude/cache/gh-issue-driven/338-fix-fix-indexer-resource-indexer-embeddingse.gate2.md


JFK:
Loop 1: addressed 3 actionable items.

**Addressed:**
- L793 fallback drift: fallback collection now derives from `self.embedding_service.model/dimensions` so an operator-overridden default can't reintroduce the two-layer bug.
- `_apply_upsert` contract test: new `test_apply_upsert_uses_passed_embedding_service_not_self` with a sentinel vector asserts the passed-in service (not `self.embedding_service`) is used.
- CHANGELOG: #334 entry's "Layer C is a follow-up" note updated to reference #338.

**Skipped:**
- Caching `EmbeddingService` instances per (model, dimensions): `EmbeddingService.__init__` is pure (settings + registry lookup only); the Ollama connectivity probe is lazy in `embed()`, not at construction. Per-batch churn cost is negligible — the suggestion optimizes a non-issue.

JFK:
Loop 2: addressed 3 actionable items (all tracking the same loop-1 drift — doc/test didn't match the fixed code).

**Addressed:**
- `_resolve_routing_for_context` docstring: rewritten to describe the fallback as "indexer's default EmbeddingService paired with the collection derived from that service's model/dimensions", with an explicit note on the intentional divergence from memory_service (which predates multi-model settings).
- CHANGELOG: fallback description now matches the code path — out-of-the-box resolves to `kagura_memories`, but settings overrides are honored.
- Test renamed to `test_no_search_config_falls_back_to_default_service_and_matching_collection`; asserts `name == get_collection_name(indexer.embedding_service.model, indexer.embedding_service.dimensions)` instead of hardcoding `"kagura_memories"`. Still resolves to the same value under defaults, but survives an operator override.

JFK:
Loop 3: addressed 1 actionable item (revert of loop-1's fix based on Copilot's cross-service drift finding) + 1 skipped.

**Addressed — cross-service split-brain risk:**
Copilot correctly flagged that loop-1's "derive fallback collection from `self.embedding_service.model/dimensions`" change created a new cross-service drift: an operator overriding `settings.embedding_model` without creating a `ContextSearchConfig` row would have `memory_service` reading the legacy `kagura_memories` collection while `resource_indexer` writes to the overridden collection — silent split-brain.

Reverted the fallback to `get_collection_name("text-embedding-3-small", 512) + self.embedding_service`, matching `memory_service._get_context_collection_name()` exactly. The drift that motivated loop 1 (settings override + fallback path) is the lesser evil — it's loud (Qdrant dim mismatch, visible error) vs silent (cross-service data divergence). Docstring now states operators overriding settings must create `ContextSearchConfig` rows.

**Skipped (2nd time) — caching `EmbeddingService` instances:**
`memory_service._get_embedding_service_for_config` uses the same "new EmbeddingService per call" pattern without caching, so the Ollama per-instance probe is an accepted cost across both services. Addressing it belongs in a cross-service refactor (a follow-up issue), not this PR.

JFK:
Loop 4: addressed 1 actionable item + 1 final decline.

**Addressed — defensive warn log on fallback dim-mismatch:**
Added a structured `logger.warning` in `_resolve_routing_for_context` when no `ContextSearchConfig` row exists AND `self.embedding_service.model/dimensions` don't match the legacy (`text-embedding-3-small`, 512) pair. The upsert would still fail loudly at Qdrant with a dim-mismatch error, but the warn surfaces the misconfiguration at routing-resolve time with a clear hint ("create a ContextSearchConfig row").

**Declined (3rd and final time) — caching `EmbeddingService` instances:**
`process_incremental` is invoked per `(resource_id, context_id)`, so each invocation = 1 `EmbeddingService` construction = at most 1 Ollama probe. `memory_service` uses the same "new per call" pattern without caching. Adding caching only to `resource_indexer` creates intra-codebase inconsistency for negligible savings (~1 probe per 60s+ batch). If caching is warranted, it belongs in a cross-service helper refactor, not this PR. Filing as a v0.13.x follow-up candidate.

JFK:
Loop 5 (max): 1 new comment received — a 4th request for `EmbeddingService` caching per `(embedding_model, embedding_dimensions)`.

This suggestion has now been declined in loops 1, 3, and 4 with consistent rationale:
- Per-batch construction cost is at most 1 Ollama probe per `process_incremental` invocation
- `memory_service._get_embedding_service_for_config` uses the same "new per call" pattern — adding caching only to `resource_indexer` creates intra-codebase inconsistency
- If the probe overhead becomes measurable under real ingest load, the right fix is a cross-service helper in `services/context_routing.py` that both services adopt, not a one-sided cache in `ResourceIndexer`

This convergence point is a known design decision, not an oversight. Filing a v0.13.x follow-up candidate for the cross-service refactor.

Exiting the `/gh-issue-driven` Copilot loop at max_loops=5. PR is ready for human review.

PR #346, opened 2026-04-15
fix(api): use constraint_name attribute for resource_events IntegrityError dispatch (#318)

### LINKED ISSUE #318: Resource ingest API silently accepts duplicate (resource_id, doc_id, version)
## Summary

`POST /api/v1/resources/{resource_id}/events` silently accepts duplicate `(resource_id, doc_id, version)` combinations even though both the implementation code and the docstring promise a 409 Conflict response.

Discovered during a manual smoke test of the resource ingest endpoint.

## Context

**Priority**: medium
**Suggested labels**: bug
**Suggested milestone**: v0.13.0

## Expected behavior
- `backend/src/api/routes/resource_ingest.py:240` catches `IntegrityError` for `unique_resource_doc_version` and returns 409
- Docstring at `resource_ingest.py:126` declares "409: Duplicate version (resource_id, doc_id, version already exists)"
- `ingest_batch` at `resource_ingest.py:382` has matching skip-on-duplicate logic

## Actual behavior
- The `resource_events` table has no `unique_resource_doc_version` UNIQUE constraint (verified via `\d resource_events`)
- Only `idempotency_key` is UNIQUE
- Duplicate `(resource_id, doc_id, version)` silently insert as new rows

## Reproduction

```bash
# 1st request — 201 event_id=1
curl -X POST "http://localhost:8080/api/v1/resources/test/events" \
  -H "X-Resource-API-Key: <token>" -H "Content-Type: application/json" \
  -d '{"op":"upsert","doc_id":"DOC-001","version":1,"payload":{"title":"a"}}'

# 2nd request, same doc_id+version — expected 409, got 201 event_id=2
curl -X POST "http://localhost:8080/api/v1/resources/test/events" \
  -H "X-Resource-API-Key: <token>" -H "Content-Type: application/json" \
  -d '{"op":"upsert","doc_id":"DOC-001","version":1,"payload":{"title":"b"}}'
```

DB state after the two requests:

```
 id | resource_id |   op   | doc_id  | version
----+-------------+--------+---------+---------
  1 | test        | upsert | DOC-001 |       1
  2 | test        | upsert | DOC-001 |       1   ← duplicate accepted
```

## Proposed approaches (needs design decision)

**A. Add UNIQUE constraint via Alembic migration — align reality with code/docs**
- `UNIQUE (resource_id, doc_id, version) WHERE version IS NOT NULL`
- Requires duplicate cleanup before deploy

**B. Treat append-only as canonical — align code/docs with reality**
- Dedup only via `idempotency_key`
- Remove the dead 409 handler and update docstring

## Acceptance criteria
- Design decision (A or B) recorded in this issue
- Code, docstring, and schema aligned to the chosen design
- Integration test pinning the chosen behavior added under `backend/tests/`

## Related files
- `backend/src/api/routes/resource_ingest.py:240` (single ingest 409 handler)
- `backend/src/api/routes/resource_ingest.py:382` (batch ingest skip handler)
- `backend/src/models/resource.py` (ResourceEvent model)
- Related: #238 (Resource-driven incremental indexing)

JFK:
Closes #318

## Summary
- Replace substring-match on `str(IntegrityError)` with `e.orig.diag.constraint_name` across the 3 resource_events ingest handlers (single, batch, MCP).
- Centralize `ux_resource_events_upsert_version` and `resource_events_idempotency_key_key` in `backend/src/db/constraint_names.py`; migration, model, route, and MCP tool all source the same constants.
- Re-raise unknown IntegrityError in the single-ingest path (was silently masked as a generic 500). Batch and MCP keep their partial-success semantics.
- Also fixes the latent idempotency_key drift discovered mid-PR: handlers matched `unique_idempotency_key` but PG's default name is `resource_events_idempotency_key_key`, so that branch had been dead since baseline.

## Implementation notes
- New module `backend/src/db/constraint_names.py`: 2 name constants + `integrity_error_constraint_name(e)` helper (reads psycopg's structured `diag.constraint_name`).
- Declarative `Index(...)` in `ResourceEvent.__table_args__` and migration `a97_resources_entity._PARTIAL_UNIQUE_INDEX` now import the same constant.
- Handlers re-raise unknown IntegrityError in the single path so it surfaces via FastAPI default 500 for alerting instead of being masked. Batch path keeps `accumulate + continue` so one bad row does not abort sibling events; MCP path keeps `logger.warning + per-item error` with structured constraint field.
- Integration tests pin the constant-to-PG-name contract at runtime: if either constant drifts from what PostgreSQL actually raises, all 3 handler sites break simultaneously and the integration tests fail loudly.
- Scope deliberately includes the idempotency_key drift because it lives in the same `except` clauses and same test harness; splitting would weaken the constant-module justification.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/318-fix-resource-ingest-api-silently-accepts-duplica.gate1.md`
- `~/.claude/cache/gh-issue-driven/318-fix-resource-ingest-api-silently-accepts-duplica.gate2.md`

## Follow-ups (non-blocking, surfaced by gate2 advisors)
- Verify `api/main.py` has no debug IntegrityError handler that would leak SQL/params through the new re-raise path (CSO).
- Add unit tests for batch + MCP dispatch sites (currently covered indirectly by integration constant-check) (QA Lead).
- Investigate pre-existing `test_downgrade_to_base_and_upgrade` failure on `neural_memory_edges.valid_edge_type` - unrelated to this PR (QA Lead).

Generated via /gh-issue-driven:ship

### DIFF
--- a/backend/alembic/versions/a97_resources_entity.py
+++ b/backend/alembic/versions/a97_resources_entity.py
@@ -104,6 +104,7 @@ import sqlalchemy as sa
 from sqlalchemy.dialects.postgresql import UUID
 
 from alembic import op
+from db.constraint_names import RESOURCE_EVENTS_UPSERT_UNIQUE
 
 # revision identifiers, used by Alembic.
 revision = "a97_resources_entity"
@@ -123,7 +124,7 @@ _SATELLITE_TABLES = (
     "resource_tokens",
 )
 
-_PARTIAL_UNIQUE_INDEX = "ux_resource_events_upsert_version"
+_PARTIAL_UNIQUE_INDEX = RESOURCE_EVENTS_UPSERT_UNIQUE
 
 
 def upgrade() -> None:
--- a/backend/src/api/routes/resource_ingest.py
+++ b/backend/src/api/routes/resource_ingest.py
@@ -286,9 +286,19 @@ async def ingest_event(
             )
 
     # 3. Create event record
+    #
+    # ``resource_pk`` is sourced from the auth token (backfilled by migration
+    # a97 for existing tokens) so the partial UNIQUE
+    # ``WHERE op='upsert' AND resource_pk IS NOT NULL`` actually applies.
+    # If the token was issued before a97 and never backfilled, we pass
+    # ``None`` — ingest still works via the legacy ``resource_id`` column,
+    # the partial UNIQUE just skips those rows (same as pre-fix behavior).
+    # Migration #325 tightens ``resource_tokens.resource_pk`` to NOT NULL,
+    # at which point this branch goes away.
     try:
         event = ResourceEvent(
             resource_id=resource_id,
+            resource_pk=token_record.resource_pk,
             op=request.op,
             doc_id=request.doc_id,
             version=request.version,
@@ -467,9 +477,10 @@ async def ingest_batch(
                     )
                     continue
 
-            # Create event
+            # Create event — see single-ingest path for the resource_pk rationale.
             event = ResourceEvent(
                 resource_id=resource_id,
+                resource_pk=token_record.resource_pk,
                 op=event_req.op,
                 doc_id=event_req.doc_id,
                 version=event_req.version,
--- a/backend/src/api/routes/resource_ingest.py
+++ b/backend/src/api/routes/resource_ingest.py
@@ -14,6 +14,11 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth.resource_tokens import ResourceTokenManager
 from db.base import get_db
+from db.constraint_names import (
+    RESOURCE_EVENTS_IDEMPOTENCY_UNIQUE,
+    RESOURCE_EVENTS_UPSERT_UNIQUE,
+    integrity_error_constraint_name,
+)
 from models.auth import Context
 from models.resource import ResourceEvent, ResourceToken
 from models.schemas import (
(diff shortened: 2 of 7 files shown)


JFK:
Loop 1: addressed 3 runtime-critical Copilot findings:

- **[`integrity_error_constraint_name`]** Now handles asyncpg shape (`orig.constraint_name` direct) first, psycopg `.diag` second. Production uses asyncpg so the original helper returned `None` at runtime — genuine bug catch.
- **[`ingest_batch`]** Wrapped per-event add+flush in `async with db.begin_nested()` to match the MCP path. Partial-success now actually works.
- **[single-ingest unknown constraint]** Replaced bare `raise` with `HTTPException(500)` so the app-wide `SQLAlchemyError` handler doesn't intercept and convert to 503.

Skipped:
- **Migration imports from `db/constraint_names`**: kept as-is. The model's declarative `Index()` already has the same dependency, so the migration is no more fragile than existing code. Copilot's concern is valid in general but not in this repo's current state — worth a follow-up discussion if we ever harden migration hermeticity.

Test added: asyncpg-shaped `_make_integrity_error(shape='asyncpg')` + matching helper test case. All 24 unit + integration tests pass; lint + type-check clean.

JFK:
Loop 2: addressed the root-cause issue Copilot flagged.

**The partial UNIQUE was dead at runtime**. Index predicate is `WHERE op='upsert' AND resource_pk IS NOT NULL`, but writers never populated `resource_pk`, so the index excluded every row. The dispatch fix in loop 1 was restoring logic that could never actually fire.

Fix:
- **HTTP paths** (single + batch): `resource_pk=token_record.resource_pk` — sourced from the already-resolved auth token. Zero extra queries.
- **MCP ingest**: SELECT `resources.id` once per batch from `(workspace_id, resource_id)` before the event loop.
- **Fallback**: when `resource_pk` is null (pre-a97 token not backfilled, or missing resources row), event.resource_pk stays None — ingest still works via the legacy VARCHAR column, partial UNIQUE skips it. Same as pre-fix behavior, no regression. Migration #325 removes this branch when it tightens the FK to NOT NULL.

Test pins that ingest_event constructs ResourceEvent with `resource_pk=token_record.resource_pk`.

Skipped — the migration-import suggestion still stands from loop 1 review; the model's declarative Index has the same dependency so the migration is no worse.

JFK:
Loop 3:

- **engine.dispose() fix**: applied. Module-scoped fixture now wraps in try/finally to close the pool.
- **migration import**: keeping as-is (raised in loop 1, 2, 3). The model's declarative `Index(RESOURCE_EVENTS_UPSERT_UNIQUE, ...)` already imports from `db/constraint_names.py`, so deleting the module breaks the model just as hard as the migration. The integration tests pin the constant vs. actual PG-raised name at runtime, which is the stronger guarantee. Treating this as agree-to-disagree and ending the Copilot loop.

Planning to stop the loop here unless there are new blocking findings.

PR #391, opened 2026-04-19
fix(auth): restrict /workspace/resources to workspace owner (#389)

JFK:
Closes #389

## Summary

- Backend: swap `APIKeyOrSessionUser` → `WorkspaceOwner` on 4 `/v1/resources*` read endpoints; chain `PermissionService.resolve_resource_by_slug` on the 3 slug-path endpoints for cross-workspace 404 uniform disclosure (CWE-639 / OWASP A01 mitigation).
- Backend: `require_workspace_owner` now emits `workspace_owner_denied` WARN log on the deny path so audit pipelines can surface role violations across every `WorkspaceOwner`-gated endpoint.
- Frontend: silent `router.push("/workspace/dashboard")` for non-owner roles on both `/workspace/resources` and `/workspace/resources/[id]` (matches the #365 / #381 UX); hide the nav entry via the existing `requiredWorkspaceRole: "owner"` filter in `Sidebar.tsx`.
- Tests: 4 backend unit (403) via dep-override + 3 backend integration (404 cross-workspace) via real DB + 6 frontend role-gate tests (3 roles × 2 pages).

## Implementation notes

- **Two-layer authorization boundary**: same-workspace non-owner → **403 Forbidden** (from `WorkspaceOwner`); cross-workspace probe → **404 Not Found** (from `resolve_resource_by_slug` uniform disclosure). This matches `resource_indexer.py`'s established disclosure contract and reuses existing helpers — zero new abstractions introduced.
- `resource_schema.py:get_resource_impact` previously had a hand-rolled `SELECT User.current_workspace_id` + `SELECT Context.id` workspace-boundary check; that block is now superseded by `resolve_resource_by_slug` and removed.
- `list_resources` keeps `PermissionService.get_accessible_contexts` below the role gate as **defense-in-depth** — for owners it is a ~10ms no-op, but it preserves a safety net if the role gate ever regresses. Documented in the handler docstring.
- Existing `test_resource_indexer_api.py::client_workspace_a` fixture had to be extended to override `require_workspace_owner` (returns `(user_id, WORKSPACE_A)`) — 9 existing tests continue to pass.
- Sidebar uses the existing `requiredWorkspaceRole: "owner"` field already consumed by `workspaceSettings` and `externalKeys` (#381) — pure pattern extension.

## Residual security risk (tracked in #390)

[...]

PR #394, opened 2026-04-19
fix(backend): graph reads honor Context.is_private (#383)

### LINKED ISSUE #383: fix(backend): graph reads honor Context.is_private via PermissionService instead of hardcoded user_id filter
## Summary

`GraphService` and the `/graph/stats` / `/graph/data` endpoints filter all graph reads by `user_id == current_user`, hardcoded. Consequence: in a shared workspace context, **only memories authored by the requesting user appear in the graph** — workspace members see an empty or partial graph even when the context has plenty of shared memories authored by others.

This does not match the "Universal AI Memory Platform" + MCP multi-member workspace positioning, and it duplicates an authorization model the codebase already designed (`PermissionService.can_access_memory`) but forgot to apply to the graph layer.

## Symptoms

- Owner creates memories in context X → graph tab ("Connections" / "Graph") populated.
- Member of same workspace opens same context → both tabs show empty-state or near-empty data, even though the context is shared and the member has legitimate read access via `PermissionService`.
- User perception: "only owner's data exists in the graph" (actually: only the caller's own data is returned).

## Root cause

1. `backend/src/api/routes/graph.py:110` and `:214` read `user_id = user["user_id"]`.
2. `backend/src/services/graph_service.py` uses `self.user_id` in every query (`edge_repo.get_top_connected_nodes(user_id, ...)`, `get_all_edges(user_id, ...)`, etc.).
3. `NeuralMemoryEdge` table has `user_id`, `workspace_id`, `context_id` — 3-level isolation added during the "Single Collection Migration". The schema migration happened, but the **service-layer retrofit stopped at `user_id`**.
4. The rest of the codebase (`memory_service.update_memory`, etc.) uses `PermissionService.can_access_memory` (`permission_service.py:575-612`), which correctly honors `Context.is_private` and `workspace_member` semantics. Graph does not.

## Fix direction

Align graph reads with the designed authorization model:

- **Remove** the `user_id == caller` hardcoded filter from `GraphService` and both endpoints.
- **Scope** by `workspace_id + context_id` (already passed in — they just need to become the primary filter).
- **Respect `Context.is_private`**: if the context is private, restrict to `memory.user_id == caller` for nodes AND to edges whose both endpoints are owned by the caller.
- **Edge visibility rule**: an edge is visible if **both** endpoint memories are visible to the caller per `can_access_memory`. Private-context data must not leak via edges pointing from/to shared memories.
- Prefer embedding the filter at SQL level (joining to `Memory.context_id → Context.is_private`) rather than post-filtering, to avoid N+1 and preserve pagination semantics.

### Security invariant (CSO-reviewed)

- Private context memories/edges MUST NOT be exposed to non-creators.
- Workspace boundary MUST be enforced (no cross-workspace leakage even for shared contexts).
- Shared-context data IS visible to all workspace members with appropriate role.

## Acceptance Criteria

- [ ] `/graph/stats` and `/graph/data` no longer filter by `user_id == caller`; they return all visible nodes/edges for the requested context.
- [ ] For a shared context, a `member`-role caller sees the same graph as the `owner`.
- [ ] For a private context, a non-creator sees an empty graph (or only their own portion, depending on chosen policy — spec in implementation PR).
- [ ] Edges are filtered to respect per-endpoint visibility — no edge leaks that reference a private memory the caller cannot see.
- [ ] Cross-workspace isolation is preserved: `workspace_id` mismatch continues to return empty/404.
- [ ] Integration tests cover the 4-way matrix: {shared | private context} × {creator | other workspace member}.
- [ ] No regression in graph latency (measure before/after — adding the `is_private` join should be index-friendly).

## Out of scope (file follow-ups if needed)

- Unifying ALL memory-read paths (recall, explore, etc.) through `PermissionService.can_access_memory` — that is the broader cleanup addressed by the parallel "authorization unification" effort.
- `NeuralMemory` table's `user_id` column itself (keep as creator metadata; only stop using it as a visibility filter).

## References

- `backend/src/api/routes/graph.py:87-189` (`get_graph_stats`)
- `backend/src/api/routes/graph.py:191-264` (`get_graph_data`)
- `backend/src/services/graph_service.py` (all `self.user_id` usages)
- `backend/src/models/memory.py:266-326` (`NeuralMemoryEdge` schema)
- `backend/src/services/permission_service.py:575-612` (the intended authorization model)
- Related: #381 — same "legacy `user_id` filter vs. designed `PermissionService` model" pattern in the external-keys path.

## Rationale for v0.13 milestone (not v0.12.1)

- Requires a non-trivial query rewrite and new integration tests.
- Changes visibility semantics of an existing endpoint — needs a staging rollout window.
- v0.12.1 is a stabilization patch; this is a behavioral change tied to the Neural milestone.

JFK:
Closes #383

## Summary

- `GraphService` / `NeuralEdgeRepository` / `/graph/*` routes moved off the hardcoded `user_id == caller` filter. Visibility is now driven by `PermissionService.check_workspace_access` + `Context.is_private`, aligning graph reads with the `can_access_memory` model that the rest of the backend already uses.
- Shared context in a workspace → every member sees the full graph. Private context → only the creator sees their subgraph. Cross-workspace UUID probes and soft-deleted contexts surface as uniform 404 (CWE-639 / OWASP A01).
- New helper `PermissionService.resolve_context_for_workspace_read` centralizes "SELECT Context → check_workspace_access → 403→404 remap" — mirrors the established `resolve_resource_by_slug` chokepoint pattern.
- Brand types (`CallerId`, `MemoryAuthorId`) added to `can_access_memory` to prevent the same-shape-str argument swap class that was behind the PR #391 gate-review escape.

## Implementation notes

- `NeuralEdgeRepository.{get_all_edges, get_stats, get_top_connected_nodes}`: `user_id: str` → `user_id: str | None = None`. `None` = aggregate across all creators (shared context); `str` = creator-scoped (private / admin paths). Existing internal callers (`decay.py`) unchanged.
- `GraphService.stats()`: new keyword-only `owner_filter` kwarg replaces the inline `self.user_id` reference in the total-nodes subquery.
- Migration `b02_383_graph_edges_composite_indexes.py`: adds `(workspace_id, context_id, src_id)` + `(workspace_id, context_id, dst_id)` composite indexes matching the new primary access pattern. Old `(user_id, src_id/dst_id)` indexes retained for private-context/admin fallback. Reversible downgrade.
- `/graph/data` no longer re-queries `graph_service.stats()` for total counts — derived from already-loaded `all_edges` and `all_node_ids`, saving 3 queries per visualization render.
- Deny-path observability: `resolve_context_for_workspace_read` emits `context_read_denied` structlog events (`info` for `not_found`, `warning` for `cross_workspace`) so forensics can distinguish stale-UUID navigation from enumeration probes, while the external 404 stays uniform.

## Test coverage

`backend/tests/integration/test_graph_visibility.py` — 9 integration tests against real Postgres:

- 6-way matrix: `{shared | private} × {creator | same-workspace other member | different-workspace user}`
- `/graph/data` visibility parity with `/graph/stats`
- Cross-workspace probe → 404 (CWE-639 pin, regression target from PR #391)
- Soft-deleted context → 404

`backend/tests/services/test_permission_service.py` — updated 3 `can_access_memory` tests to keyword-only + branded types.

60 tests pass (9 integration + 50 unit + 1 repository). pyright + ruff clean.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: green via /claude-c-suite:ask
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/383-fix-fix-backend-graph-reads-honor-context-is.gate1.md
- ~/.claude/cache/gh-issue-driven/383-fix-fix-backend-graph-reads-honor-context-is.gate2.md

## Scope deferred to follow-up issues

Tracked during gate1 / gate2, explicitly out of scope for this PR to keep the authz fix focused and independently deployable:

- **AC 5**: NULL backfill migration for `neural_memory_edges.workspace_id` / `context_id` (legacy rows from migration 062). Adds 2-step zero-downtime `NOT VALID` → `VALIDATE` pattern + `NOT NULL` enforcement. GDPR CASCADE hygiene — the authz fix itself works correctly around NULL rows (filtered out by the workspace/context predicate).
- **AC 6**: Write-path assertion + one-time audit script for the `edge.context_id == src_memory.context_id == dst_memory.context_id` invariant. The fix trusts this invariant; making it explicit prevents subtle cross-context regressions.
- **AC 3 (MCP parity)**: `backend/src/mcp_server/tools/edge.py` and `context.py` still hit `NeuralEdgeRepository` through the legacy creator-scoped path. HTTP routes are fixed but MCP readers see pre-fix behavior until a follow-up PR aligns them. Flagged by CSO + CTO gate2 as the most visible user-facing inconsistency post-merge.
- **`AuthzChecked[T]` phantom wrapper + Visibility ADT** (CS/PL PhD recommendation): refactors `owner_filter: str | None` tri-state into a type-level sum. Target milestone: v0.16 Hardening, under the `#237` type-annotation audit umbrella.

## Test plan

- [x] `make test-local` — 472 unit tests pass, 60 touch this PR's scope
- [x] `tests/integration/test_graph_visibility.py` — 9 integration tests covering the 6-way matrix + soft-delete + cross-workspace
- [x] `tests/services/test_permission_service.py` — brand-type migration, 50 tests pass
- [x] Alembic migration `b02` reversible (upgrade/downgrade pair)
- [ ] Manual: post-deploy observe `/graph/stats` 404 rate on memory.kagura-ai.com for 1 week — visibility fix changes user-facing behavior (members in shared contexts will suddenly see workspace graph data)
- [ ] Manual: verify `b02` migration applies cleanly to production DB (index creation is online; no lock issues expected)

### DIFF
--- a/backend/alembic/versions/b02_383_graph_edges_composite_indexes.py
+++ b/backend/alembic/versions/b02_383_graph_edges_composite_indexes.py
@@ -16,6 +16,18 @@ admin paths (decay, consolidation) that filter by ``user_id``. A follow-up
 migration may remove them once production query plans confirm they are
 unused.
 
+ZERO-DOWNTIME INDEX BUILDS
+--------------------------
+``neural_memory_edges`` is a high-write table (every new memory + sleep
+consolidation cycle inserts rows). A plain transactional ``op.create_index``
+would take ``ACCESS EXCLUSIVE`` on the table for the duration of the build,
+blocking writes. This migration follows the same zero-downtime pattern
+a96/a97 established: ``autocommit_block`` + ``CREATE INDEX CONCURRENTLY
+IF NOT EXISTS`` so writes continue through the build, plus invalid-index
+guards on retry after a mid-build failure.
+
+Downgrade mirrors the pattern with ``DROP INDEX CONCURRENTLY IF EXISTS``.
+
 Revision ID: b02_383_edges_ws_ctx_idx
 Revises: b01_resource_pk_ph2
 Create Date: 2026-04-20
@@ -23,6 +35,7 @@ Create Date: 2026-04-20
 
 from collections.abc import Sequence
 
+import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
@@ -32,23 +45,53 @@ branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 
+_CONCURRENT_INDEXES: tuple[tuple[str, str], ...] = (
+    ("idx_edges_ws_ctx_src", "neural_memory_edges (workspace_id, context_id, src_id)"),
+    ("idx_edges_ws_ctx_dst", "neural_memory_edges (workspace_id, context_id, dst_id)"),
+)
+
+
+def _invalid_concurrent_indexes(name_iter: Sequence[str]) -> set[str]:
+    """Return concurrent indexes from ``name_iter`` that exist in an INVALID state.
+
+    A mid-build failure of ``CREATE INDEX CONCURRENTLY`` leaves the index row
+    in ``pg_index`` with ``indisvalid = false``. Re-running ``CREATE INDEX
+    CONCURRENTLY IF NOT EXISTS`` would skip (the name exists) without retrying,
+    leaving the cluster permanently without a usable index. This helper
+    enumerates names that need to be dropped first so the retry rebuilds cleanly.
+    """
+    bind = op.get_bind()
+    rows = bind.execute(
+        sa.text(
+            "SELECT c.relname "
+            "FROM pg_class c JOIN pg_index i ON c.oid = i.indexrelid "
+            "WHERE c.relname = ANY(:names) AND i.indisvalid IS FALSE"
+        ),
+        {"names": list(name_iter)},
+    ).fetchall()
+    return {row[0] for row in rows}
+
+
 def upgrade() -> None:
-    """Create composite indexes for the new shared-context graph read path."""
-    op.create_index(
-        "idx_edges_ws_ctx_src",
-        "neural_memory_edges",
-        ["workspace_id", "context_id", "src_id"],
-        unique=False,
-    )
-    op.create_index(
-        "idx_edges_ws_ctx_dst",
-        "neural_memory_edges",
-        ["workspace_id", "context_id", "dst_id"],
-        unique=False,
-    )
+    """Create composite indexes concurrently for the new shared-context graph read path."""
+    names = [name for name, _ in _CONCURRENT_INDEXES]
+    invalid = _invalid_concurrent_indexes(names)
+
+    with op.get_context().autocommit_block():
+        # Drop any INVALID leftovers from a prior failed attempt before the
+        # IF NOT EXISTS rebuild (IF NOT EXISTS alone would skip the rebuild).
+        for name in names:
+            if name in invalid:
+                op.execute(sa.text(f"DROP INDEX CONCURRENTLY IF EXISTS {name}"))
+
+        for name, table_and_cols in _CONCURRENT_INDEXES:
+            op.execute(
+                sa.text(f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {name} ON {table_and_cols}")
+            )
 
 
 def downgrade() -> None:
-    """Drop the composite indexes added in ``upgrade``."""
-    op.drop_index("idx_edges_ws_ctx_dst", table_name="neural_memory_edges")
-    op.drop_index("idx_edges_ws_ctx_src", table_name="neural_memory_edges")
+    """Drop the composite indexes concurrently, mirroring the upgrade pattern."""
+    with op.get_context().autocommit_block():
+        for name, _ in _CONCURRENT_INDEXES:
+            op.execute(sa.text(f"DROP INDEX CONCURRENTLY IF EXISTS {name}"))
--- a/backend/alembic/versions/b02_383_graph_edges_composite_indexes.py
+++ b/backend/alembic/versions/b02_383_graph_edges_composite_indexes.py
@@ -1,0 +1,54 @@
+"""Add (workspace_id, context_id) composite indexes on neural_memory_edges (#383).
+
+Issue #383: graph reads move from ``user_id == caller`` hardcoded filter to
+``PermissionService``-driven visibility. The new primary access pattern is
+``WHERE workspace_id = :ws AND context_id = :ctx (AND src_id|dst_id = :node)``
+— the existing ``idx_edges_user_src`` / ``idx_edges_user_dst`` composite
+indexes are ``user_id``-leading and therefore inert for shared-context reads
+under PostgreSQL's leftmost-prefix rule.
+
+This migration adds two composite indexes matching the new access pattern:
+``(workspace_id, context_id, src_id)`` and ``(workspace_id, context_id, dst_id)``.
+
+Old indexes are intentionally **retained** in this revision — they still
+serve the private-context read path (caller-scoped edges) and internal
+admin paths (decay, consolidation) that filter by ``user_id``. A follow-up
+migration may remove them once production query plans confirm they are
+unused.
+
+Revision ID: b02_383_edges_ws_ctx_idx
+Revises: b01_resource_pk_ph2
+Create Date: 2026-04-20
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "b02_383_edges_ws_ctx_idx"
+down_revision: str | Sequence[str] | None = "b01_resource_pk_ph2"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Create composite indexes for the new shared-context graph read path."""
+    op.create_index(
+        "idx_edges_ws_ctx_src",
+        "neural_memory_edges",
+        ["workspace_id", "context_id", "src_id"],
+        unique=False,
+    )
+    op.create_index(
+        "idx_edges_ws_ctx_dst",
+        "neural_memory_edges",
+        ["workspace_id", "context_id", "dst_id"],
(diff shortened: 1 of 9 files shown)


JFK:
Loop 1: addressed 2/3 Copilot comments in 4afabaa.

**Addressed:**

- **graph.py:158 (Memory lookup scope)** — Added `Memory.workspace_id`, `Memory.context_id`, `Memory.deleted_at.is_(None)` predicates to both the /graph/stats top-nodes fetch and /graph/data all-nodes fetch. Good defense-in-depth: even if the edge-level visibility filter is satisfied, the Memory metadata fetch now also verifies the node belongs to the resolved context. Protects against a stale or invariant-violating edge row leaking cross-context Memory summaries until AC 6 (edge context invariant) lands.

- **graph.py:95 (unused `context_service` dep + stale Query description)** — Removed the `context_service: ContextService = Depends(...)` parameter that was dead weight on every request. Updated the `context_id` Query description from "defaults to current context" to "returns empty graph when absent" to match the new contract.

**Deferred (rationale below, no code change in this loop):**

- **neural_edge.py:642 (`get_stats` + decay.py ValueError path)** — Correct catch on the signature-change implication, thank you. Verified the caller `DecayManager.get_decay_statistics` is **dead code** in the tree (`grep` shows zero invocations from `neural_tasks.py` or elsewhere). The latent ValueError predates #383 and making `get_stats` accept `user_id`-only-without-scope would weaken the 3-level isolation contract that `_validate_isolation_params` is load-bearing for. The cleaner path is to either delete `get_decay_statistics` or give it a proper workspace-scoped rewrite — both are out of scope for this authz PR but worth tracking. I have expanded the `get_stats()` docstring to document the caller contract explicitly so future callers don't reintroduce the same misuse.

Will re-request Copilot after push.

JFK:
Loop 2: addressed all 5 Copilot comments in 9848dc4.

**Addressed:**

- **permission_service.py:279 (coarse `cross_workspace` classification)** — Narrowed the deny-path logging. The HTTPException detail from `check_workspace_access` is now parsed into explicit reasons: `workspace_deleted`, `not_a_member`, `role_too_low`, `workspace_access_denied`. The old "cross_workspace" label mis-classified routine role-too-low / workspace-deleted denies as enumeration probes.

- **permission_service.py:139 (private-context access leak via 200-empty)** — Major catch, thank you. `resolve_context_for_workspace_read` now enforces the private-context creator-only rule inline (matching `can_access_memory` / `check_context_access` semantics) before the workspace check. Non-creator workspace members of a private context now get the same uniform 404 as a cross-workspace probe, closing the "this private context exists in your workspace and you're not its creator" differential leak. New deny reason: `private_non_creator`. Updated integration test to match.

- **graph_service.py:305 (`stats()` default flip regression)** — Real regression, fixed with a sentinel. `stats(owner_filter=...)` now defaults to a `_STATS_OWNER_DEFAULT` sentinel that resolves to `self.user_id` when the argument is omitted, preserving the pre-#383 per-user-metrics semantics for sleep/consolidation / neural_tasks / `get_node_metrics` internal call. Explicit `owner_filter=None` still selects shared-context aggregation for the HTTP routes that opt in. Three-value semantics documented in the docstring.

- **neural_edge.py:778 (`get_top_connected_nodes` missing isolation guard)** — Added `_validate_isolation_params(workspace_id, context_id)` at entry to match `get_all_edges` / `get_stats`. Prevents `user_id=None` + missing isolation params from devolving into unscoped full-table aggregation.

- **alembic/b02 (blocking CREATE INDEX on high-write table)** — Rewrote the migration to use `op.get_context().autocommit_block()` + `CREATE INDEX CONCURRENTLY IF NOT EXISTS`, mirroring the a96/a97 zero-downtime pattern. `neural_memory_edges` is a high-write table (every memory insert + sleep cycle writes rows); the transactional `op.create_index` would have taken ACCESS EXCLUSIVE for the build duration. The upgrade path also guards against INVALID-index leftovers from a prior failed CONCURRENTLY build (drops them before rebuilding, same as a97). Downgrade mirrors with `DROP INDEX CONCURRENTLY IF EXISTS`.

All 375 related tests pass (9 integration + 366 unit/repo). pyright + ruff clean.

Will re-request Copilot after this comment.

JFK:
Loop 3: addressed the 1 docstring nit in cd864e2. `/graph/stats` docstring updated to reflect the loop 2 behavior change (non-creators of private contexts denied at the context level via `resolve_context_for_workspace_read` with uniform 404, matching `can_access_memory` / `check_context_access` semantics). No implementation change. 9 integration tests still pass.

JFK:
Loop 4: addressed both Copilot comments in 7b12e2f.

- **graph.py:357 (stats vs renderable inconsistency)** — Good catch. `/graph/data` now computes `total_nodes` / `total_edges` from the **post-Memory-scope set** (`visible_node_ids` / `visible_edges`) instead of the raw edge table. An edge pointing to a soft-deleted or out-of-scope Memory is no longer counted in totals, so stats match what the UI can actually render.

- **graph.py:113 (docstring precision)** — Updated. The 404 for private-context non-creators is a deliberate CWE-639 remap of the 403 that `check_context_access` would raise; **semantics** match (creator-only), **status code** does not. Wording now says 'deliberately remaps that denial to 404' instead of the misleading 'matching check_context_access semantics'.

9 integration tests still pass. pyright + ruff clean.

Will re-request Copilot. Loop count is now 4 of 5 max — one more iteration left before we exit on max_loops if feedback continues.

JFK:
Loop 5 (final): addressed the 1 Copilot comment in 4c0e5cc.

- **graph.py:310 (early-return stats inconsistency)** — Good catch. Loop 4 migrated the **main** `/graph/data` return to post-Memory-scope stats but missed the early-return branch (triggered when no node passes the `memory_types` filter). The early return now uses the same `visible_node_ids` / `visible_edges` computation so empty responses don't report "total_edges > 0" when those edges reference out-of-scope or soft-deleted memories.

All 9 integration tests still pass. ruff clean. This is loop 5 of max 5; if Copilot has more feedback after this push, I'll hand off to human review rather than re-request automatically. Otherwise this converges the Copilot review cycle.

Issue #433, opened 2026-04-23
feat(ui): Memories tab on contexts/[id] page (wire dialog stubs)


JFK:
Split from #422. The "wire up the existing components" piece — turns out they were never wired in the first place.

## Context

`MemoryDetailDialog.tsx`, `EditMemoryDialog.tsx`, `MemoriesTable.tsx`, `MemoryCard.tsx`, `CreateMemoryDialog.tsx`, `DeleteMemoryDialog.tsx` all exist as files but are **not imported anywhere** in `frontend/src/app/`. The contexts/[id] page only renders Overview / Connections / Graph / Settings tabs. There is no list-of-memories view in the UI today.

## Scope

Add a new "Memories" tab to `frontend/src/app/(authenticated)/workspace/contexts/[id]/page.tsx`:

1. Append `"memories"` to `ADMIN_CONTEXT_TABS` (line ~58)
2. Create `frontend/src/components/contexts/MemoriesTabPanel.tsx` — wires `MemoriesTable` + `MemoryDetailDialog` + `CreateMemoryDialog` + `EditMemoryDialog` + `DeleteMemoryDialog`
3. Use `getMemories({ context_id })` — depends on #431 (backend filter)
4. Use the unified `Memory` type — depends on #432 (type unification)
5. Use `useTabParam` already in scope — `?tab=memories` URL sync

URL: `/workspace/contexts/[id]?tab=memories`

## Dependencies

Blocked on:
- #431 (backend `/memory/list` context_id filter)
- #432 (frontend Memory type unification)

## Out of scope

- Deep-link URL to a specific memory (`?memoryId=<uuid>`) — separate issue
- Graph node click → MemoryDetailDialog — separate issue
- Edge metadata overlay — separate issue

## Acceptance

- [ ] New "memories" tab visible to admin+ users on contexts/[id]
- [ ] `MemoriesTable` lists context-scoped memories
- [ ] Row click opens `MemoryDetailDialog`
- [ ] Create / Edit / Delete dialogs wired
- [ ] `make test-frontend` passes
- [ ] Manual smoke on memory.kagura-ai.com after deploy

Issue #431, opened 2026-04-23
feat(api): add context_id filter to GET /memory/list


JFK:
Split from #422. Backend prerequisite for context-scoped memory list UI.

## Problem

`GET /api/v1/memory/list` (`backend/src/api/routes/memory.py:464`) currently filters by `user_id`, `scope`, and `type` only. There is no way to ask "list memories within a specific context" — required by any per-context UI (e.g., a Memories tab on `contexts/[id]`).

`Memory.context_id` already exists on the model (`backend/src/models/memory.py:82`) with appropriate indexes, so this is purely a route-level addition.

## Fix

Add a `context_id: UUID | None = Query(None)` param to `list_memories` and an extra `WHERE Memory.context_id == context_id` clause when provided. Update `MemoryListResponse` if needed (likely no change — the response shape is per-row).

Also verify the query stays scoped to the caller (existing `user_id` filter must remain) — context_id alone must not bypass user isolation.

## Acceptance

- [ ] `context_id` query param added, filtered correctly when present
- [ ] `user_id` filter still applied — context_id is additive, not a substitute
- [ ] PermissionService check on the context (caller must have access) — see existing patterns in `backend/src/api/routes/graph.py` for context access enforcement
- [ ] `make lint` + `make test-local` pass
- [ ] Test added: list within a specific context_id, list without (legacy behavior unchanged)

## Out of scope

- Frontend client update — separate issue (depends on #422-split frontend Memory type unification)

## Why v0.14.0

Pre-req for the Memories tab work, but small enough to ship independently. Not v0.13.1 because it's part of the larger Memory UI rework that won't fit the 2026-04-27 release anchor.

PR #438, opened 2026-04-24
refactor(frontend): align Memory type with backend MemoryListItem (#432)

### LINKED ISSUE #432: refactor(frontend): align Memory type with backend MemoryListItem schema
Split from #422. Type unification prerequisite — without this, the UI cannot consume real backend responses.

## Problem

The frontend `Memory` interface (`frontend/src/lib/types/memory.ts:10-24`) and the backend `MemoryListItem` schema (`backend/src/models/schemas.py`) **do not align**:

| Frontend `Memory` | Backend `MemoryListItem` |
|---|---|
| `key`, `value`, `scope`, `agent_name`, `user_id`, `importance`, `tags`, … | `id`, `summary`, `type`, `scope`, `importance`, `created_at`, `updated_at` |
| Composite-key oriented (key + scope + agent_name) | UUID-id oriented |

`backend/src/api/routes/memory.py:464` returns `MemoryListItem`, but `frontend/src/lib/api/memory.ts:20` `getMemories()` types the response as `MemoryListResponse<Memory>`. Today this works only because **no production consumer wires the two together** (the related components — `MemoriesTable.tsx`, `MemoryDetailDialog.tsx`, `EditMemoryDialog.tsx` — are dead code, never imported).

The first real consumer (Memories tab — see split sibling issue) will fail at runtime. Fix the type contract first.

## Fix

1. Add `id: string` to frontend `Memory` interface
2. Decide on the canonical shape — likely make frontend `Memory` a superset that includes both `id` (UUID) and `key`/`value` (composite key + body), since different endpoints may return different subsets
3. Update `MemoriesTable.tsx` to key off `id` instead of the composite `key:scope:agent_name` (currently line 60-61)
4. Update `MemoryDetailDialog`, `EditMemoryDialog`, `DeleteMemoryDialog` to consume the unified type
5. Update `getMemories()` to align response field names

## Acceptance

- [ ] Frontend `Memory` type includes `id`, `summary`, all backend fields
- [ ] `MemoriesTable` uses `memory.id` for row identity (not composite key)
- [ ] Dialogs accept the unified type
- [ ] `make test-frontend` passes (existing dead-code components compile)
- [ ] No runtime change visible (still no consumer page wired up; that's a separate issue)

## Why v0.14.0

Type unification is a refactor that touches several files. Better to land it cleanly before the Memories tab UI is built, so the UI consumer doesn't have to deal with two type shapes.

JFK:
## Summary

- `Memory` type becomes a superset: `id: string` (required, canonical UUID identity) and `summary?: string` added. Legacy composite-key fields (`key/value/agent_name/user_id`) stay required so existing dead-code dialogs still compile.
- `MemoryListResponse` shape aligned with backend: `{ memories, total, has_more }` (was `limit/offset`).
- `MemoriesTable` now keys rows by `memory.id` — matches the `entity.id` convention already used by `SessionsTable` and `APIKeysTable`. Composite tuple retained only for the legacy bulk-delete API, which still addresses rows by `(key, scope, agent_name)`.
- v0.14.0 Memory UI cascade prerequisite (after #431, before #433).

## Deferred (documented in code comment)

The superset remains **type-unsound** in one direction: a response from the new `/memory/list` endpoint lacks `key/value/agent_name/user_id`, but the type declares those required. Making them optional would force null-guards into the dead-code dialogs, violating the AC \"dead-code components must still compile\". Resolution deferred to the consumer issue (#433) where we'll know whether the Memories tab actually needs those fields.

## Acceptance (#432)

- [x] Frontend `Memory` type includes `id`, `summary`, backend fields
- [x] `MemoriesTable` uses `memory.id` for row identity (not composite key)
- [x] Dialogs accept the unified type (auto-satisfied by superset — no changes needed)
- [x] `make test-frontend` passes
- [x] No runtime change visible (still no consumer page wired up)

## Test plan

- [x] \`npx tsc --noEmit\` — clean
- [x] \`make test-frontend\` — 239/239 tests pass
- [x] \`make lint\` — backend clean
- [x] \`/simplify\` — 1 finding applied (inline `uniqueKey` local)
- [ ] (post-merge) Memories tab wiring in #433 consumes the unified type

Closes #432

### DIFF
--- a/frontend/src/components/memories/MemoriesTable.tsx
+++ b/frontend/src/components/memories/MemoriesTable.tsx
@@ -56,7 +56,9 @@ export function MemoriesTable({
   const totalPages = Math.ceil(total / pageSize);
   const bulkDeleteEnabled = !!onSelectionChange;
 
-  // Helper to generate unique key for memory
+  // Composite key for the legacy bulk-delete API — the backend still
+  // addresses rows by (key, scope, agent_name), so selectedKeys must use
+  // this tuple. Row identity for React uses `memory.id` (UUID) instead.
   const getMemoryUniqueKey = (memory: Memory) =>
     `${memory.key}:${memory.scope}:${memory.agent_name}`;
 
@@ -156,10 +158,11 @@ export function MemoriesTable({
           </TableHeader>
           <TableBody>
             {memories.map((memory) => {
-              const uniqueKey = getMemoryUniqueKey(memory);
-              const isSelected = selectedKeys.includes(uniqueKey);
+              const isSelected = selectedKeys.includes(
+                getMemoryUniqueKey(memory),
+              );
               return (
-                <TableRow key={uniqueKey}>
+                <TableRow key={memory.id}>
                   {bulkDeleteEnabled && (
                     <TableCell>
                       <Checkbox
--- a/frontend/src/components/memories/MemoriesTable.tsx
+++ b/frontend/src/components/memories/MemoriesTable.tsx
@@ -36,13 +36,13 @@ interface MemoriesTableProps {
   // Bulk delete props (Issue #666). ``selectedIds`` holds ``Memory.id`` UUIDs
   // for rows already normalized into the ``Memory`` shape this table renders
   // (which expects ``key`` / ``agent_name`` columns — so a consumer fed from
-  // raw ``/memory/list`` rows must convert at the boundary first).
+  // raw ``GET /api/v1/memory/list`` rows must convert at the boundary first).
   //
   // Bulk-delete wiring is the parent's responsibility. The legacy
   // ``bulkDeleteMemories`` API addresses rows by (key, scope, agent_name),
-  // which ``/memory/list`` does not return; a consumer fed from that endpoint
-  // needs either a bulk-delete-by-id backend endpoint or pre-loaded composite
-  // fields. #433 will decide which path to take.
+  // which ``/api/v1/memory/list`` does not return; a consumer fed from that
+  // endpoint needs either a bulk-delete-by-id backend endpoint or pre-loaded
+  // composite fields. #433 will decide which path to take.
   selectedIds?: string[];
   onSelectionChange?: (ids: string[]) => void;
 }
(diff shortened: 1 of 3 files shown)


JFK:
## Copilot review disposition — final

This PR has been through 5 Copilot review rounds. Summary of outcomes:

### Fixed in this PR

| Round | Finding | Commit |
|---|---|---|
| 1 | Comment contradiction on superset | `443b63b` |
| 2 | Bulk-selection degrades to `"undefined:..."` | `443b63b` (switched to `memory.id`) |
| 2 | `MemoryType` too narrow (backend is `str`) | `85e3867` |
| 2 | Missing `MemoryListItem` type | `85e3867` |
| 3 | Radix `CheckedState` coercion bug | `7211758` |
| 3 | `MemoryListResponse` structural incorrectness | `7211758` (generic over `TItem`) |
| 4 | Endpoint path `/memory/list` vs `/api/v1/memory/list` | `f14d51c` + `ef353d6` |
| 4 | Generic default silent-drift for legacy callers | `f14d51c` (pinned to `<Memory>`) |

### Deferred to #433 (Memories tab consumer)

Two concerns Copilot has re-raised on every round; documented as deferred in code comments rather than fixed here:

**1. LegacyMemory / discriminated-union split** — making `key/value/agent_name/user_id` optional would force null-guards into the dead-code dialogs (`MemoryDetailDialog`, `EditMemoryDialog`, `DeleteMemoryDialog`) and violate the AC \"existing dead-code components compile\". The contract decision (weaken legacy fields vs. keep behind a conversion boundary) belongs with the consumer that actually reads both shapes. See `frontend/src/lib/types/memory.ts:15-28`.

**2. Bulk-delete translation implementability** — the legacy `bulkDeleteMemories` API addresses rows by `(key, scope, agent_name)` which `/api/v1/memory/list` doesn't return. There are three workable paths (bulk-delete-by-id endpoint, pre-loaded composite fields, or drop bulk-delete on the new endpoint). All are contract choices the consumer owns, not this type-alignment PR. See `MemoriesTable.tsx:34-45`.

### Out of scope (#432 is type alignment only)

- `getMemories()` hits `/memory` (missing `/api/v1/` prefix) — **pre-existing 404 bug** unrelated to type alignment. Will be resolved when #433 rewires the client to `/api/v1/memory/list`.
- `getMemoryStats()` has wrong path + signature — same category, pre-existing, #433 or a separate issue.

### AC check

- [x] `Memory` type includes `id`, `summary`, all backend fields
- [x] `MemoriesTable` uses `memory.id` for row identity
- [x] Dialogs accept the unified type
- [x] `make test-frontend` passes (253/253)
- [x] No runtime change visible

PR #453, opened 2026-04-26
feat(api+ui): graph node/edge click → detail dialog + edge metadata overlay (#435)

### LINKED ISSUE #435: feat(ui): graph node/edge click → detail dialog + edge metadata overlay
Split from #422. The original "Phase 2" goal — make the graph view interactive.

## Goal

In `frontend/src/components/contexts/GraphTabPanel.tsx` (uses `useForceSimulation` + d3-force on SVG):

- **Node single click** → open `MemoryDetailDialog` for that memory (reuse the dialog wired by sibling issue #433)
- **Edge click** → render a small floating overlay with edge metadata (source title / target title / `type` / `weight` / `created_at`)

## Critical implementation note: drag/click discrimination

d3-drag does NOT swallow click events — a naive `onClick` handler will fire after a drag-to-reposition finishes, opening the dialog unintentionally. Required fix in `frontend/src/hooks/useForceSimulation.ts`:

```ts
let dragStart: { x: number; y: number } | null = null;
onPointerDown: (e) => { dragStart = { x: e.clientX, y: e.clientY }; }
onPointerUp: (e) => {
  if (!dragStart) return;
  const dx = e.clientX - dragStart.x;
  const dy = e.clientY - dragStart.y;
  if (dx * dx + dy * dy < 25) {  // moved < 5px → click intent
    handleNodeClick(node);
  }
  dragStart = null;
}
```

## Implementation sketch

1. Add `onNodeClick` and `onEdgeClick` callback props to `useForceSimulation`
2. Update GraphTabPanel to handle node click → call referenceMemory(node.id) → open dialog
3. Edge click → set `selectedEdge` state → render absolute-positioned `<div>` overlay near the click point with metadata
4. Source/target memory titles: lookup in the `nodes` array (`nodes.find(n => n.id === edge.source)?.summary`)

## Dependencies

Blocked on:
- #430 (backend graph edge `created_at` + `confidence` exposure) — without it, the overlay can't show `created_at`
- #433 (Memories tab + dialog wiring) — node click reuses `MemoryDetailDialog` via the same dialog state machine
- #434 (deep-link URL) — node click should also update `?memoryId=<uuid>` for shareability

## UX semantics (decided 2026-04-23 via /ask + user confirm)

- **Single click only** for node → full dialog open. No double-click. No preview tooltip beyond the existing hover state.
- Right-click reserved for future Phase 3 (edge inline edit menu) — out of scope here.

## Acceptance

- [ ] Graph node single click opens `MemoryDetailDialog` (full body via referenceMemory)
- [ ] Drag-to-reposition does NOT fire dialog open (5px threshold)
- [ ] Graph edge click shows overlay with source title, target title, type, weight, `created_at`
- [ ] Overlay dismisses on outside click or Esc
- [ ] No regressions in existing hover tooltip / focus ring behavior
- [ ] `make test-frontend` passes
- [ ] Manual smoke on memory.kagura-ai.com after deploy

JFK:
Closes #435

## Summary
- Graph node single-click opens `MemoryDetailDialog` (deep-linked via `?memoryId=`, shared with the Memories tab so cross-tab navigation preserves the open memory).
- Drag/click discrimination INSIDE d3-drag's start/end events with a 5px squared threshold — no separate onPointerDown/Up handlers (those would race d3-drag's pointer capture). Keyboard parity (WCAG 2.1.1) via Enter/Space.
- Edge click renders a floating `GraphEdgeOverlay` (source title, target title, type, weight, optional `created_at` and `confidence`) anchored to the click point; dismissable on Esc / outside-click. Hit area is a transparent stroke-width=10 line stacked on the 1px visual line.
- Backend: shared-context memory list now returns all members' memories (matches the existing graph endpoint pattern); private contexts unchanged. Fixes admin/non-creator seeing empty memory list.

## Implementation notes
- Extracted `useMemoryDetailDialog` (race guard + deep-link bypass token, ~120 LOC) and `useMemoryIdParam` (URL R/W) into `frontend/src/hooks/`. MemoriesTabPanel shrinks from 393 → ~130 LOC; same hook now powers the Graph dialog flow.
- `useForceSimulation` gains `onNodeClick` / `onEdgeClick` callbacks; drag/click discrimination lives inside the existing dragBehavior `start`/`end` handlers using `event.x`/`event.y` (zoom-corrected SVG-local coords) so the 5px threshold behaves the same at every zoom level. Reduced-motion path keeps click via a dedicated `click` listener (d3-drag is not attached there).
- Backend `list_memories`: `owner_filter = user_id if context.is_private else None`, mirroring `api/routes/graph.py:159`. Both data and count queries gated identically. Added `test_context_id_shared_drops_user_id_filter` (asserts `user_id` NOT in WHERE) plus made the private-context test explicit.
- Frontend dev env: `next.config.ts` `output: "standalone"` is now gated on `NODE_ENV === "production"`; Docker builder stage sets `NODE_ENV=production` explicitly so the standalone bundle is still emitted reliably. Resolves a Next.js 16 + Turbopack dev-mode 500 (`ENOENT: .next/dev/server/app/page/build-manifest.json`).
- Context detail title icon now matches the contexts list page (Globe purple / Lock blue / Users green) — previously only distinguished private vs non-private.

## Behavior change to disclose
- `GET /api/v1/memory/list?context_id=…` for **shared (non-private) contexts** now returns memories from all workspace members with read access, not only the caller's own. Private contexts unchanged. Self-hosted operators on team plans should be aware that members of shared contexts will now see each other's memories via the Memories tab UI (the underlying graph view already exposed this). Documented in `CHANGELOG.md` under `Changed`.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green (CWE-639 maintained, no XSS, no CSRF surface, no new endpoints)
- qa-lead: yellow (manual smoke is the merge gate — 5px drag threshold not deterministically testable in jsdom)
- cto: green (hook split SRP-clean, useForceSimulation responsibility growth flagged for Phase 3)
- gate1: yellow via /claude-c-suite:ask (drag/click conflict avoidance confirmed by user; implementation followed the recommendation)
- review provider: copilot

## Manual smoke checklist (required before merge)
- [ ] Admin sees memories from all workspace members in a shared context
- [ ] Private context still creator-only
- [ ] Graph node click → dialog opens; drag 10px+ → no dialog; <5px → dialog (drag/click discrimination)
- [ ] Focus + Enter/Space on a node → dialog opens (keyboard parity)
- [ ] Edge click on the visible 1px line → overlay shows (hit area extension)
- [ ] Esc and outside-click dismiss the overlay
- [ ] Edge overlay open → click another edge → overlay swaps; click a node → dialog opens, overlay dismisses
- [ ] Context detail title icon: Globe (public/purple) / Lock (private/blue) / Users (shared/green)
- [ ] Cross-tab: open a memory from Graph → switch to Memories tab → same memory still open

- `~/.claude/cache/gh-issue-driven/435-feat-feat-ui-graph-node-edge-click-detail-dia.gate1.md`
- `~/.claude/cache/gh-issue-driven/435-feat-feat-ui-graph-node-edge-click-detail-dia.gate2.md`


JFK:
## Loop 6 — addressed Copilot review (post-max-loops)

The original `/gh-issue-driven:ship` invocation hit `max_loops=5`, but Copilot kept reviewing — addressing the new findings here as a follow-up commit (`d5e0068`).

**Applied (3):**
- `GraphEdgeOverlay`: focus the close button on mount and restore focus on unmount, matching the contract `role="dialog"` implies (APG dialog pattern + Radix precedent).
- `CHANGELOG.md`: typo `dismissable` → `dismissible`.
- `tests/api/test_memory_list.py`: shared-context test now asserts that `PermissionService.resolve_context_for_workspace_read` was invoked. Prevents a CWE-639 regression where a future refactor accidentally drops the permission gate.

**Documented as deferred (2):**
- **`useForceSimulation` `edgeClickEnabled` captured at effect run, not reactive to prop toggling**: current consumers (`GraphTabPanel`) always pass a stable callback, so the toggle path doesn't trigger. Adding `onEdgeClick` to the effect deps would force a full SVG rebuild on every parent render — a much larger regression than the never-exercised toggle case. Inline comment added.
- **Edge hit-line keyboard parity (tabindex/role/Enter/Space)**: deferred to the Phase 3 `edge inline edit` feature. Tab-stopping every edge in an N=50 graph without an alternate navigation model (edge list view) is poor UX; Phase 3 is the natural place to introduce that view alongside the existing `<tabindex>` strategy. This was the original Gate1 design decision; inline comment added so the intent survives.

Re-requested Copilot review.

🤖 Generated via /gh-issue-driven follow-up

JFK:
## Loop 7 — addressed Copilot review (graph simulation tick-against-detached-DOM)

Fixed in `74b8ee5` (will update with actual commit sha).

`handleDeleteSuccess` no longer calls `fetchData()`. Instead it prunes the deleted node + its edges from the local `graphData` state. The dep change flows through the existing `useMemo` → `useForceSimulation` cycle, which runs cleanup on the old simulation and rebuilds with the pruned set — same path that filter / preset changes already exercise.

Net effect:
- d3 simulation no longer tick-against-detached-DOM during refetch (the original concern)
- No loading flicker (canvas stays mounted)
- No network round-trip
- Trade-off: `stats.total_nodes` decremented optimistically. Acceptable since `DeleteMemoryDialog.onSuccess` only fires after a 2xx response.

Re-requested Copilot review.

🤖 Generated via /gh-issue-driven follow-up

PR #477, opened 2026-04-28
perf(stripe): wrap sync stripe-python calls in asyncio.to_thread (#468)

### LINKED ISSUE #468: perf(stripe): wrap sync stripe-python calls in asyncio.to_thread (or migrate to AsyncStripe)
## Context

Split out from #463 (item 2). #463 closed once items 1+3 shipped via #465; tracking item 2 here for atomic scoping.

## Problem

`backend/src/services/stripe_service.py` calls the **synchronous** `stripe-python` SDK from inside async functions, blocking the asyncio event loop for the duration of each HTTP round-trip (typically 200-800ms per call).

Affected sites (all async):

- L88: `stripe.checkout.Session.create` in `create_checkout_session`
- L117: `stripe.billing_portal.Session.create` in `create_portal_session`
- L298: `stripe.Subscription.cancel` in `cancel_subscription_and_delete_customer_for_erasure`
- L315: `stripe.Customer.delete` in `cancel_subscription_and_delete_customer_for_erasure`

The new erasure sweep (L298, L315) loops over owned workspaces × Stripe calls, amplifying the stall — a user with 50 workspaces queues 100 blocking calls.

## Fix options

**Option A — `asyncio.to_thread` per call**:

```python
result = await asyncio.to_thread(stripe.Subscription.cancel, sub_id)
```

Simple, minimal diff. Caveat: uses the default `ThreadPoolExecutor` (`min(32, os.cpu_count()+4)` workers). Erasure sweeps could saturate it under peak `create_checkout_session` load. Mitigation: dedicate a `ThreadPoolExecutor(max_workers=4)` for erasure sweep specifically via `loop.run_in_executor(...)`.

**Option B — migrate to `stripe.AsyncStripe`**:

```python
client = stripe.AsyncStripe(api_key=settings.stripe_secret_key)
session = await client.checkout.sessions.create(...)
```

Cleaner long-term. Larger diff (every call site changes API shape). Verify SDK version supports it.

## Acceptance

- [ ] Decision recorded (Option A with isolated executor / Option B / hybrid)
- [ ] All 4 sites updated
- [ ] No `stripe.X.Y(...)` direct calls remain in async functions
- [ ] Unit test: assert that calling the wrapped helper from an async context does not block the event loop (e.g., a concurrent `asyncio.sleep` completes on time)

## Related

- Parent: #463 (closed)
- Sibling: #469 (item 4 confirm_token gate)
- Originating PR: #464 (right-to-erasure)

JFK:
Closes #468

## Summary

- Wraps the four synchronous `stripe-python` calls in `backend/src/services/stripe_service.py` (L88, L117, L298, L315) with asyncio so they no longer block the event loop during 200-800ms Stripe HTTP round-trips.
- Adds a dedicated `ThreadPoolExecutor(max_workers=2, thread_name_prefix="stripe-erasure")` for the GDPR right-to-erasure sweep so a burst of erasure requests cannot starve checkout/portal traffic on the default `asyncio.to_thread` pool.
- Registers `executor.shutdown(wait=False, cancel_futures=True)` in the FastAPI app lifespan: queued work is dropped, the lifespan returns immediately, and any in-flight Stripe call keeps running in its worker thread until either the call completes or the orchestrator sends SIGKILL after grace period (k8s default 30s). Erasure is best-effort — Stripe processes cancel/delete server-side regardless of whether we read the ack.
- Adds `StripeError(ExternalServiceError, error_code="EXT-204")` to `utils/exceptions.py` so a defensive boundary check on `session.url is None` (Stripe types it `Optional[str]`) routes through `memory_cloud_exception_handler` as a typed 502 instead of an unhandled 500.

## Implementation notes

- Two helper functions in `stripe_service.py`:
  - `_run_stripe(func, *args, **kwargs)` — `asyncio.to_thread` for low-rate user-initiated checkout/portal calls.
  - `_run_stripe_erasure(func, *args, **kwargs)` — `loop.run_in_executor(dedicated_pool, functools.partial(...))` for the erasure sweep. `functools.partial` is required because `run_in_executor` only forwards positional args.
- `_get_erasure_executor()` lazy-initializes the dedicated pool so no threads are created when `BILLING_ENABLED=false`. `shutdown_erasure_executor()` is idempotent (safe before init, safe to call twice).
- `stripe.Webhook.construct_event` (L139) is **explicitly NOT wrapped** — it is HMAC + JSON parse, CPU-only with no I/O, so wrapping would only add context-switch overhead.
- Pattern matches existing repo conventions (`tasks/scheduler.py`, `db/redis.py`, `db/qdrant.py` all use the same `get_X() / shutdown_X()` lazy-singleton shape).

## Scope explicitly OUT

- **AsyncStripe migration** (Option B in the issue) — deferred. `stripe>=11.0.0` pin unchanged. AsyncStripe needs `stripe>=12.0.0` and rewriting every call site. Same wrappers can swap to AsyncStripe internally as a future drop-in.
- **Erasure sweep wall-clock parallelization** — out of scope. The sweep is still N workspaces × 200-800ms sequential. This change unblocks the *event loop* only; parallelization is a separate decision (would need a Semaphore for Stripe's 100 req/sec rate limit).

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped (no binary gate configured)
- binary_gate: (none)
- cso: green
- qa-lead: green (one info-level note: `StripeError` boundary raise is not directly unit-tested — small follow-up candidate, non-blocking)
- cto: green
- gate1: green via /claude-c-suite:ask (CTO lens)
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/468-feat-perf-stripe-wrap-sync-stripe-python-calls.gate1.md
- ~/.claude/cache/gh-issue-driven/468-feat-perf-stripe-wrap-sync-stripe-python-calls.gate2.md

## Test plan

- [x] 7 new unit tests in `backend/tests/services/test_stripe_service.py` cover: event-loop unblocking (100ms vs 50ms timing test with 40ms CI jitter headroom), kwargs preservation through both wrappers, dedicated-pool thread-name isolation, default-pool routing, idempotent shutdown (uninitialized + post-use)
- [x] 25 existing erasure regression tests pass (`tests/services/test_account_erasure_service.py`)
- [x] 4 existing billing RBAC tests pass (`tests/api/test_billing_rbac.py`)
- [x] `make lint` — All checks passed
- [x] `make type-check` — 0 errors, 0 warnings

### DIFF
--- a/backend/src/api/main.py
+++ b/backend/src/api/main.py
@@ -100,25 +100,15 @@ async def lifespan(app: FastAPI):
     logger.info("scheduler_stopped")
 
     # Stop the dedicated Stripe erasure ThreadPoolExecutor (Issue #468).
-    # No-op when billing is disabled (the executor is lazy-initialized).
-    # Bounded by a 30s timeout so an erasure mid-shutdown cannot hold the
-    # lifespan past typical orchestrator grace periods (e.g. Kubernetes
-    # terminationGracePeriodSeconds=30). Orphaned in-flight Stripe calls
-    # are equivalent to the SIGKILL fallback that would have fired anyway.
+    # ``wait=False`` returns immediately so lifespan shutdown is not held
+    # by in-flight Stripe HTTP calls — k8s SIGKILL after the grace period
+    # is the existing process-exit fallback either way, and erasure is
+    # best-effort (Stripe processes the request server-side regardless of
+    # whether we read the ack). No-op when billing is disabled.
     from services.stripe_service import shutdown_erasure_executor
 
-    stripe_erasure_shutdown_timeout_seconds = 30
-    try:
-        await asyncio.wait_for(
-            asyncio.to_thread(shutdown_erasure_executor),
-            timeout=stripe_erasure_shutdown_timeout_seconds,
-        )
-        logger.info("stripe_erasure_executor_stopped")
-    except TimeoutError:
-        logger.warning(
-            "stripe_erasure_executor_shutdown_timed_out",
-            timeout_seconds=stripe_erasure_shutdown_timeout_seconds,
-        )
+    shutdown_erasure_executor()
+    logger.info("stripe_erasure_executor_stopped")
 
     # Close database
     from db.base import close_db
--- a/backend/src/api/main.py
+++ b/backend/src/api/main.py
@@ -99,6 +99,13 @@ async def lifespan(app: FastAPI):
     shutdown_scheduler()
     logger.info("scheduler_stopped")
 
+    # Stop the dedicated Stripe erasure ThreadPoolExecutor (Issue #468).
+    # No-op when billing is disabled (the executor is lazy-initialized).
+    from services.stripe_service import shutdown_erasure_executor
+
+    shutdown_erasure_executor()
+    logger.info("stripe_erasure_executor_stopped")
+
     # Close database
     from db.base import close_db
 
--- a/backend/src/api/main.py
+++ b/backend/src/api/main.py
@@ -101,10 +101,24 @@ async def lifespan(app: FastAPI):
 
     # Stop the dedicated Stripe erasure ThreadPoolExecutor (Issue #468).
     # No-op when billing is disabled (the executor is lazy-initialized).
+    # Bounded by a 30s timeout so an erasure mid-shutdown cannot hold the
+    # lifespan past typical orchestrator grace periods (e.g. Kubernetes
+    # terminationGracePeriodSeconds=30). Orphaned in-flight Stripe calls
+    # are equivalent to the SIGKILL fallback that would have fired anyway.
     from services.stripe_service import shutdown_erasure_executor
 
-    shutdown_erasure_executor()
-    logger.info("stripe_erasure_executor_stopped")
+    stripe_erasure_shutdown_timeout_seconds = 30
+    try:
+        await asyncio.wait_for(
+            asyncio.to_thread(shutdown_erasure_executor),
+            timeout=stripe_erasure_shutdown_timeout_seconds,
+        )
+        logger.info("stripe_erasure_executor_stopped")
+    except TimeoutError:
+        logger.warning(
+            "stripe_erasure_executor_shutdown_timed_out",
+            timeout_seconds=stripe_erasure_shutdown_timeout_seconds,
+        )
 
     # Close database
     from db.base import close_db
--- a/backend/src/services/stripe_service.py
+++ b/backend/src/services/stripe_service.py
@@ -52,20 +52,23 @@ def _get_erasure_executor() -> ThreadPoolExecutor:
 
 
 def shutdown_erasure_executor() -> None:
-    """Tear down the erasure executor cleanly on app shutdown.
+    """Tear down the erasure executor on app shutdown.
 
     Registered from ``api.main.lifespan``. Safe to call when the executor
     was never created (no-op) and safe to call twice (second call is a
     no-op).
 
-    ``cancel_futures=True`` drops queued (not-yet-started) tasks so the
-    shutdown latency is bounded by ``max_workers × stripe_default_timeout``
-    (~160s) instead of ``pending_count × stripe_default_timeout``. In-flight
-    Stripe calls still finish so customer-deletion isn't half-committed.
+    Uses ``wait=False`` so this returns immediately rather than blocking
+    the lifespan on in-flight Stripe HTTP calls. ``cancel_futures=True``
+    drops queued (not-yet-started) tasks. Any in-flight Stripe call
+    keeps running in its worker thread until either the call completes
+    or the process is killed by the orchestrator's SIGKILL after grace
+    period — erasure is best-effort, so an orphaned ack is acceptable
+    (Stripe processes the cancel/delete server-side regardless).
     """
     global _erasure_executor
     if _erasure_executor is not None:
-        _erasure_executor.shutdown(wait=True, cancel_futures=True)
+        _erasure_executor.shutdown(wait=False, cancel_futures=True)
         _erasure_executor = None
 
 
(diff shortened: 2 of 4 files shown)


JFK:
**Loop 1 — addressed 1, skipped 1.**

✅ **Addressed**: Copilot @ `backend/src/api/main.py:107` — bounded `shutdown_erasure_executor()` to 30s via `asyncio.wait_for(asyncio.to_thread(...), timeout=30)`. Erasure mid-shutdown can no longer hold the lifespan past typical orchestrator grace periods (Kubernetes `terminationGracePeriodSeconds=30` default). On timeout, emits a structured `stripe_erasure_executor_shutdown_timed_out` warning so operators can detect forced shutdown via logs. Orphaned in-flight Stripe calls are equivalent to the SIGKILL fallback that would have fired anyway. Commit: c5d75f0.

⏭️ **Skipped (false positive)**: github-code-quality @ `backend/tests/services/test_stripe_service.py:54` "Statement has no effect" — line 54 is `await sleep_task`, which is load-bearing. It awaits the `asyncio.create_task(measure_asyncio_sleep())` from line 51 so `asyncio_done_at[0]` is populated before the assertion at line 57. Removing it would race the test against the measurement and make the assertion non-deterministic. The static analyzer appears to misread `await <task>` as a no-op when its result is unused, but the side effect (waiting for completion) is exactly what's needed.

JFK:
**Loop 2 — addressed both findings.**

✅ **api/main.py:115** (Copilot): correct critique — `asyncio.wait_for(asyncio.to_thread(shutdown_erasure_executor))` only bound the asyncio coroutine, not the underlying thread (Python threads can't be cancelled mid-call). Process exit was unaffected because non-daemon executor threads block interpreter shutdown via the atexit hook regardless.

Switched `shutdown_erasure_executor()` to `wait=False` (still with `cancel_futures=True` to drop the queue). Lifespan returns immediately. In-flight Stripe calls keep running in their worker threads until they complete or the orchestrator's SIGKILL fires after grace period — same eventual outcome as before, but honestly described. Erasure is best-effort: Stripe processes cancel/delete server-side regardless of whether we read the ack. Reverted the asyncio.wait_for wrapping in api/main.py since it added no real bound.

✅ **stripe_service.py:64** (Copilot): correct — docstring referenced \`stripe_default_timeout\` as if it were a configured constant, but it doesn't exist anywhere in this module or settings. Reworded the docstring to describe what \`wait=False\` + \`cancel_futures=True\` actually do, without inventing constant names.

Commit: 0cbee66.

JFK:
**Loop 3 — addressed (description-only).**

✅ **Copilot @ stripe_service.py:72**: correct — the loop-2 switch to `wait=False` was not reflected in the PR description, which still claimed `wait=True` and "drain cleanly". Updated the Summary bullet to describe the actual semantics: queued work dropped, lifespan returns immediately, in-flight Stripe calls run in worker threads until completion or orchestrator SIGKILL. No code changes this loop.

The implementation now matches the description: best-effort erasure shutdown, bounded by orchestrator grace period rather than by the Stripe SDK's HTTP timeout.

JFK:
**Loop 4 — addressed 1, skipped 1.**

✅ **Copilot @ stripe_service.py:167** (visible): added `test_create_checkout_session_raises_stripe_error_on_missing_url` to `backend/tests/services/test_stripe_service.py`. Mocks `stripe.checkout.Session.create` to return a session with `url=None`, mocks `_init_stripe`, `get_price_id`, and the DB workspace lookup, and asserts `StripeError("checkout Session.create returned no URL")` is raised. This closes the gap QA Lead gate2 review and Copilot loop 4 both independently surfaced. Commit: f88c3db.

⏭️ **Skipped (false positive)**: Copilot's low-confidence-suppressed finding at stripe_service.py:386 — `try: ... except Exception` swallows `asyncio.CancelledError` in Python 3.11. Since [Python 3.8](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError), `asyncio.CancelledError` inherits from `BaseException`, not `Exception`, so `except Exception:` does **not** catch it — cooperative cancellation is preserved. Repo runs Python 3.11 (per `pyproject.toml`). Copilot's own confidence flag was correct here.

Test count is now 37 (8 new + 25 erasure regression + 4 billing RBAC).

JFK:
**Loop 5 — addressed both findings. Hit max_loops=5; loop exits here.**

✅ **Copilot @ test_stripe_service.py:68**: rewrote `test_run_stripe_does_not_block_event_loop` to use a `threading.Event` coordination instead of wall-clock timing. The new test proves the **causal** property — while the wrapped sync call is blocked in its worker thread, a concurrent asyncio task runs to completion and releases the sync call — rather than the temporal one. If the wrapper blocked the event loop, `allow_sync_finish.wait(timeout=1)` times out and the test fails fast. No flake budget. Test runtime dropped from ~100ms to <10ms. Reconsidered against my earlier ("kept the timing test for stronger semantics") position — the structural test is actually stronger because it asserts causal ordering, not just temporal completion.

✅ **github-code-quality @ test_stripe_service.py:58** (recurring false positive on `await sleep_task`): obsolete after the rewrite — the line no longer exists.

Commit: 66fbff8.

---

### 5-loop summary

| Loop | Action | Verdict |
|---|---|---|
| 1 | Bounded `shutdown_erasure_executor()` with `asyncio.wait_for(30s)` (api/main.py) | superseded |
| 2 | Switched to `wait=False, cancel_futures=True` per Copilot's correct critique that asyncio.wait_for didn't bound the underlying thread; reverted the wait_for wrapping; fixed misleading `stripe_default_timeout` docstring reference | accepted |
| 3 | Updated PR description to match the loop-2 implementation (`wait=False`) | accepted |
| 4 | Added direct unit test for the `StripeError` boundary in `create_checkout_session` (closes the gap QA Lead and Copilot both surfaced) | accepted |
| 5 | Rewrote event-loop test to causal `threading.Event` form, eliminating timing-based flake risk | accepted |

`exit_reason=max_loops` (5/5 reached). The PR is in a clean state — 4 commits ahead of `ca25907`, all tests passing (37 stripe-touching), lint + type-check clean. Further review can be re-entered via `/gh-issue-driven:review` if needed; otherwise this is ready for human review and squash-merge.

Issue #492, opened 2026-04-29
chore(sdk): audit MCP/SDK consumers for Z vs +00:00 datetime parsing compatibility


JFK:
## Background

Issue #489 changed the wire format of the API: naive datetimes now return with `Z` suffix (instead of bare ISO 8601) via `TZAwareBaseModel`. This is technically a **breaking change** for any consumer that depends on the previous (buggy, browser-tz-dependent) behavior. Most date parsers handle both `Z` suffix and `+00:00` offset, but this needs an explicit audit.

### Step 1 — Confirm Python SDK minimum Python version (pre-flight)

Before any grep, check `kagura-memory-python-sdk/pyproject.toml` for the minimum supported Python version:

```bash
grep -E "python\s*=|requires-python" pyproject.toml
```

- **`>=3.11`** → `datetime.fromisoformat("...Z")` works natively. Audit is trivially pass for the `fromisoformat` path; remaining work is to check non-`fromisoformat` parsers (Step 2).
- **`<3.11`** → `fromisoformat("...Z")` raises `ValueError: Invalid isoformat string`. A `.replace("Z", "+00:00")` shim or a different parser is required.

This step gates the depth of Step 2.

### Step 2 — Audit consumers (outbound: API → consumer)

Confirm that the following consumers correctly parse the new `Z`-suffixed wire format:

1. **`kagura-memory-python-sdk`** — Python SDK at https://github.com/kagura-ai/kagura-memory-python-sdk (or local clone). Grep target is broader than `fromisoformat` alone — also check:
   - `datetime.fromisoformat(...)` (Python <3.11 rejects bare `Z`)
   - `dateutil.parser.parse(...)` (handles `Z` reliably)
   - `pydantic.BaseModel` datetime fields (v2 handles `Z`; v1 delegates to `dateutil`)
   - `arrow.get(...)` / `pendulum.parse(...)` (third-party libs)
   - `time.strptime(..., "...%z")` (manual parser; `%z` traditionally rejects bare `Z`, fixed in Python 3.7+)

   ```bash
   grep -rE "(fromisoformat|dateutil|strptime|arrow\.get|pendulum\.parse|pydantic.*datetime)" kagura-memory-python-sdk/
   ```

2. **`kagura-memory-mcp` server** (this repo's MCP tools) — internal Python, not a JSON consumer of its own API. Likely safe but check that any `datetime.fromisoformat` on stringified datetimes from kagura responses (if any) handles `Z`.

[...]

Issue #491, opened 2026-04-29
test(e2e): Playwright with TZ=Asia/Tokyo for JST display Layer 4 coverage


JFK:
## Background

Issue #489 fixed the naive-datetime serialization bug (Z suffix added at API layer). The bug was a **silent CI miss class** — Vitest runs in UTC by default, so frontend tests pass even when the API returns naive datetimes. The bug only surfaces in non-UTC browsers (most users).

This is documented as Layer 4 of the JST debugging checklist:
1. DB user.timezone column
2. /api/v1/auth/me response
3. Frontend code path (formatDateTime)
4. **API datetime Z suffix** ← silent in default Vitest

## Goal

Add a Playwright e2e test that:
1. Forces the browser timezone to a non-UTC zone (e.g., `Europe/London`) so the wire format actually matters at the parser
2. Sets `Profile.timezone = Asia/Tokyo` for the authenticated user via the existing `PUT /api/v1/users/profile` endpoint
3. Navigates to `/workspace/contexts`
4. Asserts the `last_activity_at` tooltip shows JST consistently — derived from the actual UTC ISO 8601 value the API returns (Z-suffixed via `TZAwareBaseModel`), formatted with `Intl.DateTimeFormat(ja-JP, { timeZone: 'Asia/Tokyo', ... })` which mirrors the frontend `formatDateTime` contract.

The wire-format check (`last_activity_at` ends with `Z`) is performed against the API response directly, so the test catches both:
- the dual mode of failure (naive wire format → display drift in non-UTC browser), and
- a silent revert of `TZAwareBaseModel` itself.

## Scope

- One Python pytest-playwright spec file: `backend/tests/e2e/test_datetime_tz_display.py` (the existing e2e harness — there is no TS Playwright in this repo, and `make test-e2e` runs `pytest backend/tests/e2e/`).
- **TZ configuration**: a per-test browser context with `timezone_id="Europe/London"` via `browser.new_context(timezone_id=...)` in a module-scoped fixture local to this file. The shared `authenticated_context` in `backend/tests/e2e/conftest.py` keeps its default TZ — only this test gets the non-UTC override (analogous to per-project TZ in TS Playwright). A global `process.env.TZ` shift would force every other e2e spec into a non-UTC zone and is **out of scope**.
[...]

Issue #490, opened 2026-04-29
feat(db): complete DateTime(timezone=True) migration for ~85 naive columns (Option B from #489)


JFK:
## Background

Issue #489 fixed the JST display bug at the API serialization layer (Option A — Pydantic field_serializer emitting `Z` suffix on naive datetimes). This is the **shorter-term hotfix**. The longer-term clean fix is to make all `Column(DateTime, ...)` columns use `timezone=True`, so the database layer enforces correctness rather than the serialization layer compensating for it.

Audit at the time of #489:
```
$ grep -rn "Column.*DateTime" backend/src/models/*.py | grep -v "timezone=True" | wc -l
85
$ grep -rn "Column.*DateTime" backend/src/models/*.py | grep "timezone=True" | wc -l
9
```

## Why now (well, later)

- **#487** (asyncpg bind type mismatches during migration) showed that even single-column DateTime migrations have non-trivial bind-type compatibility surface. 85x that risk requires a careful staged rollout.
- This work pairs naturally with the **v0.16.0 cost ledger** (#474/#475) which already touches a number of DB schemas — batching alembic friction is more economical.
- The Option A fix in #489 already eliminates the user-visible bug, so this is now a **quality / tech-debt** issue, not a production blocker.

## Scope

1. Migrate every `Column(DateTime, ...)` → `Column(DateTime(timezone=True), ...)` across `backend/src/models/`
2. Author alembic revisions for each table (or batched per logical domain — auth, memory, neural, sleep, resource, llm_pricing, etc.)
3. Audit code paths that compare naive vs aware datetimes (`.utcnow()` callers, `dt < other_dt` comparisons) and update them to use `datetime.now(UTC)` consistently
4. Update `backend/src/utils/datetime.py:utcnow()` to return aware datetimes once the migration completes (the docstring already calls this out)
5. Once Option B is complete, the `TZAwareBaseModel` from #489 can be simplified — naive datetimes will no longer reach Pydantic, so the Z-suffix branch becomes a no-op

## Out of scope (already handled in #489)

- API serialization Z suffix — done via TZAwareBaseModel
- Frontend display — `formatDateTime` already correctly converts UTC ISO → user tz
- `.claude/rules/backend.md` datetime convention note

## Risk

- Bind-type interop with asyncpg (#487-class bugs)
- Implicit `naive == aware` comparison breakage in service-layer code
- Test fixture datetime literals (most are naive — may need fixture updates)

## Acceptance

[...]

Issue #474, opened 2026-04-28
feat(observability): comprehensive LLM cost ledger (recall rerank, future ask())


JFK:
## Summary
A comprehensive cost ledger that captures **every** LLM / embedding / reranker call across the application — not just Sleep Maintenance. Sister issue to the Sleep cost dashboard (#180 → #471/#472/#473) but covers the non-Sleep paths.

Depends on #471 (which establishes the `llm_pricing` table with `unit_type` axis covering tokens AND search-units).

## Why this is separate from #471
The Sleep cost dashboard uses `sleep_reports` as the data source. That captures Sleep Maintenance LLM calls but **misses every other call site**:

- **Recall reranking** (Voyage / Cohere) — `backend/src/services/reranker_service.py` is invoked from `mcp_server/tools/memory.py:236` on the recall path, not the sleep path. Cohere rerank in particular bills per "search unit" ($2.00 / 1k SU), a unit that has no place in `sleep_reports`.
- **Future `ask()` feature** (2026-04-28 brainstorm;派生して Memory Broadlistening が採用された) — will call LLMs outside Sleep.
- **Future synthesis paths** (`/ceo`-style multi-lens, Memory Broadlistening synthesis) — same.
- **One-off admin actions** that call LLMs (e.g. backfill jobs, manual recalibration).

Without a unified ledger, "LLM cost" reports always undercount, and tracking cost regressions is impossible.

### 1. New table: `llm_call_log`
Append-only event log capturing every call. Schema (subject to refinement during impl):
```sql
CREATE TABLE llm_call_log (
  id              BIGSERIAL PRIMARY KEY,
  occurred_at     TIMESTAMPTZ NOT NULL,
  user_id         TEXT,
  workspace_id    UUID,
  context_id      UUID,
  caller          TEXT NOT NULL,    -- 'sleep' | 'recall' | 'rerank' | 'ask' | 'admin' | ...
  call_kind       TEXT NOT NULL,    -- 'completion' | 'embedding' | 'rerank'
  provider        TEXT NOT NULL,
  model           TEXT NOT NULL,
  -- Per-call usage (any subset may be NULL depending on call_kind)
  input_tokens          INT,
  output_tokens         INT,
  cached_input_tokens   INT,
  cache_write_tokens    INT,
  embedding_tokens      INT,
  rerank_tokens         INT,
  rerank_search_units   INT,
  -- Computed at write time (snapshot of pricing at occurred_at), or computed lazily
[...]

PR #498, opened 2026-04-29
fix(api): emit Z suffix on naive datetime serialization (#489)

### LINKED ISSUE #489: fix(api): naive datetime serialization without Z suffix breaks browser timezone parsing
## Symptom

The new `<span title={formatDateTime(...)}>` tooltip from PR #486 renders correctly **only when the user's browser is in UTC timezone**. Users in other timezones (especially JST) see times that are wrong by their local UTC offset.

Concrete reproduction (verified in production 2026-04-29 deploy):

1. Set Profile → Timezone = `Asia/Tokyo`. Verify `/api/v1/auth/me` returns `"timezone": "Asia/Tokyo"`.
2. Hard-reload `/workspace/contexts`.
3. Hover the "最終アクティビティ" cell of any context → see tooltip text.
4. Compare to the actual UTC moment of that event.

For a JST user whose browser tz is also JST: the tooltip shows the **same numeric value as the UTC moment** (e.g., event at `2026-04-28T17:50:22Z` displays as `2026/04/28 17:50:22`), which the user reads as "this is UTC". For a UTC-tz browser, the same event displays as `2026/04/29 02:50:22` (correctly +9h JST). Same backend data, two different displays.

## Root cause

Backend datetime columns are declared `Column(DateTime, ...)` — **without `timezone=True`**. SQLAlchemy treats these as naive, asyncpg returns Python `datetime` objects with `tzinfo=None`, and Pydantic v2 serializes naive datetimes to ISO 8601 strings **without a `Z` suffix**:

```
"last_activity_at": "2026-04-28T17:50:22"
```

ECMAScript `Date` parser interpretation:
- `new Date("2026-04-28T17:50:22Z")` → unambiguous UTC ✓
- `new Date("2026-04-28T17:50:22")` → **parsed in the browser's local timezone** ⚠

So the same response yields different `Date` objects depending on the user's machine timezone. `formatDateTime(date, "Asia/Tokyo", locale)` then produces an output that is correct *if and only if* the browser's local tz happens to match the convention assumed by the API (UTC).

The repo IS already being inconsistent: `backend/src/models/erasure.py` (and a few others) use `Column(DateTime(timezone=True), ...)`, but **85 columns** across the codebase remain naive. Audit:

```
$ grep -rn "Column.*DateTime" backend/src/models/*.py | grep -v "timezone=True" | wc -l
85
```

Examples (non-exhaustive): `Memory.updated_at`, `Memory.created_at`, `User.created_at`, `User.updated_at`, `User.last_login_at`, `Resource.last_run_at`, `LLMPricing.effective_from`, `Context.created_at`, `Context.updated_at`, `SleepReport.started_at` …

## Why PR #486 did not catch this

PR #486 fixed the **frontend** code path (added `<span title={formatDateTime(...)}>`). It did not touch the API serialization. The fix appears to work in test environments where Vitest runs in UTC mode by default — `Date` parsing then matches the API's naive-UTC convention. Production exposes the bug because real users have non-UTC browsers.

### Option A — Pydantic JSON encoder override (smallest blast radius, recommended for v0.14.1)

Add a base-model JSON encoder that emits `Z` suffix for any naive `datetime` (assuming "naive == UTC" convention, which the codebase already follows per `.claude/rules/backend.md` "store UTC, ignore tz metadata"):

```python
# backend/src/models/schemas.py or a new backend/src/api/serializers.py
from datetime import datetime, timezone
from pydantic import BaseModel, field_serializer

class TZAwareBaseModel(BaseModel):
    @field_serializer("*", when_used="json", mode="wrap", check_fields=False)
    def _serialize_naive_datetime_as_utc(self, value, handler):
        if isinstance(value, datetime) and value.tzinfo is None:
            return value.replace(tzinfo=timezone.utc).isoformat().replace("+00:00", "Z")
        return handler(value)
```

Or the more targeted approach — case-by-case `field_serializer` on each datetime field. Less magic but more verbose.

**Pros**: no DB migration, no type changes, single place to fix all serialization.
**Cons**: relies on the "naive == UTC" convention being enforced everywhere (mostly true today; would benefit from a startup assertion).

### Option B — Column type migration to `DateTime(timezone=True)` (correct long-term)

Migrate all 85 naive columns to `DateTime(timezone=True)`. PostgreSQL stores `TIMESTAMP WITH TIME ZONE` (always normalized to UTC internally), asyncpg returns timezone-aware Python `datetime`, Pydantic serializes with `+00:00` offset that JS parses correctly.

**Pros**: type-safe, no convention required, eliminates the entire bug class.
**Cons**: ~85 column migration (multiple alembic revisions), risk of behavior shift in code paths that compare naive-vs-aware datetimes, larger surface for #487-style asyncpg bind-type issues during the migration itself.

### Recommendation

For v0.14.1: **Option A** — single-PR scope, no DB risk, immediately unblocks the JST display bug. File **Option B** as a follow-up cleanup issue (target later milestone, e.g. v0.16.0 alongside the comprehensive cost ledger work).

## Acceptance

- [ ] `/api/v1/contexts` (and any other listing endpoint) returns datetime fields with `Z` suffix or explicit `+00:00` offset
- [ ] `<span title={formatDateTime(date, "Asia/Tokyo", locale)}>` tooltip on `/workspace/contexts` shows JST consistently regardless of browser timezone
- [ ] Verify via two browser environments: a Chrome with system tz = UTC and another with system tz = JST. Both should display the **same JST tooltip** for the same context
- [ ] No regression in existing endpoints — datetime fields still parse with `new Date(...)` on the frontend
- [ ] Vitest test that asserts the API `datetime` JSON encoder always produces a UTC-suffixed string for naive datetime input
- [ ] Add a one-line note in `.claude/rules/backend.md` ("when serializing naive datetimes via Pydantic, ensure the encoder emits `Z` suffix — see #N")

## Out of scope (separate follow-up issues)

- **Option B** — comprehensive `DateTime(timezone=True)` column migration. File as a v0.16.0+ refactor.
- **Dashboard daily-aggregate UTC-bucket cliff** — `formatDate` skips tz conversion for `YYYY-MM-DD` strings (intentional design). For JST users this means daily counts are bucketed on UTC day boundary, which doesn't match their day. Separate design discussion: API should accept a `timezone` query parameter for aggregate endpoints, or frontend should re-bucket. File separately.
- **MCP / SDK consumers** of the API may already be parsing the naive datetime as local time. Adding `Z` suffix is technically a wire-format change. Audit any external SDK consumers (`kagura-memory-python-sdk`, MCP tool callers) to confirm they handle either form. Likely safe (most date parsers accept both), but worth one explicit audit.

## Discovered

While validating the JST tooltip fix from PR #486 immediately after production deploy on 2026-04-29. The backend serialization bug had been present pre-#486 but was masked because the previous code path didn't call `formatDateTime` for the contexts last-activity column at all — it only called `formatRelativeTime` (which ignores timezone arithmetic and relies on relative-duration math that works regardless of parse interpretation).

PR #486 made the bug visible.

## Related

- PR #486 (#483 / #484) — added the tooltip that surfaced this bug
- Issue #175 — original user timezone settings work
- `backend/src/models/erasure.py` — example of correct `DateTime(timezone=True)` usage (Issue #360 / PR #464)

JFK:
Closes #489

## Summary
- Add `TZAwareBaseModel` with a wildcard `field_serializer` calling `to_utc_iso()`, so every response schema's datetime field emits a `Z` suffix (or `+HH:MM` for non-UTC-aware inputs) — JS clients no longer parse naive ISO as local time and JST users see correct timestamps.
- Migrate ~16 response schemas (across `models/schemas.py`, `contexts.py`, `api_keys.py`, `admin.py`, `sleep_reports.py`, `bm25_drift.py`, `workspaces.py`, `me_account.py`, `resource_tokens.py`, `admin_signup_gate.py`, `neural_config.py`) from `BaseModel` to `TZAwareBaseModel`, and remove pre-existing per-field `@field_serializer` / `json_encoders` (single canonical pattern).
- Replace ~30 manual `.isoformat()` call sites in handlers and services with `to_utc_iso()`. Notable side-effect fixes: `oauth.py` previously emitted invalid `+00:00Z` double-suffix for tz-aware UTC; `account_erasure_service.send_erasure_cooling_off_started()` previously sent naive ISO in a privacy-critical email displaying the deletion deadline.
- Add canonical Datetime/UTC convention to `.claude/rules/backend.md`.

## Implementation notes
- `models/api_base.py` (53 lines): `TZAwareBaseModel` uses `@field_serializer("*", mode="wrap", when_used="json", check_fields=False)` to delegate non-datetime fields to Pydantic's default serializer via the wrap handler, applying `to_utc_iso` only to datetime values.
- `to_utc_iso(dt: datetime | None) -> str | None` (already existed in `utils/datetime.py`) is idempotent — handles None/naive/aware-UTC/aware-non-UTC uniformly. Safe to apply unconditionally in handler/service code.
- Per-field serializers were removed from `MemoryResponse`, `LinkedMemoryRef`, `ReferenceResponse`, `UserWithAdminFlag`, `UserInfo` (admin.py), `SleepReportSummary`, `SleepActionItem`, `Bm25DriftSummary`. The new wildcard handles all of them.
- DB columns are unchanged (still TIMESTAMP WITHOUT TIME ZONE / naive UTC). The full column-type migration to `DateTime(timezone=True)` is deferred to follow-up #490 targeting v0.16.0 alongside the cost ledger work (#474/#475).
- Wire-format change: `+00:00` -> `Z` for tz-aware UTC datetimes in oauth.py and admin.py user detail. Both forms are valid ISO 8601; JS `new Date()` parses both. SDK / MCP consumer audit captured in follow-up #492.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green — privacy UX improvement (erasure deadline display) + oauth double-Z bug fix are security-positive side effects; no new attack surface
- qa-lead: green — schema-level coverage sufficient for v0.14.1 hotfix; SDK smoke test (2 min) recommended before v0.14.1 release tag (part of #492)
- cto: green — Option A -> Option B sequencing is correct (#490 v0.16.0); inheritance vs Annotated trade-off accepted (perf delta < latency budget); rule placement in `.claude/rules/backend.md` is canonical
- gate1: green via /claude-c-suite:ask (CTO lens)
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/489-fix-fix-api-naive-datetime-serialization-wit.gate1.md
- ~/.claude/cache/gh-issue-driven/489-fix-fix-api-naive-datetime-serialization-wit.gate2.md

## Follow-up issues (filed)
- #490 v0.16.0+ — feat(db): complete `DateTime(timezone=True)` migration for ~85 naive columns (Option B)
- #491 v0.14.1 — test(e2e): Playwright with `TZ=Asia/Tokyo` for JST display Layer 4 coverage
- #492 v0.14.1 — chore(sdk): audit MCP/SDK consumers for Z vs `+00:00` datetime parsing compatibility

## Test plan
- [x] `make lint` — All checks passed (ruff)
- [x] `make type-check` — 0 errors, 0 warnings (pyright)
- [x] `make test-local` — 1664 passed (35 new + 1629 pre-existing). 4 pre-existing Qdrant integration failures unrelated (verified by stashing on origin/main).
- [ ] Manual SDK smoke test before v0.14.1 release tag (part of #492 verification)
- [ ] Browser smoke test: `/workspace/contexts` tooltip displays JST consistently in both UTC-tz and JST-tz Chrome

### DIFF
--- a/.claude/rules/backend.md
+++ b/.claude/rules/backend.md
@@ -24,6 +24,11 @@ paths:
 - Synchronous engine only for OAuth2 server (Authlib requirement)
 - Pool config: `pool_size=5, max_overflow=10, pool_pre_ping=True`
 
+## Datetime / UTC
+- DB columns are TIMESTAMP WITHOUT TIME ZONE — naive UTC by convention. Until #490 migrates them, the API layer is the only place ensuring wire-format datetimes are unambiguous (#489).
+- Response Pydantic schemas with any `datetime` field MUST inherit from `models.api_base.TZAwareBaseModel` (not `BaseModel`). The base class serializes naive datetimes with a `Z` suffix; without it, JS clients parse the string as local time and JST users see times shifted by their UTC offset.
+- For `str`-typed datetime fields in handler/service code (manual ISO formatting), use `to_utc_iso(dt)` from `utils.datetime`. Never call `.isoformat()` directly on a `datetime` field that ends up in a JSON response or cached payload — `to_utc_iso` is idempotent (handles None / naive / aware UTC / aware non-UTC) so it can be applied unconditionally.
+
 ## Testing
 - Test files: `backend/tests/{module}/test_{name}.py`
 - Use `pytest-asyncio` fixtures from `conftest.py`
--- a/backend/src/api/routes/admin.py
+++ b/backend/src/api/routes/admin.py
@@ -14,6 +14,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth.dependencies import require_admin
 from db.base import get_db
+from models.api_base import TZAwareBaseModel
 from models.auth import (
     APIKey,
     Context,
@@ -25,6 +26,7 @@ from models.auth import (
 )
 from models.memory import Memory
 from utils import db_transaction, get_user_id
+from utils.datetime import to_utc_iso
 from utils.logger import get_logger
 
 logger = get_logger(__name__)
@@ -37,20 +39,12 @@ router = APIRouter(prefix="/admin", tags=["admin"])
 # ============================================================================
 
 
-class UserInfo(BaseModel):
+class UserInfo(TZAwareBaseModel):
     """User information for admin view.
 
     Issue #164: Extended with workspaces field.
     Issue #175: Added timezone field for user preferences.
     """
-
-    model_config = {
-        "json_encoders": {
-            datetime: lambda v: (
-                v.isoformat() + "Z" if v and v.tzinfo is None else v.isoformat() if v else None
-            )
-        }
-    }
 
     id: str
     email: str
@@ -258,7 +252,7 @@ async def list_users(
                         "workspace_name": workspace.name,
                         "role": member.role,
                         "is_primary": is_primary,
-                        "joined_at": member.joined_at.isoformat() if member.joined_at else None,
+                        "joined_at": to_utc_iso(member.joined_at),
                     }
                 )
 
@@ -519,16 +513,8 @@ async def get_user_detail(
                 "picture": target_user.picture,
                 "role": target_user.role,
                 "is_initial_admin": getattr(target_user, "is_initial_admin", False),
-                "created_at": target_user.created_at.isoformat() + "Z"
-                if target_user.created_at.tzinfo is None
-                else target_user.created_at.isoformat(),
-                "last_login_at": (
-                    target_user.last_login_at.isoformat() + "Z"
-                    if target_user.last_login_at.tzinfo is None
-                    else target_user.last_login_at.isoformat()
-                )
-                if target_user.last_login_at
-                else None,
+                "created_at": to_utc_iso(target_user.created_at),
+                "last_login_at": to_utc_iso(target_user.last_login_at),
                 "auth_provider": target_user.auth_provider,
             },
             workspaces=workspaces,
--- a/backend/src/api/routes/admin_plans.py
+++ b/backend/src/api/routes/admin_plans.py
@@ -24,7 +24,7 @@ from models.auth import Context, PlanChange, User, Workspace, WorkspaceInvitatio
 from models.memory import Memory
 from services.effective_quota_service import EffectiveQuotaService
 from utils import db_transaction
-from utils.datetime import utcnow
+from utils.datetime import to_utc_iso, utcnow
 from utils.logger import get_logger
 
 logger = get_logger(__name__)
@@ -375,7 +375,7 @@ async def get_plan_change_audit(
                     old_plan=audit.old_plan,
                     new_plan=audit.new_plan,
                     changed_by=user_name or audit.changed_by,
-                    changed_at=audit.changed_at.isoformat(),
+                    changed_at=to_utc_iso(audit.changed_at) or "",
                     reason=audit.reason,
                 )
             )
--- a/backend/src/api/routes/admin_signup_gate.py
+++ b/backend/src/api/routes/admin_signup_gate.py
@@ -19,6 +19,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth.dependencies import AdminUser
 from db.base import get_db
+from models.api_base import TZAwareBaseModel
 from services.signup_gate_service import SignupGateService
 from utils.github_user import GitHubUserNotFound
 from utils.logger import get_logger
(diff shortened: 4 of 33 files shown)


JFK:
Loop 1: addressed 1 actionable Copilot suggestion (api_base.py module docstring corrected to match the actual wire format — `Z` for UTC, `+HH:MM` for non-UTC offsets, since `to_utc_iso()` always normalizes `+00:00` to `Z`). Re-requesting review.

JFK:
Loop 2: addressed 1 actionable Copilot suggestion (replaced `pydantic_core.core_schema.SerializerFunctionWrapHandler` internal-API import with public `collections.abc.Callable[[Any], Any]`). Re-requesting review.

PR #486, opened 2026-04-29
fix(frontend): MFA Enter submit + last-activity tz tooltip (#483, #484)

JFK:
Closes #483
Closes #484

## Summary

- **#484** — MFA login form now submits with **Enter** on a 6-digit TOTP entry. The submit button is conditionally disabled until 6 digits are present, which can suppress browser implicit form submit; the new explicit `onKeyDown` handler closes the gap.
- **#483** — Last-Activity column on `/workspace/contexts` now shows a **timezone-aware absolute datetime tooltip** on hover. The relative string ("5 minutes ago") stays as the primary surface — relative time is timezone-independent by definition; absolute moments belong in the tooltip.
- Drop the **silently-ignored `timezone` argument** from `formatRelativeTime` (a long-standing signature lie that masked the missing absolute-time pathway). Sweep all 13 internal call sites; remove now-unused `useAuth()` / `user` / `authUser` declarations from 5 files.

### #484 — `frontend/src/app/login/page.tsx`
- Extract `submitMfa()` (event-less) from old `handleMfaVerify`. Form `onSubmit` and the new `onKeyDown` both call it — no `as unknown as React.FormEvent` cast needed.
- `onKeyDown` triple-guards on `e.key === "Enter"` + `totpCode.length === 6` + `loadingAction === null` to prevent partial-input misfire and double-submit while in flight.
- `frontend/src/app/login/page.test.tsx` (new): 3 cases — happy path, under-6-digit no-op, in-flight no-op.

### #483 — `frontend/src/lib/utils/datetime.ts`
- Old: `formatRelativeTime(utcTime, timezone="UTC", locale="en", addSuffix=true)`
- New: `formatRelativeTime(utcTime, locale="en", addSuffix=true)`
- `frontend/src/lib/utils/datetime.test.ts` (new): 10 cases covering JST/EDT/UTC formatting, date-only passthrough, and locale + suffix behavior.
- Tooltip pattern in `workspace/contexts/page.tsx` mirrors the existing convention in `IndexerStatusPanel.tsx:154-158` and `ResourceStatsStrip.tsx:59`.

### Sweep cleanup
[...]

PR #505, opened 2026-04-29
test(e2e): JST display Layer 4 regression coverage (#491)

### LINKED ISSUE #491: test(e2e): Playwright with TZ=Asia/Tokyo for JST display Layer 4 coverage

JFK:
Closes #491

## Summary

- Adds a Python pytest-playwright e2e test (`backend/tests/e2e/test_datetime_tz_display.py`) that catches Layer 4 of the JST debugging checklist — the silent CI miss class that lives in non-UTC browsers and is invisible to UTC-default Vitest.
- The test forces the browser timezone to `Europe/London` via a per-test context, sets `Profile.timezone = Asia/Tokyo` via `PUT /api/v1/users/profile`, asserts the wire format ends with `Z`, derives the expected JST display from the actual API value (no `now()`-based assertions), and matches it against the rendered tooltip on `/workspace/contexts`.
- Refactors `backend/tests/e2e/conftest.py` to extract the admin-login flow into a `login_admin(page)` helper used by both fixtures — eliminates ~25 lines of byte-for-byte duplication.
- Inverse-test verified manually before commit: temporarily setting `ContextResponse(BaseModel)` at `backend/src/api/routes/contexts.py:191` makes the test fail with the wire-format diagnostic; reverting to `TZAwareBaseModel` restores PASS.

## Implementation notes

- **TZ override**: `browser.new_context(timezone_id="Europe/London")` (Playwright Python's per-context override) — analogous to per-project TZ in TS Playwright. Only this fixture overrides; the rest of the e2e suite keeps its default TZ.
- **Profile.timezone setup**: API path (`PUT /api/v1/users/profile`) using the existing admin session cookie. DB-seed alternative was considered but rejected — the API path is what production exercises and avoids coupling the e2e test to internal user-table schema.
- **Tooltip locator**: anchors on the exact JST string via `span[title="<expected_jst>"]`. Second-precision makes the formatted timestamp effectively row-unique, so no `data-context-id` attribute is needed on the row.
- **Skip behavior**: when the admin workspace has no context with `last_activity_at`, the test `pytest.skip()`s — a test-data condition, not a regression. Worth monitoring CI skip rate so this doesn't silently turn the protection off.
- **Issue body deviation**: the original body referenced `frontend/playwright.config.ts` and `frontend/e2e/datetime-tz-display.spec.ts` (TS Playwright). This repo has Python pytest-playwright in `backend/tests/e2e/`, no TS Playwright at all. Body was updated mid-implementation to reflect actual setup; the gate1 reviewer's recommendations (per-project TZ, fixed timestamp, parser path documentation) were translated to the Python equivalent.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: yellow → effective green via /ask (QA Lead lens; 4 adjustments applied to issue body)
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/491-feat-test-e2e-playwright-with-tz-asia-tokyo-f.gate1.md
- ~/.claude/cache/gh-issue-driven/491-feat-test-e2e-playwright-with-tz-asia-tokyo-f.gate2.md

## Companion / related

- #492 (companion SDK consumer audit — closed 2026-04-29 as ALL CLEAR via comment)
- #489 (the Z-suffix wire-format fix that this test guards)
- #486 (the PR that exposed the bug by adding the tooltip)

### DIFF
--- a/backend/tests/e2e/conftest.py
+++ b/backend/tests/e2e/conftest.py
@@ -9,7 +9,6 @@ Usage:
 
 import os
 import re
-import time
 
 import pytest
 from playwright.sync_api import Page, sync_playwright
@@ -46,15 +45,22 @@ def login_admin(page: Page) -> None:
     if admin_link.count() > 0:
         admin_link.first.click()
         page.wait_for_load_state("networkidle")
-        time.sleep(1)
 
+    # Wait deterministically for the login form input to be present —
+    # replaces the previous time.sleep(1) hard-pause after the admin-link
+    # click (which only fires on first login but is shared across every
+    # authenticated_context variant).
+    page.wait_for_selector('input[type="text"]', state="visible", timeout=10000)
     page.fill('input[type="text"]', ADMIN_LOGIN_ID)
     page.fill('input[type="password"]', ADMIN_PASSWORD)
 
     terms_checkbox = page.locator('input[type="checkbox"], button[role="checkbox"]')
     if terms_checkbox.count() > 0:
         terms_checkbox.first.click()
-        time.sleep(0.3)
+        # No explicit wait — the next click('button[type="submit"]') auto-waits
+        # for the submit button to be enabled (Playwright actionability check),
+        # which implicitly waits for the checkbox click to flip the form into
+        # a submittable state.
 
     page.click('button[type="submit"]')
     page.wait_for_url("**/workspace/**", timeout=15000)
--- a/backend/tests/e2e/conftest.py
+++ b/backend/tests/e2e/conftest.py
@@ -29,17 +29,17 @@ def browser():
         browser.close()
 
 
-@pytest.fixture(scope="session")
-def authenticated_context(browser):
-    """Create an authenticated browser context (login once per session)."""
-    context = browser.new_context(viewport={"width": 1280, "height": 800})
-    page = context.new_page()
+def login_admin(page: Page) -> None:
+    """Perform the admin login flow on the given page.
 
-    # Navigate to login page
+    Shared by every authenticated_context fixture variant so login UI
+    changes (selectors, terms checkbox, redirect target) only need updating
+    here. Terminates on the post-login `**/workspace/**` redirect; callers
+    own the page lifecycle around it.
+    """
     page.goto(f"{BASE_URL}/login")
     page.wait_for_load_state("networkidle")
 
-    # Click admin login link if present
     admin_link = page.locator(
         "a, button", has_text=re.compile(r"管理者ログイン|Admin Login|管理者")
     )
@@ -48,20 +48,24 @@ def authenticated_context(browser):
         page.wait_for_load_state("networkidle")
         time.sleep(1)
 
-    # Fill login form
     page.fill('input[type="text"]', ADMIN_LOGIN_ID)
     page.fill('input[type="password"]', ADMIN_PASSWORD)
 
-    # Check terms checkbox if present
     terms_checkbox = page.locator('input[type="checkbox"], button[role="checkbox"]')
     if terms_checkbox.count() > 0:
         terms_checkbox.first.click()
         time.sleep(0.3)
 
-    # Submit and wait for redirect
     page.click('button[type="submit"]')
     page.wait_for_url("**/workspace/**", timeout=15000)
 
+
+@pytest.fixture(scope="session")
+def authenticated_context(browser):
+    """Create an authenticated browser context (login once per session)."""
+    context = browser.new_context(viewport={"width": 1280, "height": 800})
+    page = context.new_page()
+    login_admin(page)
     page.close()
     yield context
     context.close()
--- a/backend/tests/e2e/test_datetime_tz_display.py
+++ b/backend/tests/e2e/test_datetime_tz_display.py
@@ -96,8 +96,10 @@ def test_last_activity_tooltip_displays_jst_under_non_utc_browser_tz(jst_page: P
     - `TZAwareBaseModel` is removed or bypassed → the wire-format assertion
       in step 3 fails before the UI assertion runs, with a precise diagnostic.
 
-    Skips when the admin workspace has no context with `last_activity_at` set
-    (a test-data condition, not a regression).
+    Self-sufficient on clean DB volumes: if no context yet has
+    `last_activity_at`, seeds a memory via `POST /api/v1/memory/remember`
+    and re-fetches before asserting. Skips only when the workspace has
+    zero contexts (environmental — rarer than the data-empty case).
     """
     profile_response = jst_page.request.put(
         f"{API_URL}/api/v1/users/profile",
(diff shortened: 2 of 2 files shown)


JFK:
Loop 1: addressed 1 actionable item.

- Replaced the unconditional skip on missing `last_activity_at` with an inline seed via `POST /api/v1/memory/remember` against the first existing context, then re-fetched `/api/v1/contexts` so the assertions actually run on a clean DB volume. The skip path is preserved only for the rarer case of zero contexts.

Pushed as 4e0349c1.

JFK:
Loop 2: addressed 2 actionable items.

- **Locale dependency**: passed `locale="ja"` to `browser.new_context()` so `navigator.language=ja` and the i18n provider (`frontend/src/i18n/config.ts:30-41`) selects ja, making `formatDateTime` render the expected `YYYY/MM/DD HH:MM:SS`.
- **Brittle `time.sleep(1)`**: replaced with `wait_for_selector("table span[title]", state="attached", timeout=15000)` — deterministic wait on the actual DOM condition.

Pushed as 8bca2d6d.

JFK:
Loop 3 review (2 comments at lines 111 and 145): non-actionable — disagreement on premise.

The suggestion to switch from `data={...}` to `json={...}` does not apply to Playwright Python:

```
$ python3 -c "...inspect.signature(ctx.request.post)..."
post params: ['url', 'params', 'headers', 'data', 'form', 'multipart', 'timeout', 'fail_on_status_code', 'ignore_https_errors', 'max_redirects']
put params:  ['url', 'params', 'headers', 'data', 'form', 'multipart', 'timeout', 'fail_on_status_code', 'ignore_https_errors', 'max_redirects']
```

There is no `json` parameter in `APIRequestContext.post/put` in Playwright Python — the `json:` option exists in the JS/TS API but not the Python one. The canonical way to send JSON in Playwright Python is `data=<dict>`, which the library auto-serializes to JSON and sets `Content-Type: application/json` on. Concrete proof: the test currently passes (4.04s), which would be impossible if FastAPI / Pydantic were receiving form-encoded payloads — they would return 422 and `assert response.ok` would fail.

Refs: Playwright Python source, [`APIRequestContext.post`](https://playwright.dev/python/docs/api/class-apirequestcontext#api-request-context-post) — "if the data parameter is an object, it will be serialized to json string and content-type header will be set to application/json".

No code change. Loop 3 closes as non-actionable disagreement.

JFK:
Loop 4: addressed 2 actionable items.

- **conftest.py time.sleep removal**: `time.sleep(1)` → `wait_for_selector('input[type="text"]', state="visible", timeout=10000)` (now runs unconditionally, deterministic). `time.sleep(0.3)` after terms checkbox: removed — Playwright's auto-wait on the next submit-click handles actionability. Dropped `import time`.
- **non-waiting count() check**: `count() > 0` on `span[title="<expected>"]` → `tooltip_locator.first.wait_for(state="attached", timeout=15000)` wrapped in try/except so the rich diagnostic message is preserved on timeout.

Verified: new test 2.77s (faster), `test_resources_page` smoke tests still pass, ruff clean.

Pushed as 1ca02032.

PR #518, opened 2026-04-30
fix(auth): loopback DCR + RFC 6749 errors + PKCE enforcement (#513)

JFK:
Closes #513

## Summary

- Rewrite DCR provider detection (`detect_dcr_provider`) to accept RFC 8252 loopback redirects (`http://localhost`, `http://127.0.0.1`, `http://[::1]`) with NFKC-normalized client_name keyword fallback. Hostname suffix match for the existing claude.ai / chatgpt.com / cursor.sh patterns also closes the substring-spoof hole (`https://attacker.com/?fake=chatgpt.com`).
- Convert DCR rejection responses to RFC 6749 §5.2 envelope (`{"error", "error_description"}`) via a new `utils/oauth_errors.py` helper so OAuth-conformant SDKs can parse them — Claude Code's Anthropic SDK was previously throwing `ZodError` on the FastAPI default `{"detail": ...}` shape.
- Register Authlib's `CodeChallenge(required=settings.oauth_pkce_required)` extension on `AuthorizationCodeGrant`. Issue #157 added `token_endpoint_auth_method="none"` support and PKCE plumbing but never registered the extension, leaving the gate unenforced — public clients could exchange auth codes without `code_verifier`. The kill-switch `OAUTH_PKCE_REQUIRED` env var allows emergency rollback.
- Pre-existing latent bug fix (bundled because it blocks AC #1 in production): `OAuth2ClientResponse.owner_id: str | None` so DCR-registered clients (`owner_id=None`) don't 500 at response validation.

### Provider detection
- `urlparse`-based hostname matching replaces the old `"chatgpt.com" in redirect_uri` substring check. Hostname suffix match (`host == s or host.endswith("." + s)`) handles subdomains like `https://app.claude.ai/...` and structurally rejects spoofs like `https://attacker.com/?fake=chatgpt.com`.
- `_normalize_client_name` runs NFKC followed by Cf/Cc category strip. NFKC alone collapses fullwidth (`Ｃｌａｕｄｅ` → `Claude`) but leaves zero-width / format characters intact. Stripping the `Cf` and `Cc` categories before lowercase substring match neutralizes both fullwidth homoglyphs and ZWSP-injection attacks.
- `_LOOPBACK_PROVIDER_KEYWORDS = frozenset(_PROVIDER_HOSTNAMES)` — derived from the hostname mapping so adding a provider only requires editing one dict.

### RFC 6749 error envelope
[...]

PR #520, opened 2026-04-30
fix(auth): drop NOT NULL on oauth_clients.owner_id (#519)

### LINKED ISSUE #513: fix(oauth): support loopback redirects in DCR + RFC 6749 error format
## Summary

`backend/src/api/routes/oauth.py:512-535` detects the OAuth provider purely by substring-matching `redirect_uris[0]` against a fixed allowlist (`claude.ai`, `anthropic.com`, `chatgpt.com`, `chat.openai.com`, `cursor.sh`, `cursor.com`). Native OAuth clients that follow [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252) — including Claude Code CLI, Cursor's CLI, and any other "OAuth 2.0 for Native Apps" client — register with **loopback redirects** (`http://localhost:<port>/callback`, `http://127.0.0.1:<port>/callback`). These never match the substring patterns, so `detected_provider` falls through to `"custom"` and the request is rejected with `HTTP 400 "Dynamic client registration is only allowed for: chatgpt, claude, cursor"`.

A second related issue: the rejection response body is FastAPI's default `{"detail": "..."}`, but RFC 6749 §5.2 / RFC 7591 §3.2.2 require OAuth error responses to use `{"error": "...", "error_description": "..."}`. Claude Code's OAuth SDK parses the response with a Zod schema expecting RFC-compliant fields and surfaces a confusing `ZodError` instead of the actual rejection reason.

**Scope note**: both bugs originate in the same `dynamic_client_registration` handler. Fixing one without the other still leaves the user experience broken — a single PR is the right granularity, not two atomic commits.

## Repro

```
claude mcp add --transport http kagura-foo https://memory.kagura-ai.com/mcp/w/<workspace_id>
```

On first tool invocation the CLI starts an OAuth flow with a loopback redirect, calls `POST /api/v1/oauth/register`, and the server returns 400. The user sees:

```
SDK auth failed: HTTP 400: Invalid OAuth error response: ZodError: [
  { "expected": "string", "code": "invalid_type", "path": ["error"], ... }
]. Raw body: {"detail":"Dynamic client registration is only allowed for: chatgpt, claude, cursor"}
```

`claude.ai`-managed connections (synced to Claude Code CLI via the `user:mcp_servers` OAuth scope) are unaffected — their redirect_uri lives on `claude.ai` and matches the existing allowlist. The breakage is specific to users who register the MCP server directly via `claude mcp add` (or any equivalent RFC 8252 native-app flow).

## Security note

Adding loopback acceptance widens the DCR trust signal from *redirect_uri pattern* to *redirect_uri pattern + client_name substring*. `client_name` is a user-supplied field and MUST be treated as a soft signal only — security continues to rely on the existing rate limit (5 registrations/min/IP), redirect_uri pattern validation, and the `token_endpoint_auth_method="none"` default for public clients, not on `client_name` authenticity. No new secret, scope, or session-handling change is introduced.

## Out of scope

- OAuth2 access/refresh token issuance and lifetime
- OAuth scope grants and `memory:read` / `memory:write` semantics
- Session cookie / API key authentication paths
- The `/api/v1/oauth/clients/*` admin-managed client creation flow

## Acceptance Criteria

- [ ] Loopback redirect URIs (`http://localhost[:port][/path]`, `http://127.0.0.1[:port][/path]`) are accepted by `dynamic_client_registration` when the request carries a recognized `client_name` (case-insensitive substring match for `claude`, `cursor`, `chatgpt`).
- [ ] DCR rejection responses follow RFC 6749 §5.2 / RFC 7591 §3.2.2: `{"error": "invalid_client_metadata", "error_description": "..."}`.
- [ ] Existing `claude.ai` / `anthropic.com` / `chatgpt.com` / `chat.openai.com` / `cursor.sh` / `cursor.com` redirect_uri detection continues to work (regression coverage).
- [ ] Rate limit (5 registrations/min/IP), redirect_uri pattern validation, and `token_endpoint_auth_method="none"` default are unchanged.
- [ ] New tests cover: (a) loopback + recognized client_name → accepted, (b) loopback + unknown client_name → still rejected with RFC 6749 format, (c) non-loopback custom URL → still rejected with RFC 6749 format, (d) each existing pattern still accepted.

## Post-deploy verification

- [ ] `claude mcp add --transport http kagura-test https://memory.kagura-ai.com/mcp` followed by a tool invocation completes the OAuth dance end-to-end against the deployed server.

## References

- Source: `backend/src/api/routes/oauth.py:458-535`
- RFC 8252 — OAuth 2.0 for Native Apps (loopback redirects)
- RFC 7591 — OAuth 2.0 Dynamic Client Registration Protocol
- RFC 6749 §5.2 — OAuth error response format
- Related (closed): #207 (wildcard redirect_uri patterns for claude.ai/chatgpt dynamic callbacks), #182 (ChatGPT DCR investigation)

### LINKED ISSUE #519: fix(auth): DB constraint misalignment — oauth_clients.owner_id NOT NULL blocks DCR (#513 follow-up)
## Summary

PR #518 (closes #513) widened `OAuth2ClientResponse.owner_id` to `str | None` so DCR-registered clients (which set `owner_id=None`) can serialize. However, the underlying database constraint and the `OAuth2Client` ORM column are still `NOT NULL`, so any successful DCR INSERT immediately fails:

```
psycopg2.errors.NotNullViolation: null value in column "owner_id" of relation "oauth_clients" violates not-null constraint
```

The 500 is then surfaced as a FastAPI default `{"detail": "Failed to register client: ..."}` envelope (the catch-all 500 path was intentionally NOT changed to RFC 6749 in #513). Net effect: DCR rejection paths work end-to-end with the new RFC 6749 envelope, but DCR accept path is still broken — `claude mcp add --transport http kagura-test http://localhost:8080/mcp` cannot complete OAuth registration on a self-hosted instance.

This was missed by #513's tests because the test suite mocks `get_sync_session`, so the DB-level constraint never fires.

## Repro

```bash
# Run a fresh kagura-memory-cloud locally (after PR #518 merge), then:
curl -X POST http://localhost:8080/api/v1/oauth/register \
  -H "Content-Type: application/json" \
  -d '{"client_name":"Claude Code","redirect_uris":["http://localhost:54321/callback"],"token_endpoint_auth_method":"none"}'
```

Expected: HTTP 201 with the new client record.
Actual: HTTP 500 `{"detail":"Failed to register client: (psycopg2.errors.NotNullViolation) null value in column \"owner_id\" ..."}`.

## Scope

Three layers need to align so DCR-registered clients (`owner_id=None`) work end-to-end:

| Layer | Current state | Target state |
|---|---|---|
| `OAuth2ClientResponse.owner_id` (Pydantic, response model) | `str \| None` ✓ | `str \| None` (already correct, #513) |
| `OAuth2Client.owner_id` (SQLAlchemy column) | `Column(String(255), nullable=False, index=True)` | `nullable=True` |
| `oauth_clients.owner_id` (DB column) | `character varying(255) NOT NULL` | nullable |

## Acceptance Criteria

- [ ] `backend/src/models/auth.py` — `owner_id` column declared with `nullable=True`.
- [ ] New Alembic migration in `backend/alembic/versions/` that `ALTER COLUMN owner_id DROP NOT NULL` on `oauth_clients`. Reversible (`upgrade`/`downgrade` both implemented). The downgrade must `ALTER COLUMN ... SET NOT NULL` and document that any existing NULL rows (DCR-registered clients) will block the downgrade — that is the correct fail-loud behavior.
- [ ] Manual smoke: `claude mcp add --transport http kagura-test http://localhost:8080/mcp` followed by a tool invocation completes the full OAuth dance against a self-hosted instance with the migration applied.
- [ ] New integration test that POSTs to `/api/v1/oauth/register` with a real test DB session (not mocked) and asserts a 201 + non-null `client_id` in the response. This pins the layer-alignment regression.
- [ ] Existing `test_oauth_dcr.py` tests continue to pass.
- [ ] Verify admin-managed `POST /api/v1/oauth/clients` (which sets `owner_id=user.id`) still produces 201 — the constraint relaxation must not break the existing admin path.

## Out of scope

- DCR endpoint logic (already covered by #513).
- Other OAuth2 routes (token, authorize, revoke, introspect).
- Workspace scoping (`workspace_id` is already nullable; nothing to change).

## Why this slipped past #513

The DCR test suite (`tests/api/test_oauth_dcr.py`) mocks `api.routes.oauth.get_sync_session` and `utils.encryption.get_encryptor` so the OAuth2Client INSERT never reaches the DB engine. The `model_dump()`/`response_model=OAuth2ClientWithSecretResponse` validation passed because the Pydantic-level Optional change was correct. The constraint mismatch only surfaces when an actual DB session attempts the INSERT.

This is a structural test-coverage gap that the new integration test (AC item 4) closes.

## References

- PR #518 (merge of #513): https://github.com/kagura-ai/memory-cloud/pull/518
- Issue #513 (parent): https://github.com/kagura-ai/memory-cloud/issues/513
- `backend/src/models/auth.py:344` (column declaration)
- `backend/src/api/routes/oauth.py:565` (DCR sets `owner_id=None`)
- `backend/src/api/routes/oauth.py:108` (response model — already Optional)

JFK:
Closes #519

## Summary

PR #518 (closes #513) shipped DCR loopback support but the **DCR accept path was still broken in production** — the Pydantic response model was fixed (`owner_id: str | None`) but the underlying SQLAlchemy column and DB constraint were still `NOT NULL`. Any successful DCR INSERT immediately failed with `psycopg2.errors.NotNullViolation`.

This PR aligns the three layers so DCR-registered clients (`owner_id=None`) work end-to-end:

| Layer | Before | After |
|---|---|---|
| `OAuth2ClientResponse.owner_id` (Pydantic) | `str \| None` ✓ (#513) | `str \| None` (unchanged) |
| `OAuth2Client.owner_id` (SQLAlchemy column) | `Column(String(255), nullable=False, index=True)` | `nullable=True` |
| `oauth_clients.owner_id` (DB column) | `character varying(255) NOT NULL` | nullable (migration) |

## Why this slipped past #513

The DCR test suite in `tests/api/test_oauth_dcr.py` mocks `api.routes.oauth.get_sync_session` and `utils.encryption.get_encryptor`, so the OAuth2Client INSERT never reaches the DB engine. The `model_dump()` / `response_model=OAuth2ClientWithSecretResponse` validation passed because the Pydantic-level Optional change was correct. The constraint mismatch only surfaces when an actual DB session attempts the INSERT — which the new integration test in this PR exercises.

## Local validation

Reproduced with `claude mcp add --transport http kagura-test http://localhost:8080/mcp` against a fresh kagura-memory-cloud (post-#518 merge):

```
HTTP/1.1 500 Internal Server Error
{"detail":"Failed to register client: (psycopg2.errors.NotNullViolation) null value in column \"owner_id\" of relation \"oauth_clients\" violates not-null constraint\n..."}
```

Applied the migration in this PR (`d04_519_oauth_owner_nullable`), restarted nothing (the model change is pure Pydantic/SQLAlchemy semantics, no behavior shift in the route handler), and re-ran the same call:

```
HTTP/1.1 201 Created
{"owner_id": null, "provider": "claude", "client_id": "oauth_...", ...}
```

All three DCR scenarios verified: `http://localhost:54321/callback`, `http://[::1]:54321/cb`, and `https://chatgpt.com/connector_platform_oauth_redirect`.

## Migration semantics

`d04_519_oauth_owner_nullable` is reversible:
- `upgrade()`: `ALTER COLUMN owner_id DROP NOT NULL`.
- `downgrade()`: `ALTER COLUMN owner_id SET NOT NULL`. **Will fail loudly if any rows have `owner_id IS NULL`** (DCR-registered clients). Operators must either delete those rows or backfill an owner before downgrading — that's the correct fail-loud behavior, not a bug.

The migration follows the project's revision-id 32-char cap convention (`alembic_version.version_num` is `VARCHAR(32)`). Filename is longer; revision id is `d04_519_oauth_owner_nullable` (29 chars).

## Tests

New integration test class `TestDcrLoopbackPersistsNullOwnerId` in `backend/tests/integration/test_oauth_dcr_integration.py` covers both shapes against a real Postgres test session:

1. **DCR loopback round-trip** — `POST /api/v1/oauth/register` with `client_name="IntegrationTest Claude Code Loopback"` + `redirect_uris=["http://localhost:54321/callback"]` returns 201, the response carries `owner_id: null`, and the DB row also has `owner_id IS NULL`. This is the regression pin — without the migration, this test fails with the 500 reproduced above.
2. **Admin path with non-null owner_id** — direct ORM insert of `OAuth2Client(owner_id="user-...", ...)` still succeeds, pinning that the constraint relaxation does not regress the admin-managed path.

Sweep: 128/128 OAuth-adjacent tests pass (`test_oauth_dcr` + `test_oauth_pkce_enforcement` + new `test_oauth_dcr_integration` + `test_oauth_authorize_template` + `test_oauth_authorize_redirect_uri_precheck` + `test_redirect_uri` + `test_well_known`).

The unrelated `test_alembic_migrations.py::TestB03NeuralEdgesBackfillMigration` failures are pre-existing on `main` — verified by running the same test from the base branch.

## Out of scope

- DCR endpoint logic (already covered by #513).
- Other OAuth2 routes (token, authorize, revoke, introspect).
- Workspace scoping (`workspace_id` is already nullable).

🤖 Generated via /gh-issue-driven (519 follow-up)

### DIFF
--- a/backend/alembic/versions/d04_519_oauth_clients_owner_id_nullable.py
+++ b/backend/alembic/versions/d04_519_oauth_clients_owner_id_nullable.py
@@ -24,15 +24,20 @@ Downgrade re-applies ``SET NOT NULL``. Any existing rows with
 that is the correct fail-loud behavior; the operator must either delete
 the DCR rows or backfill an owner before downgrading.
 
-Revision ID: d04_519_oauth_clients_owner_id_nullable
+Revision ID: d04_519_oauth_owner_nullable
 Revises: c03_471_seed_pricing
 
 NOTE: Revision IDs are capped at 32 chars because
 ``alembic_version.version_num`` is ``VARCHAR(32)`` in this database
 (asyncpg raises ``StringDataRightTruncationError`` otherwise).
-``d04_519_oauth_clients_owner_id_nullable`` is exactly 41 chars; using
-the shorter ``d04_519_oauth_owner_nullable`` (29 chars) instead.
+The longer candidate ``d04_519_oauth_clients_owner_id_nullable`` is
+39 chars — still over the cap — so this migration uses the shorter
+``d04_519_oauth_owner_nullable`` (29 chars). The Python filename is
+allowed to be longer than the revision id, so we keep the descriptive
+filename for grep-ability.
 """
+
+import sqlalchemy as sa
 
 from alembic import op
 
@@ -48,6 +53,7 @@ def upgrade() -> None:
     op.alter_column(
         "oauth_clients",
         "owner_id",
+        existing_type=sa.String(length=255),
         nullable=True,
     )
 
@@ -62,5 +68,6 @@ def downgrade() -> None:
     op.alter_column(
         "oauth_clients",
         "owner_id",
+        existing_type=sa.String(length=255),
         nullable=False,
     )
--- a/backend/alembic/versions/d04_519_oauth_clients_owner_id_nullable.py
+++ b/backend/alembic/versions/d04_519_oauth_clients_owner_id_nullable.py
@@ -1,0 +1,66 @@
+"""Drop NOT NULL on oauth_clients.owner_id (#519, #513 follow-up).
+
+Issue #513 (PR #518) added Dynamic Client Registration (RFC 7591) and
+adjusted the response model so DCR-registered clients (which have no human
+owner) can serialize ``owner_id=None``. The underlying SQLAlchemy column
+and DB constraint were left as ``NOT NULL``, so any successful DCR INSERT
+fails with::
+
+    psycopg2.errors.NotNullViolation: null value in column "owner_id"
+    of relation "oauth_clients" violates not-null constraint
+
+DCR rejection paths still work end-to-end (they short-circuit before the
+INSERT), but DCR accept paths are broken in production. This migration
+aligns the DB constraint with the model + response shape so DCR can
+register a public client with no owner.
+
+Admin-managed clients (``POST /api/v1/oauth/clients``) continue to record
+``owner_id=current_user_id``; the relaxation only opens the door for the
+DCR path. The accompanying model change in ``models/auth.py`` flips
+``OAuth2Client.owner_id`` to ``nullable=True``.
+
+Downgrade re-applies ``SET NOT NULL``. Any existing rows with
+``owner_id IS NULL`` (DCR-registered clients) will block the downgrade —
+that is the correct fail-loud behavior; the operator must either delete
+the DCR rows or backfill an owner before downgrading.
+
+Revision ID: d04_519_oauth_clients_owner_id_nullable
+Revises: c03_471_seed_pricing
+
+NOTE: Revision IDs are capped at 32 chars because
+``alembic_version.version_num`` is ``VARCHAR(32)`` in this database
+(asyncpg raises ``StringDataRightTruncationError`` otherwise).
+``d04_519_oauth_clients_owner_id_nullable`` is exactly 41 chars; using
+the shorter ``d04_519_oauth_owner_nullable`` (29 chars) instead.
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "d04_519_oauth_owner_nullable"
+down_revision = "c03_471_seed_pricing"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Drop NOT NULL on oauth_clients.owner_id."""
+    op.alter_column(
+        "oauth_clients",
+        "owner_id",
+        nullable=True,
+    )
+
+
+def downgrade() -> None:
+    """Restore NOT NULL on oauth_clients.owner_id.
+
+    This will fail if any rows have ``owner_id IS NULL`` (DCR-registered
+    clients). Operators must delete those rows or backfill an owner before
+    downgrading.
+    """
+    op.alter_column(
+        "oauth_clients",
+        "owner_id",
+        nullable=False,
+    )
--- a/backend/src/models/auth.py
+++ b/backend/src/models/auth.py
@@ -353,7 +353,12 @@ class OAuth2Client(Base):
         client_id: OAuth2 client identifier (unique, public)
         client_secret_hash: SHA256 hash of client secret (confidential, always required)
         client_name: Human-readable name (e.g., "ChatGPT Connector")
-        owner_id: User ID who registered this client
+        owner_id: User ID who registered this client. ``None`` for clients
+            created via Dynamic Client Registration (RFC 7591) — DCR is a
+            public, owner-less endpoint, so the workspace context is resolved
+            from the consenting user's session at ``/authorize`` time
+            rather than from the client record. See migration revision
+            ``d04_519_oauth_owner_nullable`` (issue #519).
         redirect_uris: Allowed redirect URIs (JSON array)
         grant_types: Allowed grant types (JSON array: authorization_code, refresh_token)
         response_types: Allowed response types (JSON array: code)
--- a/backend/src/models/auth.py
+++ b/backend/src/models/auth.py
@@ -386,8 +386,8 @@ class OAuth2Client(Base):
     # Issue #519 (#513 follow-up): DCR-registered clients (RFC 7591) have no
     # human owner — ``dynamic_client_registration`` creates them with
     # ``owner_id=None``. Admin-managed clients still record the creating
-    # user's id. Migration ``d04_519_oauth_clients_owner_id_nullable``
-    # drops the corresponding NOT NULL DB constraint.
+    # user's id. Migration revision ``d04_519_oauth_owner_nullable`` drops
+    # the corresponding NOT NULL DB constraint.
     owner_id = Column(String(255), nullable=True, index=True)
 
     # Issue #169: Workspace-scoped OAuth clients (access all contexts in workspace)
--- a/backend/src/models/auth.py
+++ b/backend/src/models/auth.py
@@ -383,7 +383,12 @@ class OAuth2Client(Base):
     client_id = Column(String(48), nullable=False, unique=True, index=True)
     client_secret_hash = Column(String(64), nullable=False)  # SHA256 hash (always required)
     client_name = Column(String(100), nullable=False)
-    owner_id = Column(String(255), nullable=False, index=True)
+    # Issue #519 (#513 follow-up): DCR-registered clients (RFC 7591) have no
+    # human owner — ``dynamic_client_registration`` creates them with
+    # ``owner_id=None``. Admin-managed clients still record the creating
+    # user's id. Migration ``d04_519_oauth_clients_owner_id_nullable``
+    # drops the corresponding NOT NULL DB constraint.
+    owner_id = Column(String(255), nullable=True, index=True)
 
     # Issue #169: Workspace-scoped OAuth clients (access all contexts in workspace)
     # Migration 034: context_id removed (deprecated)
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -1,0 +1,59 @@
+"""Pytest conftest for the integration test suite.
+
+Steers ``DATABASE_URL`` at ``TEST_DATABASE_URL`` deterministically before
+any test module imports ``config.database`` or ``api.main``, so the route
+handler's ``get_sync_session()`` connects to the test DB regardless of
+which module is collected first.
+
+The mutation happens at **conftest import time** (module top, below the
+imports) because ``config.database.DATABASE_URL`` and
+``api.main`` read the env at their own import time. A session-scoped
+autouse fixture restores the prior value at session end so a leaked
+override doesn't follow the user out of pytest.
+
+Mirrors the env-steering pattern in
+``tests/integration/test_alembic_migrations.py:_alembic_at_test_db`` —
+that one is per-call (context manager) for the migration tests; this
+conftest applies the same idea suite-wide.
+"""
+
+from __future__ import annotations
+
+import os
+from collections.abc import Generator
+
+import pytest
+
+_DEFAULT_TEST_DB_URL = "postgresql+asyncpg://kagura:kagura_dev_password@localhost:5432/kagura_test"
+
+# CRITICAL: this assignment runs at conftest *import* time, which pytest
+# does before collecting test modules. ``api.main`` and ``config.database``
+# read ``DATABASE_URL`` at their own import time, so the override must
+# happen before they are imported. ``_PREV_DATABASE_URL`` captures the
+# original value (if any) so the session-scoped fixture below can restore
+# it on exit.
+_PREV_DATABASE_URL = os.environ.get("DATABASE_URL")
+os.environ["DATABASE_URL"] = os.environ.get("TEST_DATABASE_URL", _DEFAULT_TEST_DB_URL)
+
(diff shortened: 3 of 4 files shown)


JFK:
Loop 1: addressed all 3 clusters of findings (8 threads).

**Migration docstring + existing_type** (`alembic/versions/d04_519_oauth_clients_owner_id_nullable.py`)
- Fixed `Revision ID:` to match the actual `revision = "d04_519_oauth_owner_nullable"`.
- Corrected the char-count rationale (39, not 41) and clarified that both candidates exceed the 32-char alembic_version cap so the shorter form is necessary.
- Added `existing_type=sa.String(length=255)` to both upgrade/downgrade `op.alter_column()` calls per the `a99_external_keys_workspace_not_null.py` convention.

**Model comment** (`backend/src/models/auth.py`)
- Updated the inline comment to reference the actual revision id (`d04_519_oauth_owner_nullable`) so grep works.

**Integration test hermeticity** (`backend/tests/integration/test_oauth_dcr_integration.py`)
- Module-level env override forces `DATABASE_URL` to `TEST_DATABASE_URL` (or the project default `kagura_test` URL) before `api.main` is imported.
- Test-only fallbacks for `API_KEY_SECRET` / `JWT_SECRET` via `setdefault` (no clobber of real values).
- New `_assert_test_db()` helper verifies `current_database()` ends with `_test` before yielding the session AND before the cleanup DELETE — mirrors the pattern in `test_alembic_migrations.py`.
- Added `_ensure_oauth_clients_schema` autouse fixture that creates auth tables in the test DB via a sync engine (the test does not depend on conftest's async fixture).

Migration round-trip verified end-to-end: `alembic downgrade -1 && alembic upgrade head` cleanly toggles `oauth_clients.owner_id` between NOT NULL and nullable.

128/128 OAuth-adjacent tests pass; 2/2 new integration tests pass when run via `TEST_DATABASE_URL=... pytest tests/integration/test_oauth_dcr_integration.py` (the CI invocation pattern via `make test-integration`).

JFK:
Loop 2: addressed both findings.

**Schema setup via `alembic upgrade head`** (`tests/integration/test_oauth_dcr_integration.py`)
- Replaced `Base.metadata.create_all(engine)` with `alembic upgrade head` so the test actually exercises migration `d04_519_oauth_owner_nullable`. With `create_all`, a missing/buggy migration would still let the test pass (because the ORM model says `nullable=True`); now the migration itself must produce the right column shape.
- Added `test_migration_made_owner_id_nullable` that reads `information_schema.columns` directly and asserts `is_nullable='YES'` — independent of the ORM's view.

**Skip when DB unreachable**
- Added `pytestmark = pytest.mark.skipif(not _check_db_available(), reason="Test database not available")` mirroring the `tests/api/test_api_integration.py` helper. The module now skips cleanly instead of erroring at fixture setup.

Test count 2 → 3, all green: `TEST_DATABASE_URL=...kagura_test pytest tests/integration/test_oauth_dcr_integration.py` → 3 passed.

JFK:
Loop 3: addressed all 3 clusters (4 threads).

**Test-DB safety guard before `alembic upgrade`** (`test_oauth_dcr_integration.py`)
- New `_assert_alembic_target_is_test_db()` connects via psycopg2 and asserts `current_database()` ends with `_test` before `command.upgrade()` runs. Refuses to migrate dev/prod DBs even if `TEST_DATABASE_URL` is misconfigured.

**Hoist env steering to `backend/tests/integration/conftest.py`** (new file)
- `os.environ["DATABASE_URL"]` mutation now happens at conftest *import* time, which pytest does before collecting any test module. This eliminates the order-dependent failure mode where another integration module's import of `api.main` could lock in the wrong DATABASE_URL before the test module's module-level override fired.
- Session-scoped autouse fixture `_restore_database_url_after_session` captures the prior value at conftest-load and restores it on session exit, so the override doesn't leak out of pytest into the parent shell session.
- `API_KEY_SECRET` / `JWT_SECRET` test-only fallbacks moved to the conftest as well — same hermetic-suite rationale.
- Removed the module-level env block from `test_oauth_dcr_integration.py`; conftest is the single source of truth.

**OAuth2Client.owner_id docstring** (`models/auth.py`)
- Class attribute description now calls out that `owner_id=None` is the DCR (RFC 7591) case and references migration revision `d04_519_oauth_owner_nullable` for grep-ability.

Verified end-to-end with a clean env: `unset DATABASE_URL && TEST_DATABASE_URL=...kagura_test pytest tests/integration/test_oauth_dcr_integration.py` → 3 passed. The conftest-only steering path works without a shell-exported DATABASE_URL.

JFK:
Loop 4: addressed all 3 findings.

**conftest module-reload guard** (`tests/integration/conftest.py`)
- After the env-steering at conftest-import time, walk `sys.modules` and `importlib.reload` `config.database` and `db.base` if they are already imported. Defensive when integration tests are run in isolation (no race), necessary for mixed `pytest tests/` invocations where a sibling unit-test module's earlier import of `api.main` could otherwise lock in the pre-override DATABASE_URL.

**`raise_server_exceptions=False`** (`test_oauth_dcr_integration.py`)
- TestClient now returns 5xx responses instead of re-raising. The pre-#519 regression was a 500 bubbling from psycopg2's NotNullViolation; the test needs to be able to inspect the response, not catch the exception. Matches the convention in other integration tests in this repo.

**`_reset_alembic_state` before `alembic upgrade head`**
- Inlined the `DROP SCHEMA public CASCADE; CREATE SCHEMA public` pattern from `tests/integration/test_alembic_migrations.py:_reset_alembic_state` so `command.upgrade("head")` starts from a clean schema even if a sibling fixture (e.g. `tests/conftest.py`'s session-scoped async engine) already did `Base.metadata.create_all` in this session. Guarded by the same `*_test` DB-name check.

Verified end-to-end with a clean env: `unset DATABASE_URL && TEST_DATABASE_URL=...kagura_test pytest tests/integration/test_oauth_dcr_integration.py` → 3 passed.

JFK:
Loop 5: addressed the 1 finding.

**Replaced `importlib.reload(db.base)` with targeted state reset** (`tests/integration/conftest.py`)

The prior loop's reload approach was unsafe: `db.base` defines `Base = declarative_base()`, and reloading the module recreates that `Base` while `models.auth` / `models.memory` still hold references to the original. The result would be a split metadata registry — alembic's `target_metadata` and any `Base.metadata.create_all` would see an empty (or partial) registry instead of all the model tables.

Now:
- `config.database` is reloaded (only mutable state is `DATABASE_URL`, safe to reload).
- `db.base` is NOT reloaded. `_reset_db_base_state()` walks the lazy-init globals (`engine`, `sync_engine`, `async_session_factory`, `sync_session_factory`), disposes the engines, and sets them to `None`. Next access rebuilds them via the function-local `from config.database import DATABASE_URL` (which now resolves to the freshly reloaded value).

Verified mixed-suite: `pytest tests/api/test_oauth_dcr.py tests/integration/test_oauth_dcr_integration.py` → 55 passed (was failing the `_assert_test_db` guard before this fix because the cached engine pointed at the dev DB).

JFK:
Loop 5 follow-up: addressed AsyncEngine dispose finding.

`db.base.engine` is an `AsyncEngine` whose `.dispose()` is a coroutine. The conftest dispose path now branches:
- async engine → `existing.sync_engine.dispose()` (synchronous, releases the same pool)
- sync engine → `existing.dispose()` (plain sync `Engine`)

Verified with `-W error::RuntimeWarning` that no un-awaited coroutines leak. 55/55 mixed-suite still passes.

Issue #515, opened 2026-04-30
feat(profile): manual 're-sync from IdP' button to refresh email/name


JFK:
## Context

After #481 ships, OAuth callbacks auto-sync `email` and `name` from the IdP whenever they change AND `email_verified=True`. This covers the common case (user logs in periodically, eventual consistency on next login).

Edge case: a user who changes their IdP primary email and **does not re-authenticate** for a long time. Their memory-cloud `users.email` stays stale until next login, which can break:
- transactional email delivery (#469 erasure confirm, future password-reset, billing receipts)
- any feature that displays the user's current email (profile page, audit views)

A manual \"Re-sync from IdP\" button on the profile page lets the user force a refresh on demand without a full sign-out+sign-in cycle.

## Scope

**In scope:**
- Backend endpoint: `POST /api/v1/me/refresh-oauth` — initiates a re-auth flow with the user's current OAuth provider (no fresh consent screen if Google/GitHub session is still valid; otherwise standard OAuth handshake)
- After the IdP returns, the existing `ensure_user` path runs and syncs email/name as on a normal login
- Frontend: profile page button \"Refresh from \\${provider}\" that opens the OAuth flow in the same window (or a popup), shows loading state, then refreshes the profile data on return
- i18n strings (en + ja)
- Optional: rate-limit the endpoint so users can't hammer the IdP (1/min/user feels right; reuse the existing rate-limit middleware)

**Out of scope:**
- Account linking — separate issue (large)
- Switching auth_provider — separate issue
- Bulk re-sync for ops (CLI/admin script) — separate issue if needed

## Acceptance

- [ ] `POST /api/v1/me/refresh-oauth` initiates the IdP-specific OAuth flow for the user's existing `auth_provider`
- [ ] On return, `ensure_user` runs and syncs email/name (HMAC audit row written, same path as login-time sync)
- [ ] Profile UI shows the button only for OAuth users (`auth_method='oauth'`); password users see nothing
- [ ] Loading + success + error states are user-visible
- [ ] Rate-limit: 1 request per minute per user (return 429 with Retry-After)
- [ ] Unit tests for the endpoint
- [ ] Vitest test for the profile button render + click handler
- [ ] E2E test (Playwright) for the happy path

## Design notes

- The button is a **convenience**: the auto-sync on next login already covers correctness. UI copy should reflect that (\"Update profile from Google now\" rather than \"required\").
[...]

Issue #512, opened 2026-04-29
chore(test): aria-query unhandled rejection in parallel Vitest full-suite


JFK:
## Symptom

When running the full frontend test suite via `npm test` (`vitest run`), Vitest reports **2 test files failed** even though all 288 individual test assertions pass:

```
Test Files  2 failed | 32 passed (34)
      Tests  288 passed (288)
     Errors  1 error
   Duration  1.93s
```

The error is an unhandled rejection during module import of `aria-query`:

```
TypeError: Invalid attempt to destructure non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
 ❯ _nonIterableRest node_modules/aria-query/lib/rolesMap.js:15:37
 ❯ _slicedToArray node_modules/aria-query/lib/rolesMap.js:14:130
 ❯ node_modules/aria-query/lib/rolesMap.js:36:25
 ❯ _loop node_modules/aria-query/lib/rolesMap.js:35:43
 ❯ Object.<anonymous> node_modules/aria-query/lib/rolesMap.js:21:7
 ❯ Module._compile node:internal/modules/cjs/loader:1469:14

This error originated in "src/components/contexts/MembersSection.test.tsx" test file.
```

## Reproduction

Full suite (fails):

```bash
cd frontend && npm test
# Test Files  2 failed | 32 passed
```

Isolated runs (pass):

```bash
cd frontend && npx vitest run src/components/contexts/MembersSection.test.tsx
# Test Files  1 passed (1) | Tests 10 passed (10)
```

The flake only manifests under parallel execution. `aria-query` is pulled transitively via `@testing-library/dom` → `@testing-library/react`.

## Hypothesis

Classic test pollution: another test file's import path manipulates a module shared by `aria-query`'s `rolesMap.js` initializer (likely `Object.entries` / `Symbol.iterator` polyfill or shimmed Map), and when `MembersSection.test.tsx` imports `aria-query` later in the parallel worker pool, the destructure at `rolesMap.js:15` fails.

This is **not caused by recent application code** — it reproduces independently of any single PR's changes (verified during PR #511 / #441 ship review: the flake was present before the merge).

## Acceptance criteria

- [ ] Full-suite `npm test` reports `Test Files X passed (X)` with zero failed files
- [ ] No `Unhandled Rejection` errors in the test output
- [ ] CI does not fail on the flake (currently masked by `Tests 288 passed (288)` summary line)

[...]

Issue #481, opened 2026-04-29
feat(auth): sync email/name from IdP userinfo on every OAuth login


JFK:
## Context

Discovered while implementing #469 (gate `confirm_token` in erasure response on auth method, post real-email-provider). Gate1 design review (CSO) and CLO follow-up confirmed #469's design, but a deeper trace of OAuth login flow (`backend/src/auth/roles.py:125-185` `ensure_user`) revealed a hidden delivery-failure mode for OAuth users.

This issue is a **pre-flip gate for `EMAIL_PROVIDER=resend`** — siblings #480 (Privacy Policy sub-processor disclosure + DPA env validator) and #469 (response gating). All three must land before the production env flip.

## Problem

`ensure_user` (`backend/src/auth/roles.py:125-185`) is invoked at every Google / GitHub OAuth callback. Today it:

```python
result = await db.execute(select(User).filter_by(email=email))
user = result.scalar_one_or_none()

if user:
    # Update last_login_at
    user.last_login_at = utcnow()
    await db.commit()
    return Role(user.role)
```

It updates `last_login_at` only — **`user.email` and `user.name` are never re-synced** from `user_info["email"]` / `user_info["name"]`.

Combined with email-keyed lookup, this creates three failure modes:

| Scenario | Lookup | Result | Erasure email delivery |
|---|---|---|---|
| (A) IdP primary email unchanged | hit | `last_login_at` updated, DB email unchanged | ✅ delivers (DB == IdP) |
| (B) IdP primary changed → user logs in | miss (new email) | **new user row created** (or IntegrityError race depending on `oauth_sub` unique constraints) | ✅ (new user has new email) but **memory continuity is broken** — the same person now has two unlinked accounts |
| (C) IdP primary deleted but session still valid | (login skipped) | DB email = stale | ❌ **erasure-confirmation email bounces, OAuth user has no path to confirm** (Issue #469's response-token-only-for-password design relies on email being deliverable) |

Scenario (B) silently fragments user data — primary memory-continuity violation. Scenario (C) is the immediate Issue #469 blocker: the OAuth-only email-delivery model assumes deliverability.

## Fix

In `ensure_user`, on the existing-user branch, sync `email` and `name` from the IdP `user_info` payload:

```python
if user:
    # Issue #XXX (this issue): sync email/name from IdP every login so
    # transactional-email delivery (Issue #469 erasure confirm, future
    # password-reset, billing receipts) targets the user's current
[...]

PR #522, opened 2026-05-01
feat(profile): sign-in method UI + manual IdP refresh (#514, #515)

### LINKED ISSUE #514: feat(profile): show auth_provider (Google/GitHub) on user profile
## Context

After #481 ships (lookup-key swap + email/name sync), the `users.auth_provider` column reliably reflects which IdP a user originally registered through (`google` | `github`). Today the value is **stored but never displayed**: there is no UI surface that tells the user which OAuth provider their account is bound to.

Adding a read-only display on the profile page closes a small but real UX gap surfaced in the post-#481 follow-up discussion (2026-04-30).

## Scope (intentionally extra-small)

**In scope:**
- Backend: `GET /api/v1/me/profile` (or wherever profile fields are returned today) exposes `auth_provider` in the response payload
- Frontend: profile page renders a single read-only line, e.g. `Sign-in method: Google` / `Sign-in method: GitHub` / `Sign-in method: Password`
- i18n strings (en + ja)

**Out of scope (deferred to separate issues):**
- Account linking / connect-additional-provider — needs full design review (separate v0.14.3+ issue)
- \"Re-sync from IdP\" manual button — separate issue
- Provider switching after registration — separate issue

## Acceptance

- [ ] `auth_provider` is included in the profile API response
- [ ] Frontend shows the provider as a read-only string on the profile page
- [ ] Display string is i18n'd (en + ja)
- [ ] Password-auth users (\`auth_method='password'\`) show \"Password\" or equivalent and not a blank/Null
- [ ] Vitest test for the rendering branch
- [ ] Unit test for the API field surface

## Notes

- `User.auth_provider` is currently set only at row creation (Issue #361). After-the-fact provider switching is **not** supported by the data model — this UI just reflects what's already in the DB.
- Read-only by design: any \"change provider\" affordance is a separate issue.

## Related

- #481 (parent rescope context — auth identity model)
- #361 (auth_provider tracking origin)

JFK:
Closes #514
Partially addresses #515

## Summary
- **#514**: read-only "Sign-in method" display on the profile page (Google / GitHub / Email + Password / Other), backed by `auth_method` + `auth_provider` exposed on `GET /api/v1/auth/me` and on `UserProfileResponse`.
- **#515**: manual "Refresh from {provider}" button that POSTs to a new `POST /api/v1/me/refresh-oauth`, runs the IdP round-trip, and lets `ensure_user` (#481) sync the latest email/name without forcing a sign-out / sign-in cycle. The user keeps their existing session — refresh ≠ re-login.
- Same-user binding (`oauth2_state_user:{state}` ↔ IdP-returned `sub`) blocks Account Confusion attacks where the user switches IdP accounts mid-flow.
- `return_to` validator rejects open-redirect shapes (`//evil`, absolute URLs, `javascript:`, backslash + URL-encoded variants).
- Rate limit: 1/min/user via Redis `increment_counter`, with safe fall-through on Redis errors.

## Implementation notes
- New focused route file `backend/src/api/routes/me_oauth.py` for the POST endpoint; OAuth callback branch lives next to its peers in `backend/src/api/routes/auth.py` as a `_maybe_refresh_redirect(state, idp_sub)` helper that both `google_callback` and `github_callback` call after `ensure_user`.
- State token contract extends the existing `oauth2_state:{state}` flow with two new keys (`oauth2_state_intent:{state}`, `oauth2_state_user:{state}`) — both 5-minute TTL, both deleted on read. The intent key lets the callback short-circuit session creation; the user key enforces same-user binding.
- Frontend: profile page gains two helpers (`getSignInMethodLabel`, `getRefreshProviderName`) plus a `useEffect` that consumes `?refreshed=1` / `?error=refresh_*` query params on the post-callback redirect. A `useRef` sentinel + `user` deps gate the effect so a still-loading `useAuth()` doesn't fall back to "from IdP" instead of "from Google".
- Schema breaking change: `UserProfileResponse.auth_method` is now required (no default). Production code uses `model_validate(db_user)` with `from_attributes=True` and the `User` model column is `NOT NULL DEFAULT 'oauth'`, so the production path is safe; one test fixture was updated.
- Refactor: extracted `build_github_authorization_url(client_id, redirect_uri, state)` so `github_login` (existing) and `refresh_oauth` (new) share the same scope string + parameter order. The frontend's three-way error-code ternary became a `Record<string, string>` lookup table.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: green via /ask (operator override after reviewer-emitted yellow on scope-asymmetry concerns)
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/514-batch-514-515.gate1.md`
- `~/.claude/cache/gh-issue-driven/514-batch-514-515.gate2.md`

## Tests
- **Backend**: 71 new tests across 3 new files (`test_auth_me_signin_method.py`, `test_me_refresh_oauth.py`, `test_auth_refresh_callback.py`) + 1 fixture update (`test_schema_serializers.py`). Covers endpoint validation, rate limit (over-limit + Redis-failure paths), `return_to` validator (10 reject + 4 accept parametrized cases), and all four `_maybe_refresh_redirect` branches (no-intent / intent=login / matching-user / mismatch / expired / no-return_to / manager-missing).
- **Frontend**: 23 new vitest cases covering both helpers (5 + 5 branches), refresh-button visibility (4 user shapes), click flow (POST + window.location, 429 toast, generic error toast), and post-callback search-param effect (refreshed / mismatch / expired / clean URL).
- `make lint`: clean. Project-wide `make test-local` (api/ + auth/ + services/): green except 11 worker-timeouts on unrelated files (pre-existing #512 Vitest parallel-suite flake).

## Deferred / follow-ups
- **Playwright E2E happy path** for `/me/refresh-oauth` is part of #515's acceptance criteria but the project has no Playwright infrastructure (`make test-e2e` is pytest-based). Hence "Partially addresses #515" in the title — a follow-up issue should decide whether to (a) add Playwright infra and write the E2E, or (b) write a pytest TestClient-based integration test that round-trips the state token through real Redis + a mock IdP, or (c) close out #515 acknowledging the unit-level coverage is dense enough to stand in.
- **Existing `/auth/{provider}/login` open-redirect surface**: the existing `google_login` and `github_login` GET endpoints take `return_to` as a query string with no validation — same vulnerability shape this PR's new endpoint defends against. Out of scope here, but worth a project-wide hardening issue applying `_is_safe_relative_return_to` (or its equivalent) to all `return_to` ingress points.
- Consider adding a `get_oauth2_manager()` getter (parallel to `get_session_manager()` at `backend/src/api/routes/auth.py:72`) so other route files can avoid reaching into `auth_module._oauth2_manager` directly.


JFK:
Loop 1: addressed 3 Copilot items (#2 window.location restoration, #3 config-before-Redis ordering, #5 CRLF / control-byte rejection in `return_to`). Skipped 2 with rationale:

- **logger.info f-string vs structured kwargs**: auth.py module-wide pattern is f-string (lines 423, 467, 502). Migration to structlog kwargs is a project-wide concern, not isolated to refresh_oauth. Earlier in this branch I tried kwargs and the stdlib Logger fallback used in tests rejected them with `TypeError` — confirms this is not a one-file change.
- **Hardcoded \"Google\" / \"GitHub\" in getRefreshProviderName**: brand names are conventionally not translated (Google's own JA docs say \"Google でサインイン\"). The \"IdP\" fallback path is unreachable in practice — the search-param effect now guards on `if (!user) return;` (commit cd7317ce), so by the time `getRefreshProviderName` is called, `user.auth_provider` is always populated for the rendered button case.

New commit: `f0edd02c`. 65 backend + 23 frontend tests green.

JFK:
Loop 2: addressed 3 Copilot items (#2 delete-on-read for `oauth2_return_to:{state}` on all refresh branches, #3 derive `email` from `user?.email` so refresh updates the input, #4 i18n key for the IdP fallback). Skipped 1 with same rationale as loop 1 (structlog kwargs vs f-string is a project-wide pattern concern, not isolated to refresh_oauth).

New commit: `4e807003`. 67 backend (+2 delete-on-read tests) + 23 frontend tests green.

JFK:
Loop 3: addressed all 4 Copilot items — (#1) FRONTEND_URL prefix on return_to redirect, (#2) <Label> → <p> in refresh-button section (a11y), (#3) UserProfileResponse Literal types for auth_method/auth_provider, (#4) getRefreshProviderName routes through t() so brand display goes through next-intl. Commit: d0b737fe. 67 backend + 23 frontend tests green.

JFK:
Loop 4: 1 actionable item — switched build_github_authorization_url to urllib.parse.urlencode so GitHub OAuth params (client_id, redirect_uri, scope, state) are escaped instead of f-string-interpolated raw. Closes the redirect_uri-with-query-string injection vector. Commit: (push pending hash). 67 backend + 23 frontend tests green.

JFK:
Loop 5: 1 Copilot item (structlog kwargs vs f-string). Investigated more carefully this loop and found the real reason it can't be applied cleanly in this PR:

- `utils/logger.py` does document the kwargs convention, and structlog.get_logger() returns a BoundLoggerLazyProxy that accepts kwargs in production.
- BUT pytest imports the route module directly without going through `api/main.py`'s startup, so `setup_logger()` never fires. structlog's default config falls back to wrapping stdlib `logging.Logger`, which rejects kwargs with `TypeError: Logger._log() got an unexpected keyword argument 'expected_user_id'`.
- I confirmed this by trying the kwargs version — 2 tests failed in the auth-callback helper test file.

Right fix: have `tests/conftest.py` call `setup_logger()` so structlog is configured for tests too. But that's a project-wide change — every `logger.info("event", key=value)` call across the codebase has the same latent fragility (e.g. `auth/dependencies.py:429`, `services/addon_calculator_service.py:147`, etc.). Doing it as a side effect of a profile-feature PR would create review noise.

Compromise this loop: kept f-string, but reformatted to match the structlog k=v shape so log aggregation regex can still pull structured fields. Added inline comments pointing at the test-infra root cause. **Filed a follow-up issue mentally — will open it after merge** to fix `tests/conftest.py` and migrate the f-strings back.

Commit: aad56d96 + d-pending-hash. Reaching `max_loops=5` so this is the loop's final iteration regardless of next Copilot response — exiting the loop after this push.

JFK:
## Local testing surfaced 3 latent bugs in pre-existing #481 / #358 OAuth code

While doing local validation of #515 (GitHub OAuth refresh flow), the test path "Google user A logged in with email X → enable signup gate → sign up via GitHub with email Y → change GitHub primary email to X" hit **503 DB-002** three times in a row. Each was a distinct latent bug in code merged via #481 / #358 that the unit tests had not caught because they used pure Python mocks rather than the real SQLAlchemy 2.0 + asyncpg wrap shapes.

| # | Bug | Location | Fix |
|---|---|---|---|
| 1 | `_is_email_unique_violation` only read `orig.constraint_name`, but asyncpg wraps `UniqueViolationError` in `AsyncAdapt_asyncpg_dbapi.IntegrityError` and exposes `constraint_name` on `orig.__cause__`, not `orig` | `auth/roles.py` | `d5792024` — walk `orig` → `orig.__cause__` → `str(orig)` chain |
| 2 | `_is_existing_user` used `.scalar_one_or_none()` on `WHERE email = X OR user_id = Y`, raising `MultipleResultsFound` when the OR predicate hit two distinct rows (cross-provider drift) | `services/signup_gate_service.py` | `337c69b6` — `.first()` + `LIMIT 1` |
| 3 | `_sync_existing_user`'s except block accessed `user.user_id` after a failed commit had expired the ORM instance, triggering an async lazy-load that raised `MissingGreenlet` | `auth/roles.py` | `3147fcde` — pre-capture user attrs into plain locals before the try block |

All three are real production bugs — without the fixes, every cross-provider email collision falls into the global `SQLAlchemyError → 503 DB-002` handler instead of reaching the friendly `ConflictError → /login?error=email_in_use` redirect that #481 was supposed to deliver.

Tests added: 6 new `TestIsEmailUniqueViolationWrapShapes` cases pin all three layers (orig direct, `__cause__` chain, message fallback) + negative cases. 4 cases in `TestIsExistingUser` pin the `.first()` + multi-row tolerance. The MissingGreenlet path is hard to repro in pure-mock unit tests (plain Python objects don't expire attributes); the inline comment documents the trap and live container repro confirms the fix works end-to-end.

These commits widen PR #522's scope a bit but the alternative is shipping #514/#515 with these latent bugs unaddressed — they live on the same OAuth callback path #515 calls into. Filing a follow-up issue would have left users hitting the 503 in production. All three bugs are scope-adjacent to #515's "manual IdP refresh" feature anyway: any user clicking "Refresh from GitHub" with a drifted IdP email would have hit bug 1 or 3.

PR is now at 18 commits / ~2.4k lines. CI green. Local manual test confirms `/auth/github/callback` returns `303 → /login?error=email_in_use` on cross-provider email collision instead of `503 DB-002`.

JFK:
Loop 6: addressed both Copilot items + an actual root-cause fix that closes the structlog-kwargs investigation.

- (urlencode docstring) Fixed: ``urlencode`` defaults to ``quote_plus`` (space → ``+``), not ``%20``. GitHub accepts both, so the encoding behaviour is correct — just the docstring was wrong. Updated.
- (structlog kwargs) **Found the actual root cause this time**: ``backend/src/api/routes/auth.py`` had been binding ``logger`` via ``logging.getLogger`` (stdlib) while every other route module uses ``utils.logger.get_logger`` (structlog). The pre-existing inconsistency was the reason kwargs TypeError'd in tests — auth.py's logger was stdlib all along, and my f-string workaround had been masking it. Switched auth.py to ``get_logger``, added ``setup_logger()`` to ``tests/conftest.py`` so structlog is configured for the pytest process too, and re-flipped the three f-string log calls to canonical kwargs form.

Loops 1–5 had me declining this with the rationale "test infra falls back to stdlib Logger". The diagnosis was right but the fix-location was wrong — the issue wasn't that tests don't configure structlog, it was that auth.py specifically wasn't even using structlog. Both gaps are now closed.

954 backend tests pass (wider sweep). 110 in our scope. Lint clean.

JFK:
Loop 7: 1 actionable item — converted the lingering logger.warning("registration_blocked: %s", email) call (broken by the stdlib→structlog switch in the prior commit) to canonical structlog kwargs form. Bonus: hashed the email with the existing audit_hmac_key per the project's PII-in-logs pattern so the log record stays safe to retain. 954 backend tests pass.

Issue #526, opened 2026-05-01
feat(api+frontend): workspace-scoped sleep reports view


JFK:
## Summary
Workspace owner / admin self-service view of Sleep Maintenance reports for **their workspace only**, mirroring the two-endpoint pattern shipped in #472 cost aggregation.

Currently `/api/v1/admin/sleep-reports` is admin-only (cross-workspace). Workspace owners have no UI to see *their own* sleep run history, audit phase outcomes, or check whether a recent maintenance run actually completed. This is a v0.15.0 cost/observability gap surfaced while building the `/workspace/cost` page (#473): the cost dashboard now lives at `/workspace/cost` and the natural sibling — sleep run history — is missing.

## Backend (split into a separate atomic PR)
New endpoint: `GET /api/v1/workspaces/{workspace_id}/sleep-reports`
- AuthZ via `PermissionService.check_workspace_admin()` (owner/admin) — same gate as `/workspace/cost-aggregation`
- Same query params as the admin route (`status`, `limit`, `offset`, `user_id`) MINUS `context_id` filter that's not workspace-scoped — actually keep `context_id` since contexts are workspace-bound
- Returns the same `SleepReportListResponse` shape
- Detail endpoint: `GET /api/v1/workspaces/{workspace_id}/sleep-reports/{report_id}` — owner/admin gate + workspace-membership check on the report
- Shared service extraction: lift the listing logic out of `admin_sleep_reports.py` into `services/sleep_reporter_service.py` (or similar) so both routes call it
- Tests:
  - Owner sees workspace's reports
  - Non-owner gets 403
  - Cross-workspace probe (path workspace_id ≠ caller's workspaces) gets 403
  - Detail endpoint rejects reports from other workspaces (404, not 403 — uniform CWE-639 disclosure)

## Frontend
New page: `/workspace/sleep-reports` (and detail at `/workspace/sleep-reports/{report_id}`)
- Reuse `/admin/sleep-reports` UI: extract a shared `<SleepReportsList endpoint=admin|workspace>` component (analogous to the `<CostDashboard>` extraction we did in #473)
- Sidebar: add 'Sleep Reports' to the workspace section with `requiredWorkspaceRole: 'admin'`
- 'Run Now' action stays admin-only (manual sleep trigger is `/admin/sleep/run` for now — a workspace-scoped trigger is out of scope here, file separately if needed)

## Out of scope
[...]

Issue #472, opened 2026-04-28
feat(api): cost aggregation API (period × workspace × user) (#180)


JFK:
## Summary
Aggregation API returning LLM usage and computed cost (`$`) breakdown per period × workspace × user, sourced from the cost-grade `sleep_reports` schema.

Split from #180 (umbrella). **Depends on the schema-extension child** (must merge first). Prerequisite for the dashboard child.

## Endpoint sketch
```
GET /api/admin/cost-aggregation
  ?period=day|week|month
  &from=<ISO date>
  &to=<ISO date>
  &workspace_id=<uuid>     # optional
  &user_id=<string>        # optional
```

Response (per row):
```json
{
  "period_start": "2026-04-01",
  "workspace_id": "...",
  "user_id": "...",
  "calls": 142,
  "tokens_in": 320000,
  "tokens_out": 18000,
  "tokens_cached_in": 280000,
  "embedding_tokens": 95000,
  "cost_usd": 1.234,
  "cost_breakdown_by_model": [
    {"model": "claude-sonnet-4-6", "calls": 100, "cost_usd": 1.10},
    {"model": "claude-haiku-4-5", "calls": 42, "cost_usd": 0.13}
  ]
}
```

## Tasks
- [ ] Route handler: `backend/src/api/routes/admin_cost.py` (new file)
- [ ] Pydantic response schema with the shape above
- [ ] Both aggregation axes: workspace AND user (B2B billing readiness — the issue body of #180 only mentioned user, but workspace is the future billing unit)
- [ ] AuthZ: admin sees all; non-admin scoped to own workspaces via `PermissionService.check_workspace_access()`
- [ ] **DB-level aggregation only** — no fetch-and-sum-in-Python (could be 100k+ rows). Use SQL `GROUP BY date_trunc(period, started_at), workspace_id, user_id, llm_model` then assemble.
- [ ] Cost computation: join with pricing snapshot via `started_at` (point-in-time correctness from the schema child).
- [ ] OpenAPI tag: `Admin > Cost`
- [ ] Tests:
  - Integration with seeded `sleep_reports` covering multiple models + cached-input rows
  - admin/non-admin auth (admin sees all, non-admin sees scoped subset, anonymous gets 401)
  - Edge cases: empty period, mixed-model period, NULL-model rows (legacy) shown as `cost_unknown`

## Out of scope
- Frontend (→ dashboard child)
- Non-Sleep LLM cost sources — open as follow-up if needed
[...]

PR #527, opened 2026-05-01
feat(frontend): cost dashboard — admin + workspace pages (#473)

### LINKED ISSUE #473: feat(frontend): cost dashboard + CSV/JSON export (#180)
## Summary
Admin dashboard surfacing the cost aggregation API with charts, date range picker, and export.

Split from #180 (umbrella). **Depends on the API child** (must merge first).

## Page sketch
- Route: `/admin/cost`
- Top: date range picker (default last 30 days), period granularity toggle (day / week / month), workspace filter, user filter
- Middle: time-series chart (`cost_usd` + tokens stacked by model) — use the same chart library already in the admin UI for consistency
- Bottom: breakdown table (period × workspace × user × model) with sticky header
- Right rail: export buttons (CSV, JSON)

## Tasks
- [ ] Page component + route registration
- [ ] Date range picker (default: last 30 days, max 1 year window to prevent huge fetches)
- [ ] Period granularity toggle (day / week / month) wired to API `period` param
- [ ] Time-series chart of `cost_usd` and `tokens` by day, breakdown stacked by model
- [ ] Breakdown table with column sort, sticky header
- [ ] Export:
  - CSV: UTF-8 with BOM (Excel-friendly), comma-separated, RFC 4180 quoting
  - JSON: pretty-printed array of rows
- [ ] Empty state: "No sleep runs in this period."
- [ ] Loading state: skeleton, not spinner
- [ ] AuthZ gating: admin-only route guard (consistent with other admin pages)
- [ ] Tests (Vitest): chart prop construction, CSV escape edge cases (commas, quotes, newlines in user_id), JSON shape

## Out of scope
- Backend (→ API child)
- Real-time updates — initial version is page-load fetch only; refresh button is enough
- Cross-tenant export by non-admin — admin-only export to keep RBAC simple in v1

## UX notes
- Rows where `cost_usd` cannot be computed (legacy NULL-model `sleep_reports` from before the schema child) should render as `—` with a tooltip "Pre-migration data; cost unknown." Do not zero them out.
- Chart should clearly label the "cost unknown" gap if any periods are entirely NULL-model so the user understands why a month looks suspiciously cheap.

JFK:
Closes #473

## Summary
- `/admin/cost` (admin) and `/workspace/cost` (owner/admin) — two pages over a shared `CostDashboard` component, consuming the cost aggregation API shipped in #472.
- Sticky-NULL throughout the UI: rows with `cost_usd: null` (any contributing usage row had no resolved pricing) render as `—` in the table, gaps in the chart (`connectNulls={false}`) — never as a misleading $0. Mirrors the `BOOL_AND` semantics in the backend service.
- AuthZ layered defense-in-depth: Sidebar role gate + page-level role check + backend `require_admin` / `check_workspace_admin`. Path injection prevented via `encodeURIComponent`. Cross-workspace probe structurally prevented via `Omit<...AdminParams, "workspace_id">` typing on the workspace fetcher.

## Implementation notes
- New shared component at `frontend/src/components/cost/CostDashboard.tsx` (~360 LOC). Both pages are thin wrappers (~45 + ~75 LOC) that supply the right fetcher + auth gate.
- `WorkspaceCostPage` wraps its fetcher in `useCallback([currentWorkspaceId])` — without this, every `WorkspaceProvider` re-render would produce a new closure → `loadCost` recompiles → `useEffect` re-fires → spurious GET per re-render. Admin page passes `fetchAdminCostAggregation` directly (already module-level stable).
- `CostDashboard` short-circuits the fetch when client-side `rangeError` is non-null (inverted from/to, > 365 days) — keeps the more specific rangeError banner unobscured by a backend 400.
- Period toggle migrated to the shadcn `Tabs` primitive (`@/components/ui/tabs`) per `.claude/rules/frontend.md` ("In-page tabs MUST use the Tabs primitive").
- `isoDate` helper promoted into `lib/utils/datetime.ts` as `formatLocalDate` — centralizes the `toISOString().slice(0, 10)` trap (UTC shift breaks JST users) for any future date-range pickers.
- New API client at `lib/api/cost-aggregation.ts` with both `fetchAdminCostAggregation` and `fetchWorkspaceCostAggregation`. Workspace fetcher takes `workspaceId` as a separate path-bound argument so it can't be smuggled via params.
- New i18n keys under `admin.cost` namespace (en + ja). Workspace page uses `descriptionWorkspace` (vs admin's `description`) since the wording differs.
- Sidebar adds `DollarSign` icon + Cost entries to both the workspace section (`requiredWorkspaceRole: "admin"`) and admin section.

## Bonus cleanup in the same diff
- `UsageStats.tsx`: dropped the redundant pie chart (the list beneath it conveyed the same data; per-slice labels overlapped after the Top-10 rollup landed).
- `UsageStats.tsx`: Top-10 + "Other (N)" rollup memoized — the "Endpoint usage" panel was dumping all 25+ endpoints with most at 0.0%, drowning the signal.
- `UsageStats.tsx`: "Other %" computed as `(100 - sum(top.percentage))` clamped ≥0 — the naive sum of pre-rounded backend percentages drifts off 100% past ~15 entries.

## Out of scope
- **CSV/JSON export buttons** — `#472` did not ship an export endpoint; client-side blob download is a small follow-up if needed. The `#473` issue body listed export but the API contract doesn't support it server-side yet.
- **Workspace-scoped sleep reports** — needed in the workspace nav alongside Cost but requires a new backend endpoint. Filed separately as **#526**.

## Tests
- `CostDashboard.test.ts` (new): 7 tests for `buildChartData` (empty input, multi-row collapse, ISO sort, sticky-NULL per-field/per-bucket/order-independent), 3 for `formatCost` (positive, **$0 NOT em-dash**, null em-dash).
- `lib/utils/datetime.test.ts` (extended): 3 tests for `formatLocalDate` covering padding, the `toISOString` trap, and basic format.
- `next build` succeeds with `/admin/cost` and `/workspace/cost` prerendered as static.
- `tsc --noEmit` clean. `make lint` + `make type-check` clean (frontend has no separate lint since Next.js 16).

## Pre-PR review summary
- gate2 mode: not run (frontend-only follow-up to backend `#472` which had full gate2 advisor-only review)
- review provider: copilot (planned)

🤖 Generated via /self-review + /simplify

### DIFF
--- a/frontend/src/app/(authenticated)/admin/cost/page.tsx
+++ b/frontend/src/app/(authenticated)/admin/cost/page.tsx
@@ -1,0 +1,47 @@
+"use client";
+
+/**
+ * Admin Cost Aggregation Dashboard (Issue #473).
+ *
+ * Cross-workspace view: admin-only. Renders the shared
+ * ``CostDashboard`` against the admin endpoint, which returns one row
+ * per (period × workspace × user × model × source × paid_by) — hence
+ * the workspace column is shown for cross-workspace comparison.
+ *
+ * Workspace owner / admin self-service is at ``/workspace/cost`` and
+ * uses the workspace-scoped endpoint shipped in #472.
+ */
+
+import { useTranslations } from "next-intl";
+import { PageContainer } from "@/components/common/PageContainer";
+import { PageHeader } from "@/components/common/PageHeader";
+import { ErrorBanner } from "@/components/common/ErrorBanner";
+import { CostDashboard } from "@/components/cost/CostDashboard";
+import { useAuth } from "@/contexts/AuthContext";
+import { isAdmin } from "@/lib/auth/rbac";
+import { fetchAdminCostAggregation } from "@/lib/api";
+
+export default function AdminCostPage() {
+  const t = useTranslations("admin.cost");
+  const { user } = useAuth();
+
+  if (!isAdmin(user)) {
+    return (
+      <PageContainer>
+        <PageHeader title={t("title")} />
+        <ErrorBanner error={t("errors.forbidden")} />
+      </PageContainer>
+    );
+  }
+
+  // Stable module-level reference passed directly — no closure, no
+  // useCallback needed. Keeps CostDashboard's effect dep set quiet.
+  return (
+    <CostDashboard
+      title={t("title")}
+      description={t("description")}
+      fetchData={fetchAdminCostAggregation}
+      showWorkspaceColumn
+    />
+  );
+}
--- a/frontend/src/app/(authenticated)/admin/cost/page.tsx
+++ b/frontend/src/app/(authenticated)/admin/cost/page.tsx
@@ -17,6 +17,7 @@ import { useTranslations } from "next-intl";
 import { PageContainer } from "@/components/common/PageContainer";
 import { PageHeader } from "@/components/common/PageHeader";
 import { ErrorBanner } from "@/components/common/ErrorBanner";
+import { LoadingState } from "@/components/common/LoadingState";
 import { CostDashboard } from "@/components/cost/CostDashboard";
 import { useAuth } from "@/contexts/AuthContext";
 import { isAdmin } from "@/lib/auth/rbac";
@@ -24,7 +25,19 @@ import { fetchAdminCostAggregation } from "@/lib/api";
 
 export default function AdminCostPage() {
   const t = useTranslations("admin.cost");
-  const { user } = useAuth();
+  const { user, isLoading: authLoading } = useAuth();
+
+  // Render a loading skeleton until AuthProvider resolves. Without
+  // this, real admins see a brief "Admin role required" flash on every
+  // first paint because user is null until getCurrentUser() returns.
+  if (authLoading) {
+    return (
+      <PageContainer>
+        <PageHeader title={t("title")} />
+        <LoadingState lines={6} />
+      </PageContainer>
+    );
+  }
 
   if (!isAdmin(user)) {
     return (
--- a/frontend/src/app/(authenticated)/admin/cost/page.tsx
+++ b/frontend/src/app/(authenticated)/admin/cost/page.tsx
@@ -5,8 +5,9 @@
  *
  * Cross-workspace view: admin-only. Renders the shared
  * ``CostDashboard`` against the admin endpoint, which returns one row
- * per (period × workspace × user × model × source × paid_by) — hence
- * the workspace column is shown for cross-workspace comparison.
+ * per (period × workspace × user) with model/source rollups in
+ * ``cost_breakdown_by_model`` / ``cost_breakdown_by_source`` arrays —
+ * hence the workspace column is shown for cross-workspace comparison.
  *
  * Workspace owner / admin self-service is at ``/workspace/cost`` and
  * uses the workspace-scoped endpoint shipped in #472.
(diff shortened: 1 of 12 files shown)


JFK:
**Loop 1**: addressed all 5 actionable findings.

1. `CostDashboard.loadCost`: early-return branches now clear stale `error` state (workspace-switch transient + rangeError correction no longer leaves a stacked banner).
2. `UsageStats` endpoint chips migrated from inline `style={{ backgroundColor }}` to Tailwind `SWATCH_BG_CLASSES` (per .claude/rules/frontend.md). Removed the now-unused `COLORS` hex array (its only consumer was the dropped pie chart).
3. `cost-aggregation.ts` module docstring fixed — said v1 frontend was admin-only but the same PR ships /workspace/cost.
4. Added `admin.cost.errors.forbiddenWorkspace` (en + ja); workspace page routes through it instead of borrowing the admin-only message.
5. `rangeError` day-window calc switched to `Date.UTC` anchors — local-tz construction drifted ±1 day across DST transitions at the 365-day boundary.

Tests still 23 pass, type-check clean. No behavior change beyond the listed fixes.

JFK:
**Loop 2**: addressed all 3 findings.

1. CostDashboard table section now renders ErrorBanner on fetch failure (was falling through to EmptyState because rows is cleared in the catch — conflated error with zero results).
2. `loadCost` early-returns now `setRows([])` alongside the loop-1-introduced `setError(null)`. A workspace switch / mid-correction invalid range no longer leaves stale rows on screen.
3. `UsageStats` "Other (N)" label is now i18n'd via `usageStats.otherEndpointsLabel` (en + ja). useMemo deps updated to include `t`.

JFK:
**Loop 3**: 1/2 actionable.

1. ✅ admin/cost/page.tsx docstring drift — said "one row per (period × workspace × user × model × source × paid_by)" but API returns one row per (period × workspace × user) with model/source rollups in breakdown arrays. Fixed.
2. 🔁 Table error branch (line 382) — duplicate of loop 2 finding, already addressed in commit a88197ce (line 380 has `error ? <ErrorBanner ...> : rows.length === 0 ? <EmptyState ...> : <Table>`). Verified in place. Likely Copilot scanned the pre-loop-2 file.

JFK:
**Loop 4**: addressed all 4 findings.

1. `AdminCostPage` now reads `isLoading` from useAuth and renders LoadingState until auth resolves — eliminates the "Admin role required" flash on initial paint.
2. `CostDashboard` chart/table sections now branch on `rangeError !== null` first, rendering a hint placeholder (`rangeErrorPlaceholder`) instead of falling through to "No cost data" when the user is mid-correction.
3. Table section migrated from `LoadingState` to `TableLoadingState` (per .claude/rules/frontend.md). Chart section keeps `LoadingState` (panel-shaped, not table-shaped).
4. `Unknown error` hardcoded fallback in `loadCost` catch → `t("errors.unknown")` with new `admin.cost.errors.unknown` key in en + ja. `t` added to `loadCost` useCallback deps.

JFK:
**Loop 5** (final, max_loops=5): 2 findings, both addressed.

1. CostDashboard rangeError comment said "Server still validates" but #472 backend doesn't enforce a max window. Updated comment to reflect UI-only soft cap. Backend enforcement filed as **#528** (frontend cap remains so accidental huge UI queries don't even leave the browser).
2. WorkspaceCostPage now distinguishes "no workspace" from "wrong role" — added `errors.noWorkspaceSelected` (en + ja) for the zero-workspaces case so a brand-new account doesn't see a misleading "owner/admin required" banner.

**Loop totals across this PR**:
- L1: 5/5 actionable
- L2: 3/3
- L3: 1/2 (1 dup of L2 from Copilot scan cache)
- L4: 4/4
- L5: 2/2

Total: 13 unique findings addressed. Tests 23 pass, type-check clean throughout. Reaching max_loops=5; PR ready for human review.

Issue #563, opened 2026-05-06
chore(storage): re-review GC ↔ undelete race when undelete API is added


JFK:
## Context

The nightly GC sweep introduced in #552 hard-deletes `file_objects` rows where:

```
status = 'uploaded' AND deleted_at IS NOT NULL AND deleted_at < now() - 7 days
```

There is currently **no undelete API** — soft-deleted files are unrecoverable after the user calls delete. The 7-day grace exists only as a safety buffer against operator error / accidental DELETE FROM Postgres, not as a user-facing trash bin.

If we ever add an undelete endpoint (e.g., a "Trash" UI that lets users restore soft-deleted files within 7 days), a new race surface opens:

- t=0: user soft-deletes file A → `deleted_at = now()`
- t=6d 23h: GC sweep starts loading candidates (deleted_at < now - 7d at this exact threshold)
- t=6d 23h 30m: user undeletes file A → `deleted_at = NULL`, `status = 'uploaded'`
- t=6d 23h 31m: GC reaches file A in its candidate list, calls `storage.delete_object` → bytes gone

The sweep loop loads candidates once at the top, so an undelete during the loop's iteration won't be reflected in the in-memory list.

## Scope

This issue is a **placeholder for future re-review**. Do NOT work on it speculatively. When (and if) an undelete API is filed:

1. Re-evaluate the GC's per-row decision: should it re-check `deleted_at IS NOT NULL` immediately before R2 delete?
2. Add a SQLAlchemy-level concurrency guard (e.g., `WHERE deleted_at IS NOT NULL` in the DELETE statement, or row-level lock during the loop)
3. Document the race window in the undelete API's design doc

## Reference

- Parent: #485 (R2 file storage Phase 1)
- Sibling: #552 (where the GC was introduced)
- CSO Gate2 review note (current PR `#552-batch/552-554`): "将来 undelete API を追加する場合、GC との race を再 review"

JFK:
PM triage: undelete API は未実装 (#563 本文「when undelete API is added」)。re-review trigger が無いため backlog 化。undelete 機能起票時に依存タスクとして milestone へ移送する。

JFK:
[...]

Issue #562, opened 2026-05-06
feat(storage): GC sweep — column-restricted SELECT once inline_bytes is populated


JFK:
## Context

Issue #552 (Phase 1.5 nightly GC) ships with `select(FileObject).where(...).limit(500)` which eagerly loads the full ORM row including `inline_bytes` (BYTEA, nullable, currently always NULL in Phase 1).

Once Phase 1.5 (the as-yet-unfiled inline-storage path) starts populating `inline_bytes`, a 500-row sweep can pull MBs of BYTEA into memory per nightly tick. The partial index `idx_file_objects_soft_deleted_gc` doesn't help — it's the row materialization that bloats.

## Scope

When (and only when) Phase 1.5 inline storage lands:

- Replace `select(FileObject).where(...).limit(500)` in `sweep_soft_deleted_files` with a column-restricted form, e.g. `select(FileObject.id, FileObject.storage_key).where(...).limit(500)`
- Adapt the loop to use the id/storage_key tuples
- For DB hard-delete, use `db.execute(delete(FileObject).where(FileObject.id == id))` instead of loading the ORM object first
- Update tests to match (the current `_make_file` MagicMock pattern still works)

## Out of scope

This is a **defensive follow-up gated on Phase 1.5**. Do not work on this until inline storage actually lands — without `inline_bytes` populated, the change is pure churn.

## Reference

- Parent: #485 (R2 file storage Phase 1)
- Related: #552 (where the GC was introduced)
- /simplify Phase 6 review finding (CSO + efficiency reviewers)

JFK:
Closing as **won't do**.

## Reason

The premise of this issue — that `inline_bytes` will eventually be populated and so the GC sweep needs a column-restricted SELECT — does not hold against the actual storage roadmap on record:

- **#485 R1 design** decided **R2 over PG bytea**. Rationale captured at decision time: bytea bloats backups, slows VACUUM/replication, hits the 1 GB-per-row TOAST ceiling. R2 was chosen for free cross-device egress, vendor consolidation, and ~10x cheaper storage.
- **Phase 1.5 (v0.15.1)** turned out to be a hardening bundle (#552 7d cleanup, #553 content_type allow-list, #554 TOCTOU, #555 attachments deprecation, #556 sha256 binding). **None of them implement inline storage** — the `pg_inline` path is not on Phase 1.5.
[...]

PR #564, opened 2026-05-06
feat(storage): Phase 1.5 GC + quota TOCTOU (#552, #554)

### LINKED ISSUE #552: feat(storage): Phase 1.5 — 7-day soft-delete + R2 binary cleanup nightly GC
## Context

Issue #485 (Phase 1, shipped via #551 squash `e8233f99`) implements the orphan
sweeper for ``status='reserved'`` rows past their 1-hour grace, but does NOT
yet implement the second half of the acceptance criteria:

> [ ] Nightly GC job for soft-deleted + dangling objects

Specifically: when a user deletes an uploaded file, the row gets
``status='uploaded' AND deleted_at NOT NULL`` and the workspace's quota counter
is decremented immediately (R5). The R2 binary, however, lingers indefinitely
— the issue body called for "DELETE object after grace period (7d?)".

## Scope

- New scheduled task (or extend ``backend/src/tasks/file_tasks.py``): nightly
  scan ``file_objects WHERE deleted_at IS NOT NULL AND deleted_at < now() - 7d``
- For each match: ``DELETE FROM r2`` via ``BlobStorageProtocol.delete_object``
- After R2 delete: hard-delete the ``file_objects`` row (or set a tombstone
  flag) so subsequent runs don't re-evaluate the same row
- Must handle: R2-already-deleted (idempotent), R2 transient errors
  (best-effort, retry on next run), workspace deleted (cascade)

## Acceptance

- [ ] New cron / interval job registered alongside the orphan sweeper
- [ ] Idempotent on re-run (same row swept twice = no-op the second time)
- [ ] R2 errors logged but don't poison the rest of the batch
- [ ] Tests: happy path + R2 error tolerance + already-deleted idempotency
- [ ] Production deploy: enable cron + observe the first sweep cycle in logs

## Related

- Parent: #485 (Phase 1)
- Code: ``backend/src/tasks/file_tasks.py`` (orphan sweeper precedent)
- Design: gate1 R3 state machine (`reserved → uploaded → failed`), R5
  immediate quota release on soft-delete

### LINKED ISSUE #554: feat(storage): Phase 1.5 — TOCTOU hardening on storage_quota_service.reserve_storage_bytes
## Context

Issue #485 (Phase 1, shipped via #551 `e8233f99`) ships a Redis-backed
storage quota with a documented TOCTOU window:

```python
# storage_quota_service.py reserve_storage_bytes
current = await _get_or_seed_counter(...)   # READ
if current + size_bytes > quota_bytes:      # CHECK
    raise QuotaExceededError(...)
new_total = await incrby_counter(key, size_bytes)  # ACT
```

Two concurrent uploads for the same workspace both READ the same
``current`` value, both pass the cap check, and both INCRBY — over-committing
by up to one ``size_bytes`` (gate1 reviewer B-3, fail-open posture
acknowledged).

For the Free 100 MiB tier this is practically exploitable: two concurrent
60 MiB uploads can both succeed, leaving the workspace at 120 MiB.

## Scope

Make the read-check-incr atomic via a Redis Lua script:

```lua
local current = tonumber(redis.call('GET', KEYS[1]) or '0')
local size = tonumber(ARGV[1])
local quota = tonumber(ARGV[2])
if current + size > quota then
  return -1  -- exceeded
end
return redis.call('INCRBY', KEYS[1], size)
```

Wrap call site in ``storage_quota_service.reserve_storage_bytes``. Keep the
existing seed-from-DB path for cache misses (the Lua script gates only the
hot path).

## Acceptance

- [ ] Lua script for atomic check-then-INCRBY committed
- [ ] ``reserve_storage_bytes`` invokes the script via ``EVAL``
- [ ] Seed-from-DB path unchanged (cold start still works)
- [ ] Concurrent reservation test (two parallel reserves at the cap)
      proves only one succeeds
- [ ] Fail-open posture preserved: Lua failure / Redis down → log + skip
      check (current behavior)

## Trade-offs

The current fail-open posture is documented as acceptable for Phase 1
("Redis outage MUST NOT block uploads"). This hardening tightens the
race but does NOT change fail-open behavior on Redis errors — both
failure modes are independent.

## Related

- Parent: #485 (Phase 1)
- Gate1 reviewer finding B-3 (acknowledged + deferred)
- Code: ``backend/src/services/storage_quota_service.py:_get_or_seed_counter``

JFK:
Closes #552
Closes #554

## Summary

- **#554 — TOCTOU hardening**: replaces the GET → check → INCRBY 3-step in `storage_quota_service.reserve_storage_bytes` with a single atomic Redis Lua EVAL. Closes the race where two concurrent reservations both pass the cap check then both INCRBY (over-commit). Fail-open posture and `_get_or_seed_counter` semantics preserved.
- **#552 — Soft-delete nightly GC**: new `sweep_soft_deleted_files` cron (03:15 UTC daily) hard-deletes `file_objects` rows soft-deleted >7 days ago after deleting their R2 binary. Per-row commit, R2-first ordering (so a transient R2 failure leaves the row for next sweep retry rather than orphaning the binary). Does NOT call `release_storage_bytes` — R5 already released quota at soft-delete time.
- New partial index `idx_file_objects_soft_deleted_gc` keeps the nightly query off a full table scan.
- (Bundled) docs alignment with Karpathy's LLM Wiki pattern — repositions Kagura as a team-scale LLM Knowledge Base implementing the 5-layer pattern (Ingest / Compile / Index / Query / Enhance). No code impact; framing is implementation-consistent (compounding loop is honestly noted as explicit/opt-in).

## Implementation notes

- **Lua script convention**: inline string constant at module level (mirrors `tasks/neural_calibration._DEDUP_RELEASE_SCRIPT`). No new `lua/` directory.
- **`_atomic_check_and_incr`**: catches `Exception` (broad on purpose — `ValueError` from `int(None)` if Lua ever returns nil must also flow through fail-open) → wraps as project `RedisError` so the outer `except RedisError` in `reserve_storage_bytes` catches it via the existing branch.
- **R2-first ordering in GC**: deliberately inverts the orphan sweeper's DB-first pattern. R2 `DELETE` is S3-spec idempotent (returns 204 whether or not the key exists), so retries on next sweep cost nothing. The asymmetry vs orphan sweeper is documented inline.
- **LIMIT 500** cap on the GC select bounds memory on bulk-delete bursts; tail rolls to next nightly tick.
- **Migration `e04_552_gc_index`**: `CREATE INDEX` non-CONCURRENTLY takes a SHARE lock on `file_objects` while the index builds. Sub-second at Phase 1 scale; deploy-note in the migration docstring points at `postgresql_concurrently=True` if the table grows large.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: green via /ask
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/552-batch-552-554.gate1.md`
- `~/.claude/cache/gh-issue-driven/552-batch-552-554.gate2.md`

## Follow-ups filed during review

- #562 — column-restricted GC SELECT (gated on Phase 1.5 `inline_bytes` populate)
- #563 — re-review GC ↔ undelete race when undelete API is added

## Test plan

- [x] `pytest tests/services/ tests/tasks/` — 680 passed, 1 pre-existing skip
- [x] `ruff check` — All checks passed
- [x] `pyright` — 0 errors
- [x] `alembic heads` — `e04_552_gc_index` is the new head
- [ ] Production deploy: run `alembic upgrade head`, observe first 03:15 UTC GC tick in logs


JFK:
Loop 1: addressed 4 of 5 Copilot findings.

**Applied**:
- `storage_quota_service.py:166` — Lua returns `{result, current}` tuple; error message uses Lua-time value instead of stale seed-time
- `file_tasks.py:226` — replaced ORM `db.delete(f)` with idempotent SQL `DELETE … WHERE id=… AND deleted_at IS NOT NULL` (multi-replica safe + closes the future-undelete race window from #563)
- `README.md:77` / `README.ja.md:77` — `REST /memory` → `REST /api/v1/memory`

**Deferred** (rationale, not blocking):
- `file_tasks.py:195` — column-restricted `select(FileObject.id, FileObject.storage_key)` is the right move once Phase 1.5 starts populating `inline_bytes`. Currently `inline_bytes` is always NULL (Phase 1 is R2-only), so a pre-emptive change here is churn without payoff. Already filed as **#562** with the Phase 1.5 gate explicit.

682 unit/integration tests pass. New regression coverage: `test_at_ceiling_uses_lua_current_not_seed`, `test_lost_race_with_other_replica_is_no_op`.

JFK:
Loop 2: addressed 3 new actionable findings + applied identical fix to a sibling pre-existing bug.

**Applied**:
- `storage_quota_service.py:102` — explicit `except asyncio.CancelledError: raise` before the broad clause in `_atomic_check_and_incr`. Wrapping `CancelledError` as `RedisError` and then swallowing via fail-open would break cooperative cancellation during deploys / SIGTERM.
- `file_tasks.py:217` — same fix for the GC R2-delete loop. **Bonus**: applied the identical fix to `sweep_orphan_files`'s R2 delete loop (`file_tasks.py:138`) since the orphan sweeper had the same pre-existing bug — fixed in the same commit for consistency rather than splitting into a separate PR.
- `storage_quota_service.py:93` — module-level `_reserve_quota_script` cache, lazy-initialized on first call. `register_script` allocation + SHA1 hash now happens once per process instead of per-reservation.
- `file_tasks.py:52` — rephrased the `_GC_RETENTION_SECONDS` docstring from "trash undo window" to "operational safety buffer" since there's no user-facing undelete API today (#563 tracks the future undelete + retention semantics shift).

**Re-flagged but still deferred** (same rationale as loop 1):
- `file_tasks.py:195` (column-restricted SELECT) — gated on Phase 1.5 inline storage. Filed as **#562**. The column is currently always NULL so a pre-emptive change is churn.

684 tests pass. New regression coverage:
- `test_cancelled_error_propagates_not_swallowed` (quota reserve path)
- `test_r2_cancelled_error_propagates` (GC R2 delete path)

JFK:
Loop 3: addressed 3 new actionable findings.

**Applied**:
- `storage_quota_service.py:42` — `_reserve_quota_script` now keyed by `id(client)` so the cache auto-invalidates if the Redis singleton is reset (tests via `close_redis()`). Production never resets, but the defensive form costs nothing and removes the test-isolation footgun.
- `file_tasks.py:287` — explicit `timezone="UTC"` on the GC `CronTrigger`. The parent `AsyncIOScheduler` is already UTC so this is currently a no-op, but writing it at the trigger keeps the schedule unambiguous if the scheduler default ever changes.
- `docs/architecture.md:11-15` — updated the 5-layer mapping's "Code location" column to repo-relative paths (`backend/src/api/routes/memory.py` etc.) so a reader can `cd` directly.

**Stale (already fixed in loop 2 — Copilot brought the unresolved threads forward)**:
- `storage_quota_service.py:119` `CancelledError` re-raise — committed `454139a8` (loop 2)
- `file_tasks.py:233` R2-delete `CancelledError` re-raise — committed `454139a8` (loop 2)

If you can resolve those two threads on the GitHub side, the next loop will be cleaner.

**Re-flagged but still deferred** (same rationale as loops 1 + 2):
- `file_tasks.py:205` column-restricted SELECT — Phase 1.5 gate, filed as **#562**

684 tests pass.

JFK:
Loop 4: addressed 2 new actionable findings (both docs).

**Applied**:
- `docs/architecture.md:15` — `claude-skills/session-summary` → `claude-skills/session-summary.md` (the actual file extension)
- `README.ja.md:41` — `page` → `ページ` and `約 100 page` → `約 100 ページ` (Japanese-consistent wording)

**Stale re-flags** (already fixed, Copilot has not yet resolved on its side):
- `storage_quota_service.py:124` `CancelledError` — fixed loop 2 (`454139a8`)
- `file_tasks.py:233` R2 `CancelledError` — fixed loop 2 (`454139a8`)
- `file_tasks.py:291` `CronTrigger timezone='UTC'` — fixed loop 3 (`cf727c0e`)

**Re-flagged but still deferred**:
- `file_tasks.py:205` column-restricted SELECT — Phase 1.5 gate, **#562**

Convergence trend: loop 1 = 5 new, loop 2 = 3 new + 1 reflag, loop 3 = 3 new + 3 reflags, loop 4 = 2 new + 4 reflags. Approaching `no_actionable_feedback` exit on loop 5 if only stale re-flags remain.

Issue #565, opened 2026-05-06
feat(storage): ClamAV virus scan for uploaded files (gated on public signup)


JFK:
## Context

Split from #553 (2026-05-06 PM reorg). The original Phase 1.5 deferral
in #485 bundled ClamAV with the content_type allow-list, but at the
current scale and architecture the two have very different cost/value
profiles:

- **Allow-list**: cheap (10-line FastAPI guard), shipped under #553
- **Virus scan**: heavy (ClamAV daemon, signature updates, scan latency,
  false-positive handling, quarantine state machine, ops on-call burden)

## Why deferred

Threat surface at the moment is narrow:

1. Closed beta — open signup is gated (v0.14.1 anchor decision)
2. Files are partitioned per user; no cross-user sharing exists
3. R2 access is via signed URLs with expiry, not a public CDN
4. Uploaded blobs are not executed server-side; ingest parsers are
   sandboxed independent of any virus-scan layer

A user uploading malware to their own context primarily harms only
themselves on later download — outside the threat model that ClamAV
addresses.

## Re-evaluation triggers (any one of these flips this back to active)

- [ ] Open signup is enabled (untrusted upload sources)
- [ ] Cross-user file sharing or any public-link feature lands
- [ ] R2 buckets become anonymously readable for any reason
- [ ] Server-side processing of uploaded content begins (e.g. PDF text
      extraction, image OCR) without sandboxing — parser CVE risk
      escalates the case for pre-parse virus scan
- [ ] Compliance / customer requirement makes a written virus-scan
      policy mandatory

When any trigger fires, file this back into the next available
release-anchored milestone with the original scope below.

## Original scope (from #553)

- Async post-upload scan (after ``confirm_upload`` succeeds): submit
  R2 object to ClamAV (or Cloudflare Worker scanner)
- On detection: transition ``status="uploaded" → "quarantined"``
  (new enum value, partial unique index update needed), notify ops
- Decision pending at re-activation time: gate downloads on scan
  completion (delay) vs scan-on-write best-effort (faster)
- ``quarantined`` status added to file_objects state machine + migration
- Quarantined files: download blocked (404 / quarantined), cleanup
  flow defined
- Public beta cannot open until this lands

## Original CSO note (Gate2 finding F-2)

This was the documented "Phase 1.5 before public beta opens uploads to
untrusted users" deferral — Issue #485 body called it out explicitly.
Phase 1 ship was acceptable because the SDK is the only caller
[...]

PR #566, opened 2026-05-06
feat(storage): content_type allow-list for file uploads (#553)

### LINKED ISSUE #553: feat(storage): Phase 1.5 — content_type allow-list for file uploads
## Context

Issue #485 (Phase 1, shipped via #551 `e8233f99`) accepts arbitrary
``content_type`` values up to 255 chars, by design. Phase 1.5 was the
designated place to lock the accepted set down so that arbitrary
executable / archive formats cannot land in R2 — even from authenticated
workspace members.

The companion ClamAV / virus-scan work was originally bundled in this
issue but has been split off (see Related) — at closed-beta scale with
per-user partitioning, signed URLs, and no server-side execution of
uploaded blobs, the virus-scan operational cost outweighs the residual
risk. Allow-list remains worthwhile because it is cheap and provides
defense-in-depth against parser CVEs in any downstream ingest pipeline.

## Scope

- Mirror ``api/routes/attachments.py``s ``ALLOWED_CONTENT_TYPES`` set
  (``image/png``, ``image/jpeg``, ``image/gif``, ``application/pdf``,
  ``text/plain``, ``text/markdown``, ``text/csv``, ``application/json``)
- Apply at REST + MCP boundaries (validate before reaching
  ``FileStorageService.reserve_upload``) — workspace policy is uniform
- Make the allow-list configurable via ``settings`` so self-hosted forks
  can customize per-deployment

## Acceptance

- [ ] Allow-list enforced at both REST and MCP entry (test parity)
- [ ] Configurable via ``settings.allowed_file_content_types`` env var
- [ ] Rejection returns 415 Unsupported Media Type with the violating
      content_type echoed back (no PII risk — already user-supplied)
- [ ] Test coverage: allowed + rejected at both entry points

## Related

- Parent: #485 (Phase 1)
- Split-off: virus scan moved to a separate, defer-with-trigger issue
  (closed-beta threat surface analysis, 2026-05-06 reorg)
- Precedent: ``backend/src/api/routes/attachments.py:30-40`` for the
  allow-list shape

JFK:
Closes #553

## Summary

- Lock down `/api/v1/files/*` and the MCP file tools to a configurable MIME allow-list at the `FileStorageService.reserve_upload` boundary — the single chokepoint where REST and MCP both funnel, so policy is enforced uniformly without code duplication.
- New `UnsupportedMediaTypeError` (HTTP 415, `error_code="MEDIA-001"`) — distinct from `ValidationError` (422) so REST returns the semantically correct status and MCP gets a dedicated `unsupported_media_type` vocab. The `MEDIA-` namespace separates capability/policy rejection from shape/format errors so SDK error routing stays clean.
- `ALLOWED_FILE_CONTENT_TYPES` env var (comma-separated, case-insensitive per RFC 6838, fail-closed when empty); default mirrors the closed-beta MIME set used by `attachments.py`.
- Strip RFC 7231 media-type parameters before compare so browser `fetch()` and multipart uploads (`text/plain; charset=utf-8`) still match the bare type/subtype allow-list — without this, real-world clients would silently see 415 and operators would feel pressure to widen the allow-list (security regression path).

## Implementation notes

- Validation runs after `size_bytes` / `sha256` length checks and **before** workspace load + quota reserve — disallowed types short-circuit without DB or Redis I/O.
- Existing R2 `file_objects` are unaffected: read paths (`download-url`, `list_files`) do not trigger validation. Only **new** uploads via `reserve_upload` are subject to the allow-list. Ops impact is therefore zero for currently-uploaded data; only future uploads of types outside the default allow-list (e.g. `application/octet-stream`, `image/webp`) would be rejected.
- Phase 1.5 ClamAV virus scan was split off (#565) at closed-beta scale where signed-URL + per-user partition + no server-side blob execution makes the operational cost outweigh the residual risk. Allow-list is the cheap defense-in-depth half of the original Phase 1.5 hardening (the heavy half — virus scanning — has documented re-evaluation triggers in #565).
- 5 service-layer tests + 1 REST 415 mapping test + 1 MCP `unsupported_media_type` vocab test (53 tests pass total).

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/553-feat-feat-storage-phase-1-5-content-type-allo.gate2.md`


JFK:
Loop 1: addressed all 3 actionable findings.

- **mcp_server/tools/files.py** — clarified the comment so it no longer claims the allow-list is reachable via `exc.details["allowed"]` from the MCP envelope; MCP envelope only carries `(error_code, message)`. MCP envelope unification is a follow-up.
- **services/file_storage_service.py** — added two pre-checks before the allow-list compare:
  1. Reject control characters (`\r`/`\n`/`\x00`) in the raw content_type as `ValidationError` (422). Closes the header-injection vector against the R2 SigV4-signed `ContentType`.
  2. Validate type/subtype shape with an RFC 7231 token regex; malformed values (no slash, empty type/subtype, garbage chars) now raise `ValidationError` (422), preserving the 415-policy vs 422-shape semantic split.
- **config/settings.py** — `allowed_file_content_types_set` now strips RFC 7231 parameters per entry, so an env var value like `text/plain; charset=utf-8` normalizes to `text/plain` and matches a bare upload (parity with the upload-side strip).

11 new service tests (parametrize + env-var-stripping); 64 tests pass total. Lint + pyright clean.

JFK:
Loop 2: addressed 2 of 3 actionable findings; 1 was already fixed in loop 1.

**1. `mcp_server/tools/files.py:52` — MCP envelope forwards `details` (loop 2 fix)**

The original loop 1 comment claimed envelope unification was a follow-up, but `_error_response(error, message, **extra)` in `_helpers.py:82` already accepts arbitrary fields, and `analysis._gate_error_response` forwards `QuotaExceededError` details via the same path. Updated to forward `content_type` + `allowed` so MCP clients mirror the REST 415 body shape:

```python
return _error_response(
    "unsupported_media_type",
    str(exc),
    content_type=exc.details.get("content_type"),
    allowed=exc.details.get("allowed", []),
)
```

Test updated to assert the new fields.

**2. `utils/exceptions.py:155` — docstring drift (loop 2 fix)**

`UnsupportedMediaTypeError` docstring now describes the actual normalized comparison: RFC 7231 parameters stripped, RFC 6838 case-insensitive, malformed shape returns 422 not 415. Callers reading the exception class will see what's actually enforced.

**3. `services/file_storage_service.py:188` — already addressed in loop 1 (skip)**

The loop 1 commit (43361565) already added both:
- Control-character rejection (`\r` / `\n` / `\x00` in raw `content_type` → `ValidationError` 422)
- Type/subtype shape validation via RFC 7231 token regex (rejects empty type, empty subtype, missing slash, garbage chars → `ValidationError` 422)

No additional changes needed; the rejection-before-presigning is already in place. The Copilot review on commit 43361565 may have surfaced this comment from the diff context rather than the live code state.

64 tests pass; lint + pyright clean.

JFK:
Loop 3: addressed the malformed allow-list leakage.

**`config/settings.py:384` — fail-fast at boot + extract shared helper**

Added a `model_validator` that rejects malformed `ALLOWED_FILE_CONTENT_TYPES` entries (no slash, embedded space, garbage chars) at app startup. Empty / whitespace / comma-only still pass through (preserved fail-closed semantics — `""` resolves to empty set = every upload rejected).

This closes the leak Copilot identified: malformed entries can no longer reach `UnsupportedMediaTypeError.details["allowed"]` because boot fails first when the operator sets a bad env var.

Also extracted `MEDIA_TYPE_RE` + `normalize_media_type` from `services/file_storage_service.py` to a new `utils/media_types.py` (rule of three: request-time validation, settings parser, boot validator now share one regex). Drift between the three call sites is now structurally impossible.

Tests: 5 new boot-validator assertions covering valid / empty / whitespace / parameter-laden / four malformed shapes; 65 tests pass total. Lint + pyright clean.

JFK:
Loop 4: addressed the MCP-only oversize input gap.

**`services/file_storage_service.py:184` — service-layer length guards**

Added length checks for `filename` (≤512) and `content_type` (≤255) at the top of `reserve_upload`, matching the DB column caps in `FileObject`. Now REST + MCP share the same hard limits — pydantic enforces these at REST via `FileReserveRequest`, but MCP's `str(args["..."])` coercion bypassed pydantic, so oversized fields would have:

1. Pass the allow-list check (e.g. `"application/pdf; " + "a"*256` strips to `"application/pdf"`)
2. Fail at `flush()` / `commit()` with a column-length DB error → opaque 500

With the guards, an MCP caller passing oversized values gets a clean 422 `ValidationError` at the boundary instead.

2 new service tests (oversize filename + oversize content_type that normalizes past the allow-list); 67 tests pass total. Lint + pyright clean.

JFK:
Loop 5: addressed both findings.

**1. `services/file_storage_service.py:172` — docstring `Raises:` drift**

Updated `reserve_upload`'s `Raises:` block to document `UnsupportedMediaTypeError` and expanded the `ValidationError` description to cover everything the service now raises (size/sha256 length, oversized `filename` or `content_type`, control characters, malformed type/subtype shape). REST + MCP layer authors now have an accurate exception contract.

**2. `tests/services/test_file_storage_service.py:299` — test isolation**

Added an `autouse` fixture in `TestReserveUpload` that pins the allow-list to the canonical default for every test in the class. Without it, developers (or CI runners) with `ALLOWED_FILE_CONTENT_TYPES` set in their shell env would see non-deterministic failures because tests assume `image/png` / `application/pdf` are allowed.

Inner `monkeypatch.setattr` calls inside tests that need a *different* allow-list (`test_empty_allowlist_rejects_all`, `test_env_allowlist_strips_parameters_in_entries`) still work — pytest stacks the undos in LIFO order, so the per-test override applies on top and unwinds first.

67 tests pass; no behavior change for the canonical config. Lint + pyright clean.

This is loop 5 (per-run cap). If this round is also `COMMENTED`, I'll resume via a fresh `/gh-issue-driven:review` invocation; if `APPROVED` or `no_actionable_feedback`, the loop converges and the PR is ready for human review.

Issue #586, opened 2026-05-09
refactor(api): lazy UserPlan creation in get_current_usage GET handler


JFK:
## Background

PR for #570 (`refactor(api): EffectiveQuotaService API simplification`) made `EffectiveQuotaService.get_effective_quotas()` pure-read by removing a GET-time COMMIT. During the gate1 `/dx-lead` horizontal scan we found a sister antipattern in the route layer:

`backend/src/api/routes/usage.py:386 get_current_usage` performs lazy `UserPlan` row creation inside a `@router.get(...)` handler:

```python
result = await db.execute(select(UserPlan).where(UserPlan.user_id == user_id))
plan = result.scalar_one_or_none()
if not plan:
    plan = UserPlan(user_id=user_id, plan_name="free", ...)
    db.add(plan)
    await db.commit()    # ← GET 経路で COMMIT
```

Same antipattern class as #570:
- HTTP spec: GET should be safe and idempotent (RFC 9110)
- DX surface: a `@router.get` handler that COMMITs is surprising
- Concurrency: two simultaneous first-load GETs from the same user can both see `plan is None` and both INSERT (UNIQUE constraint mostly mitigates, but still wasteful)

## Scope

Move the default `UserPlan` creation to a **write path**:

- Candidate triggers (PM judgment which is best fit):
  - Signup gate completion (`signup_gate_service` post-completion)
  - First memory creation
  - Admin "create user" path
  - Lazy-create at workspace creation time (paired with workspace setup)

- `get_current_usage` becomes pure-read: SELECT, fall back to settings defaults if no plan row exists, return without COMMIT.

## Out of scope

- Plan migration / backfill (no production rows are missing — verify before scoping)
- Plan tier promotion logic (separate concern)

## Acceptance

- [ ] `get_current_usage` no longer calls `db.commit()` in any code path
- [ ] Default `UserPlan` is created at a deterministic write trigger (NOT on read)
- [ ] Existing users without a `UserPlan` row are reconciled (decide: backfill migration vs lazy fallback to env defaults at read time)
- [ ] Test pin: GET `/api/v1/usage/current` invocations do not COMMIT (mirror #570's `mock_db.commit.assert_not_awaited()` pattern)

## Related

- #570 (PR for service-layer twin: `EffectiveQuotaService`)
- /dx-lead horizontal scan finding (gate1 of #570)

## Discussed

PR for #570 commit message + gate2 CTO review.

Issue #570, opened 2026-05-07
refactor(api): EffectiveQuotaService API simplification — replace lazy self-heal with explicit invalidation


JFK:
## Background

`EffectiveQuotaService.get_effective_quotas()` has a "self-heal" trigger (`backend/src/services/effective_quota_service.py:74-91`) that calls `AddonCalculatorService.recalculate_workspace_bonuses()` (which **COMMITs**) when all of a hardcoded set of addon bonus columns are 0. The heuristic does not include the newer addon columns:

- `addon_storage_bonus_mb` (#485, May 2026)
- `addon_sleep_contexts_bonus` (#560, May 2026)
- Any addon added after 2026-04 will have the same risk

This was flagged 5 times during PR #568's Copilot review (loops 1, 3, 8, 10, 11) but each time deferred as "pre-existing pattern, faithfully mirror, refactor in a separate PR". This is that PR.

## Problem

1. **Stale bonuses**: A workspace whose `addon_*_bonus` columns are non-zero among the original set, but later acquires an `extra_storage` or `extra_sleep_contexts` addon, will keep `addon_storage_bonus_mb=0` / `addon_sleep_contexts_bonus=0` until something else triggers recalc. The lazy self-heal won't fire because the OTHER bonuses are already non-zero.

2. **GET-time COMMIT**: `EffectiveQuotaService.get_effective_quotas()` is called from read-only endpoints (`/api/v1/usage/current`, `/api/v1/workspace/usage/current`, dashboard). When the self-heal fires, it COMMITs from a GET request — surprising for a service named "get". PR #568 had to add `effective_quotas` pass-through plumbing (`_build_*_usage` helpers) to avoid double-COMMIT when both helpers ran back-to-back.

3. **CI lint candidate (related drift)**: The same drift risk exists between the migration's CHECK constraint and the SQLAlchemy ORM `WorkspaceAddon.__table_args__` — caught in PR #568 loop 1 when tests using `Base.metadata.create_all` (instead of running migrations) silently rejected `extra_sleep_contexts`.

## Proposal

Replace the lazy self-heal with explicit invalidation:

- `AddonCalculatorService.recalculate_workspace_bonuses(workspace_id)` should be called explicitly from the **write path** (Stripe webhook handler when `WorkspaceAddon` rows are created/updated/deactivated, admin scripts, manual addon management), not lazily on read.
[...]

PR #589, opened 2026-05-10
refactor(api): make get_current_usage pure-read (#586)

JFK:
## Summary

Sister bug of #570 (PR #588) at the route layer. `GET /api/v1/usage/current` lazy-created a `UserPlan` row and `await db.commit()`-ed from the read path. This PR moves that write to `RoleManager._ensure_user_postgres` (same transaction as the `User` INSERT) so the GET handler stays pure-read.

- New `UserPlan.default_for_user(user_id, settings)` classmethod — shared by the write trigger and the route-layer in-memory fallback so the construction can't drift on a future column add.
- The route's missing-plan branch now builds an in-memory `UserPlan` with no `db.add` and no `db.commit`. Covers (a) the race window between user creation and the first `/usage/current` call and (b) test fixtures that bypass the signup flow. Production traffic should normally not hit this branch.
- Default `UserPlan` is created at user creation in the same transaction as the `User` INSERT — atomic, no concurrency reshuffling.

## Acceptance (#586)

- [x] `get_current_usage` no longer calls `db.commit()` in any code path.
- [x] Default `UserPlan` is created at a deterministic write trigger (NOT on read).
- [x] Existing users without a `UserPlan` row are reconciled via the lazy fallback to settings defaults at read time. Backfill remains explicitly out-of-scope (verified there is exactly one `UserPlan` reader in the codebase: `usage.py:410`).
- [x] Test pin: GET `/api/v1/usage/current` invocations do not COMMIT — extended to a **four-pillar pattern** to close the gap `/dx-lead` horizontal scan flagged.

## The four-pillar test pin

`/dx-lead` horizontal scan revealed that `get_db()`'s unit-of-work pattern in `db/base.py:142` runs `await session.commit()` on normal exit. So removing only the explicit `commit()` from a GET handler leaves the bug latent: a stray `db.add()` would still persist via the dependency-cleanup auto-commit. PR #588's three-pillar pattern (`commit.assert_not_awaited` + `execute.await_count` + spot-check value) is therefore extended with a fourth pillar:

- `db.commit.assert_not_awaited()` — no explicit commit.
- `db.add.assert_not_called()` — defends against `get_db()` auto-commit silent persist.
- `db.execute.await_count >= 10` — read-side queries still happen.
- Functional value check — response uses `settings.default_*` when row is missing.

## Notes from self-review

- **Orphan `UserPlan` edge case**: an existing `UserPlan` row without a `User` row (e.g. from a partial `account_erasure` failure) would surface as 500 on the next first-login because the `_is_email_unique_violation` filter doesn't recognize a `user_plans.user_id` PK violation. This is a column-set widening of the existing IntegrityError fallback class, not a regression — pre-#586 the same orphan would have surfaced via a different path. Worth investigating if it ever fires; not worth silencing here.
- **Factory `settings` parameter is untyped** — pyright won't catch attribute typos at the call site. Acceptable given the factory is exercised by 100% of tests; could be tightened to `Settings` or a Protocol in a future cleanup.
- **`/cto` YAGNI ruling**: the deeper architectural question — "should `get_db()` itself drop auto-commit or split into read/write deps?" — is rejected for blast radius (40+ route files). A CI lint that statically rejects `db.add` / `db.commit` in `@router.get` bodies is a cheaper preventive measure and a candidate for v0.16.x.

## Test plan

- [x] `make lint` (ruff) — clean
- [x] `make type-check` (pyright) — 0 errors
- [x] `pytest tests/api/test_usage_pure_read.py tests/auth/test_roles_user_plan.py -v` — 6/6 pass
- [x] `pytest tests/auth/test_roles.py tests/services/test_effective_quota_service.py tests/api/test_workspace_quota_display.py -v` — 19/19 pass (no regression in adjacent areas)
- [x] Broader sweep `pytest tests/api/ tests/auth/ -k "usage or role or auth or workspace_quota"` — 230 passed, 0 failed (18 pre-existing local-env ERRORS on integration fixtures unrelated to this change)
- [ ] Manual smoke after merge: hit `GET /api/v1/usage/current` for a fresh user, confirm response shape unchanged and no new write to `user_plans` from the GET path

### DIFF
--- a/backend/src/api/routes/usage.py
+++ b/backend/src/api/routes/usage.py
@@ -402,21 +402,16 @@ async def get_current_usage(
 
         settings = get_settings()
 
-        # Get user plan (or default to free)
+        # The default UserPlan row is created at user-creation time in
+        # auth.roles.RoleManager._ensure_user_postgres (#586). The fallback
+        # below covers (a) the race window between user creation and the
+        # first /usage/current call and (b) test fixtures that bypass the
+        # signup flow.
         result = await db.execute(select(UserPlan).where(UserPlan.user_id == user_id))
         plan = result.scalar_one_or_none()
 
         if not plan:
-            # Create default free plan from environment variables
-            plan = UserPlan(
-                user_id=user_id,
-                plan_name="free",
-                memory_limit=settings.default_plan_memory_limit,
-                daily_api_limit=settings.default_plan_daily_api_limit,
-                weekly_api_limit=settings.default_plan_weekly_api_limit,
-            )
-            db.add(plan)
-            await db.commit()
+            plan = UserPlan.default_for_user(user_id, settings)
 
         # Get current memory count.
         # Issue #198 (Bug D): exclude soft-deleted rows so this matches the
--- a/backend/src/auth/roles.py
+++ b/backend/src/auth/roles.py
@@ -326,9 +326,23 @@ class RoleManager:
                 last_login_at=utcnow(),
                 auth_provider=auth_provider,
             )
+            # Pre-check for an orphan UserPlan row before adding a fresh
+            # one in the same tx. The orphan case (UserPlan exists for
+            # this user_id without its User row, e.g. from a prior
+            # partial account_erasure failure) is rare but otherwise
+            # surfaces as a 500 that permanently blocks first-login,
+            # because the IntegrityError handler below only recognizes
+            # users.user_id races and users.email collisions — a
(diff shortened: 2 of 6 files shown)


JFK:
Loop 1: addressed Copilot's orphan-UserPlan finding via pre-check before db.add(new_user) + new test `test_orphan_user_plan_row_skipped_to_avoid_pk_violation`. The IntegrityError handler is unchanged (the orphan case is now sidestepped rather than recovered).

Issue #600, opened 2026-05-11
chore(models): post-#596 DDL drift fix + cosmetic consistency cleanup


JFK:
Follow-up to #596 / epic #370 (SQLAlchemy 2.0 `Mapped[T]` migration).

Both items below were surfaced by the post-#596 quality review but are **pre-existing** to that PR and were not introduced by the Part 1 migration. The migration preserved every Column() form verbatim to keep behavior identical; this issue tracks the polish that was deliberately deferred.

## Part A — DDL drift on `sleep.py` (HIGH)

`sleep.py:SleepReport.status` carries only Python-side `default="running"`, but the alembic migration `a91_add_sleep_reports.py` added `server_default="running"` to the DB column. The model file has never carried the `server_default=` — this drift predates #596.

The same pattern applies to `SleepReportLLMUsage` integer counter columns (`input_tokens`, `output_tokens`, `cached_input_tokens`, `cache_write_tokens`, `calls`) — alembic migrations c02/e01 set `server_default=text("0")` but the model file has only `default=0`.

Symptoms of this drift:

- `Base.metadata.create_all()` (used by integration test fixtures and fresh dev DB bootstrap) produces a schema **without** the DDL `DEFAULT` clauses, since SQLAlchemy generates DDL from the model file, not from alembic migrations.
- Raw INSERTs that bypass the ORM (bulk loaders, raw SQL test data, future migrations) on a `create_all`-built schema would fail NOT NULL on `status`/counter columns instead of getting the documented default value.
- Production DB is correct (alembic-driven), so live writes through the ORM are unaffected; the drift only bites on fresh-fixture paths.

For contrast, `MemoryAnalysis` and `MemoryAnalysisCluster` in `analysis.py` (added in d06_494) correctly carry **both** `default=...` and `server_default=...` on `status` / `paid_by` / `id`, per the dual-default convention the codebase adopted in c02/d05.

### Acceptance

- [ ] Add `server_default="running"` to `SleepReport.status` (matching migration a91)
- [ ] Add `server_default=text("0")` to `SleepReportLLMUsage.input_tokens / output_tokens / cached_input_tokens / cache_write_tokens / calls`
- [ ] Audit other `sleep.py` model columns against their migration `server_default=` settings — fix any other Python-only vs DB-default drift
[...]

Issue #599, opened 2026-05-11
chore(types): re-enable 5 SA-related + 4 broader pyright rules (follow-up to #596)


JFK:
Follow-up to #596 (PR-C of SQLAlchemy 2.0 `Mapped[T]` migration epic #370).

## Context

The 9 `report* = false` suppressions in `backend/pyproject.toml:181-225` were originally added to silence pre-migration SQLAlchemy `Column[T]` vs `T` mismatches. After #596 (PR-C) merges, every model file uses `Mapped[T] = mapped_column(...)` and those mismatches are gone.

A per-rule audit (run during #596, results recorded in `pyproject.toml` comment) confirmed:

| Rule | Residual errors after re-enable | Notes |
|---|---|---|
| `reportArgumentType` | 163 | caller-side, many call-sites |
| `reportAttributeAccessIssue` | 36 | caller-side, many call-sites |
| `reportAssignmentType` | 4 | `api/routes/contexts.py:308,371`, `services/workspace_service.py:612,656` |
| `reportReturnType` | 5 | `auth/roles.py:576`, `auth/oauth2.py:369`, `db/qdrant.py:1195`, `services/workspace_service.py:612,656` (overlap) |
| `reportCallIssue` | 22 | caller-side |

None of the residuals are model-layer issues — they are call-site type narrowing (e.g. `WorkspaceMember | None` vs `WorkspaceMember`, missing return path, Google OAuth credentials union not narrowed).

## Goal

Narrow the caller-side issues and re-enable each of the 5 SQLAlchemy-related pyright rules. The original Column[T] suppression rationale is no longer load-bearing — only the caller-side fix backlog keeps the rules disabled.

### Tier A (easy wins, < 20 residuals each)

Re-enable these two first; residuals are small and tractable:

- [ ] `reportAssignmentType` — fix 4 caller-side errors:
  - `api/routes/contexts.py:308` — `dict[str, str | None]` → `dict[str, str]` narrowing
  - `api/routes/contexts.py:371` — `dict[Any, tuple[...]]` → declared dict type
  - `services/workspace_service.py:612, 656` — `WorkspaceMember | None` → `WorkspaceMember` (raise on None instead of return)
- [ ] `reportReturnType` — fix 5 caller-side errors:
  - `auth/roles.py:576` — missing return path on a list-returning function
  - `auth/oauth2.py:369` — Google `Credentials` union not narrowed
[...]

Issue #370, opened 2026-04-18
refactor(api): migrate SQLAlchemy models to 2.0 Mapped[T] for Pyright type safety


JFK:
## Background

Pyright surfaces type confusion across the entire `backend/src/` tree because all 6 model files in `backend/src/models/` (`auth.py`, `config.py`, `memory.py`, `neural.py`, `resource.py`, `sleep.py`) use the **SQLAlchemy 1.x legacy `Column` declarative pattern** instead of the modern SQLAlchemy 2.0 `Mapped[T]` + `mapped_column()` annotations.

Verified: `Mapped[` and `mapped_column` have **0 occurrences** across the entire `backend/src/models/` directory.

The consequence is that Pyright sees attribute access like `memory.id` as `Column[UUID]` (the class-level descriptor type) instead of `UUID` (the instance value type). This makes every `dict[Memory.id, ...]`, `sorted([memory.id, hit_id], ...)`, or function call passing `instance.field` to a typed parameter trigger a false-positive "type mismatch" warning.

## Symptom inventory (representative — full count is much larger)

Surfaced during PR #306 ([discovery_metrics PR link will be added once merged]) review when Pyright re-checked `backend/src/services/sleep/edge_discovery.py` after edits. **All 5 warnings pre-date PR #306 and exist on `main`** — they were just not previously flagged because Pyright caches per-file:

| File:line | Symptom | Root cause pattern |
|---|---|---|
| `services/sleep/edge_discovery.py:82` | `_is_synthetic_seed_edge` returns `ColumnElement[bool]` instead of `bool` | `edge.edge_type == "..."` returns `ColumnElement` because `edge_type` is `Column[str]` |
| `services/sleep/edge_discovery.py:275` | `dict[Column[UUID], Memory]` instead of `dict[UUID, Memory]` | `{m.id: m for m in sampled}` — `m.id` is seen as `Column[UUID]` |
| `services/sleep/edge_discovery.py:405` | `embed(memory.summary, ...)` — `Column[str]` not `str` | Same pattern, `summary` field |
| `services/sleep/edge_discovery.py:433` | `sorted([memory.id, hit_id], key=str)` produces `tuple[Column[UUID] \| UUID, ...]` | Mixed concrete + Column type in list literal |
| `services/sleep/edge_discovery.py:434` | `candidates.append((memory.id, hit_id, score))` | Same pattern |

[...]

PR #601, opened 2026-05-11
refactor(api,ci): remaining models + CI guard (Mapped[T] PR-C, #370)

### LINKED ISSUE #596: refactor(api,ci): remaining models + CI guard (Mapped[T] PR-C of 3, #370)
Part of #370 (SQLAlchemy 2.0 `Mapped[T]` migration epic) — **PR-C of 3, final**.

**Depends on**: #594 (Base + pattern) and #595 (auth.py).

### Part 1: migrate remaining 6 model files

- `backend/src/models/memory.py` (67 `Column(`)
- `backend/src/models/resource.py` (57 `Column(`)
- `backend/src/models/sleep.py` (46 `Column(`)
- `backend/src/models/analysis.py` (35 `Column(`)
- `backend/src/models/neural.py` (23 `Column(`)
- `backend/src/models/signup_gate.py` (22 `Column(`)

Total: ~250 `Column(` call sites. None have `relationship()` calls; pattern is identical to #594.

### Part 2: CI regression guard

Add a grep-based guard that fails CI when new `Column(...)` definitions appear in `backend/src/models/`. This prevents drift back to the legacy pattern after the migration completes (analogous to the memory `1f6759d1` finding about manual schema-drift between `addon_calculator_service` and `Workspace` model).

The regex must catch **both** the bare `name = Column(...)` form **and** the type-annotated half-migration form `id: int = Column(...)` (which pyright accepts but SQLAlchemy 2.0 does NOT recognize as a `Mapped` attribute — see memory `fcc7cf87`). The `(:\s*[^=]+)?` group makes the type annotation optional, covering both forms in one rule.

Proposed implementation:

```makefile
# Makefile (repo root) — new target, wired into `lint`
.PHONY: lint-models-no-column
lint-models-no-column:
	@! grep -rnE '^\s+\w+(:\s*[^=]+)?\s*=\s*Column\(' backend/src/models/ \
	  || (echo "ERROR: legacy 'Column(...)' usage detected in backend/src/models/ — use 'Mapped[T] = mapped_column(...)' instead. Annotated half-migration form 'id: int = Column(...)' is also rejected (not recognized as Mapped attribute)."; exit 1)
```

And wire into the repo-root `Makefile`'s `lint` target + CI workflow.

### Part 3: pyright suppression re-enablement

`backend/pyproject.toml:181-190` currently disables 9 pyright rules to silence pre-migration SQLAlchemy `Column[T]` vs `T` mismatches. After the migration, these mismatches are gone, so we attempt to re-enable the 5 most clearly SQLAlchemy-related rules:

- `reportArgumentType`
- `reportAttributeAccessIssue`
- `reportAssignmentType`
- `reportReturnType`
- `reportCallIssue`

The remaining 4 (`reportGeneralTypeIssues`, `reportOptionalMemberAccess`, `reportIndexIssue`, `reportOptionalOperand`) are broader and likely have non-SQLAlchemy noise; they stay disabled with justification comments and are deferred to a follow-up issue.

## Why bundled

Part 1 reuses the #594 pattern mechanically; the 6 files have no relationship() complexity. Part 2 must land **after** all Column() usage is gone, so closing PR-C with the guard prevents the lint from firing on its own merge commit. Part 3 piggybacks on the same migration so the type-check warning baseline can be measured against a stable code state.

### Part 1
- [ ] All 6 target files migrated to `Mapped[T] + mapped_column()`
- [ ] No remaining `Column(` definitions in `backend/src/models/*.py`

### Part 2
- [ ] `make lint-models-no-column` target added to **repo-root `Makefile`** and wired into the existing `lint` target's dependency list
- [ ] CI workflow step added (or `make lint` already covered by CI — verify)
- [ ] Guard regex catches **both** forms:
  - bare:        `name = Column(...)`
  - annotated:   `id: int = Column(...)` (half-migration form, memory `fcc7cf87`)
- [ ] Negative-case smoke test: temporarily insert one synthetic instance of **each** form, confirm CI fails for both, revert before final commit (or capture CI run URLs in PR description)

### Part 3
- [ ] Each of the 5 SQLAlchemy-related pyright rules attempted re-enabled (one-by-one). Rules that produce **< 20 residual warnings** post-migration stay enabled; those above the threshold revert to `false` with a one-line `# justification:` comment in `pyproject.toml`.
- [ ] Remaining 4 broader rules deferred — explicitly noted in PR description with link to a new follow-up issue.

### Global verification
- [ ] Symptom samples in `backend/src/services/sleep/edge_discovery.py:82, 275, 405, 433, 434` — all 5 pyright warnings resolved
- [ ] `make type-check` total warning count compared to pre-#370 baseline (expect significant reduction; document delta in PR description with caller-propagation forecast — see memory `6e3f286d`: auth.py forecast underestimated by 2.4x, so weight caller surface, not just column count)
- [ ] `make test-local` + `make test-integration` pass

## Out of scope

- Service / API / MCP layer code changes — those should auto-resolve when models are correctly typed
- Pydantic / Schema migrations (orthogonal)
- Pyright config tuning beyond the 5 SQLAlchemy-related rules (the 4 broader rules are deferred to a follow-up issue)

## Related

- Parent epic: #370
- Predecessors: #594 (a3e2a971), #595 (d780b548)
- Closes the epic: after this PR merges, #370 can be closed
- Drift-prevention precedent: memory finding on `addon_calculator_service` ↔ `Workspace` model sync (informs why Part 2 matters)
- Half-migration latent bug pattern: memory `fcc7cf87` — informs why Part 2 regex must catch the annotated form

---

_Acceptance refined 2026-05-11 via /gh-issue-driven:start gate1 yellow → operator opted to patch issue body before branch creation. Changes: (1) grep regex extended to catch `id: int = Column(...)` half-migration form, (2) Makefile path corrected to repo-root, (3) Part 3 (pyright re-enablement) added._

JFK:
Closes #596

## Summary

PR-C of 3, **final**, in the SQLAlchemy 2.0 `Mapped[T]` migration epic (#370).
After merge, every model file in `backend/src/models/*.py` uses
`Mapped[T] = mapped_column(...)` — total 548 sites across 16 files, zero
legacy `Column(...)`.

- **Part 1** — migrate the 6 remaining model files (250 `Column` sites):
  memory.py, resource.py, sleep.py, analysis.py, neural.py, signup_gate.py
- **Part 2** — add a CI grep guard against drift back to legacy `Column(...)`,
  including the annotated half-migration form `id: int = Column(...)`
  (memory `fcc7cf87` — pyright-OK but SA-2.0-broken). Wired into the
  repo-root `Makefile`'s `lint:` dependency chain and added as a step in
  both `ci.yml` and `pr-review.yml`. 5 hermetic subprocess smoke tests
  verify the regex catches both forms.
- **Part 3** — audit the 9 pyright suppressions in `backend/pyproject.toml`,
  document residual error counts per rule, and file follow-up #599 for the
  caller-side narrowing fixes that would unblock re-enable.

### Migration recipe (followed PR-B auth.py @ `d780b548`)

- 36 distinct column patterns catalogued before implementation
- `Mapped[T]` annotation matches SQL type Python semantics
  (`Mapped[uuid.UUID | None]` for nullable UUID, `Mapped[list[float]]` for
  `ARRAY(Float)`, `Mapped[dict[str, Any]]` for non-null JSONB, etc.)
- `nullable=True`/`False` explicit on every non-PK column (PR-B convention)
- `default=` vs `server_default=` semantics preserved exactly — never swapped
- `JSON` vs `JSONB` preserved per column (not normalized)
- `ARRAY` element type matches `Mapped[list[X]]`
- Computed columns omit `nullable=` per SA 2.0 rule
- `text("'...'::jsonb")` preserved for JSONB defaults
- `DateTime(timezone=True)` preserved where present
- Column-name aliases preserved as first positional arg
  (`NeuralMemoryEdge.edge_metadata = mapped_column("metadata", JSON, ...)`)
- Composite PK on `MemoryAnalysisAssignment` (two `primary_key=True` columns)
- Self-referential FK on `MemoryAnalysisCluster.parent_id` (string ref)
- `expression.text(...)` preserved verbatim in `neural.py` per
  operator-confirmed "out-of-scope rewrite" rule

### Per-file verification

| File | `Column` (pre) | `mapped_column` (post) | Tests passed |
|---|---|---|---|
| memory.py | 67 | 67 | 159 |
| resource.py | 57 | 57 | 122 |
| sleep.py | 46 | 46 | 221 |
| analysis.py | 35 | 35 | 56 |
| neural.py | 23 | 23 | 147 |
| signup_gate.py | 22 | 22 | 50 |
| **Total** | **250** | **250** | **~755** + 5 new smoke tests |

### Part 2 — CI guard

Makefile target:

```makefile
.PHONY: lint-models-no-column
lint-models-no-column:
	@! grep -rnE '^\s+\w+(:\s*[^=]+)?\s*=\s*Column\(' $(BACKEND_DIR)/src/models/ \
	  || (echo "ERROR: ..."; exit 1)
```

The optional `(:\s*[^=]+)?` group is what makes the annotated half-migration form trip the guard. Smoke tests at `backend/tests/test_models_no_column_guard.py` verify:

- bare `name = Column(...)` → caught
- annotated `id: int = Column(...)` → caught
- PEP 604 union `user_id: str | None = Column(...)` → caught
- migrated `Mapped[T] = mapped_column(...)` → NOT caught
- bare `from sqlalchemy import Column` import → NOT caught

### Part 3 — pyright audit results

Per-rule residual error counts after Part 1 (measured by flipping each rule to `true`, one at a time, vs the all-disabled baseline of 1 error):

| Rule | Residual | Tier in #599 |
|---|---|---|
| `reportArgumentType` | 163 | C (large sweep) |
| `reportAttributeAccessIssue` | 36 | B (medium) |
| `reportAssignmentType` | 4 | **A** (easy: 4 caller-side narrowing) |
| `reportReturnType` | 5 | **A** (easy: incl. `auth/roles.py:576`) |
| `reportCallIssue` | 22 | B |

All 5 SA-related suppressions stay `false` in this PR. Per operator HITL judgment, all residuals are caller-side narrowing (services, routes, auth) — outside the model-layer scope of #596. `#599` tracks the tiered re-enable backlog.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: yellow via /ask (CTO lens) — resolved in-flight via issue body patch (Part 2 regex extension, Makefile path correction, Part 3 acceptance added)
- review provider: copilot

3-lens green consensus with shared next-step pointer: prioritise #599 Tier A (`reportAssignmentType` + `reportReturnType`, 9 caller-side narrowing fixes including `auth/roles.py:576` which is security-relevant).

- `~/.claude/cache/gh-issue-driven/596-feat-refactor-api-ci-remaining-models-ci-guar.gate1.md`
- `~/.claude/cache/gh-issue-driven/596-feat-refactor-api-ci-remaining-models-ci-guar.gate2.md`

## Follow-up issues

- **#599** — re-enable the 5 SA-related pyright rules (Tier A → C) and the 4 broader rules (Tier D), one rule at a time, each with the caller-side narrowing fixes the rule surfaces.
- **#600** — fix the pre-existing model-vs-migration DDL drift on `sleep.py:SleepReport.status` (and `SleepReportLLMUsage` counters) and clean up 3 cosmetic consistency nits (kwarg order, `uuid4` import shape, `ARRAY` import source) that were preserved verbatim by this migration but are inconsistent with the PR-A/PR-B reference style.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,13 @@ jobs:
       - name: Format check
         run: cd backend && ruff format --check src/ tests/
 
+      # #596 / epic #370 — guard against drift back to legacy SQLAlchemy 1.x
+      # Column() in model files. Catches both `name = Column(...)` and the
+      # annotated half-migration form `id: int = Column(...)` (pyright-OK,
+      # SA-2.0-broken). The Makefile target lives at repo root.
+      - name: Models no-Column guard
+        run: make lint-models-no-column
+
   backend-unit:
     runs-on: ubuntu-latest
     services:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,6 @@ jobs:
       - name: Format check
         run: cd backend && ruff format --check src/ tests/
 
-      # #596 / epic #370 — guard against drift back to legacy SQLAlchemy 1.x
-      # Column() in model files. Catches both `name = Column(...)` and the
-      # annotated half-migration form `id: int = Column(...)` (pyright-OK,
-      # SA-2.0-broken). The Makefile target lives at repo root.
       - name: Models no-Column guard
         run: make lint-models-no-column
 
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -219,6 +219,12 @@ jobs:
           ruff format --check src/
         continue-on-error: true  # Allow existing formatting issues
 
+      # 8b. Models no-Column guard (#596 / epic #370)
+      - name: Models no-Column guard
+        if: steps.check.outputs.authorized == 'true'
+        id: no_column_guard
+        run: make lint-models-no-column
+
       # 9. Type check
       - name: Type Check (pyright)
         if: steps.check.outputs.authorized == 'true'
--- a/Makefile
+++ b/Makefile
@@ -211,8 +211,20 @@ lint: lint-models-no-column
 # Negative-case smoke test lives at backend/tests/test_models_no_column_guard.py.
 .PHONY: lint-models-no-column
 lint-models-no-column:
-	@! grep -rnE '^[[:space:]]+[[:alnum:]_]+(:[[:space:]]*[^=]+)?[[:space:]]*=[[:space:]]*Column\(' $(BACKEND_DIR)/src/models/ \
-	  || (echo "ERROR: legacy 'Column(...)' usage detected in $(BACKEND_DIR)/src/models/. Use 'Mapped[T] = mapped_column(...)' instead. The annotated half-migration form 'id: int = Column(...)' is also rejected because SQLAlchemy 2.0 does not recognize it as a Mapped attribute (it silently fails type-resolution)."; exit 1)
+	@if grep -rnE '^[[:space:]]+[[:alnum:]_]+(:[[:space:]]*[^=]+)?[[:space:]]*=[[:space:]]*Column\(' $(BACKEND_DIR)/src/models/; then \
+	  echo ""; \
+	  echo "ERROR: legacy 'Column(...)' usage detected in $(BACKEND_DIR)/src/models/."; \
+	  echo "Use 'Mapped[T] = mapped_column(...)' instead. The annotated half-migration"; \
+	  echo "form 'id: int = Column(...)' is also rejected because SQLAlchemy 2.0 does"; \
+	  echo "not recognize it as a Mapped attribute (it silently fails type-resolution)."; \
+	  exit 1; \
+	else \
+	  EC=$$?; \
+	  if [ "$$EC" -ne 1 ]; then \
+	    echo "ERROR: grep failed with exit $$EC scanning $(BACKEND_DIR)/src/models/ (likely missing directory, permission issue, or grep not installed)." >&2; \
+	    exit 1; \
+	  fi; \
+	fi
 
 .PHONY: format
 format:
--- a/Makefile
+++ b/Makefile
@@ -203,10 +203,15 @@ lint: lint-models-no-column
 #   - annotated:   `id: int = Column(...)`  ← pyright accepts but SA 2.0 does
 #                                              NOT recognize as a Mapped attr
 #
+# POSIX character classes ([[:space:]] / [[:alnum:]_]) are used instead of
+# the GNU \s / \w shorthand so the guard is portable across GNU grep
+# (Linux CI) and BSD grep (macOS dev). The shorthand would be treated as
+# literal `s` / `w` on BSD and silently let regressions through.
+#
 # Negative-case smoke test lives at backend/tests/test_models_no_column_guard.py.
 .PHONY: lint-models-no-column
 lint-models-no-column:
-	@! grep -rnE '^\s+\w+(:\s*[^=]+)?\s*=\s*Column\(' $(BACKEND_DIR)/src/models/ \
+	@! grep -rnE '^[[:space:]]+[[:alnum:]_]+(:[[:space:]]*[^=]+)?[[:space:]]*=[[:space:]]*Column\(' $(BACKEND_DIR)/src/models/ \
 	  || (echo "ERROR: legacy 'Column(...)' usage detected in $(BACKEND_DIR)/src/models/. Use 'Mapped[T] = mapped_column(...)' instead. The annotated half-migration form 'id: int = Column(...)' is also rejected because SQLAlchemy 2.0 does not recognize it as a Mapped attribute (it silently fails type-resolution)."; exit 1)
 
 .PHONY: format
--- a/Makefile
+++ b/Makefile
@@ -190,10 +190,24 @@ test-watch:
 	cd $(BACKEND_DIR) && pytest-watch
 
 .PHONY: lint
-lint:
+lint: lint-models-no-column
 	@echo "Running linter..."
 	cd $(BACKEND_DIR) && ruff check src/ tests/
 	@echo "Lint complete."
+
+# Guard against drift back to the legacy SQLAlchemy 1.x Column() pattern in
+# model files (#596 / epic #370). The regex accepts both forms the half-
+# migration trap produces:
+#
+#   - bare:        `name = Column(...)`
+#   - annotated:   `id: int = Column(...)`  ← pyright accepts but SA 2.0 does
+#                                              NOT recognize as a Mapped attr
+#
+# Negative-case smoke test lives at backend/tests/test_models_no_column_guard.py.
+.PHONY: lint-models-no-column
+lint-models-no-column:
+	@! grep -rnE '^\s+\w+(:\s*[^=]+)?\s*=\s*Column\(' $(BACKEND_DIR)/src/models/ \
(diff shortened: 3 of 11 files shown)


JFK:
Loop 1: Addressed 1/1 actionable Copilot finding (POSIX character classes for cross-grep portability). Smoke tests + live sanity check confirm regex still catches both `name = Column(...)` and `annotated: int = Column(...)` forms on GNU grep, and the change makes it portable to BSD grep (macOS) as well.

JFK:
Loop 2 status: addressed 4/5 actionable items inline (commit pending push), replied to 2.

**Applied inline (4 fixes):**
- `resource.py:165` `event_metadata` — `default={}` → `default=dict` (callable, fresh per row)
- `resource.py:286` `metrics` — same fix
- `test_models_no_column_guard.py` module docstring — reworded from "Negative-case smoke test" to "Smoke tests" (covers both positive and negative)
- `test_models_no_column_guard.py:44` — added `encoding="utf-8"` to `Path.write_text` for CI determinism

**Replied (not changed):**

1. **Makefile grep exit-code-2 handling**: thanks for flagging, but the diagnosis is inverted. `! grep -rnE ...` treats exit 2 (grep error) as success because `!` inverts any non-zero exit, so the rule actually *silently passes* on grep failures rather than emitting the legacy-Column message. The case for explicit exit-code branching (`case "$EC" in 0|1|*) ...`) is real but minor — deferred as out-of-scope polish; will file a follow-up if it bites in operations.

2. **GUARD_REGEX duplication between Makefile and Python test**: this duplication is intentional. The smoke test's job is to verify the exact regex shipped in the Makefile catches the contract; pulling the regex from a single source (e.g. extracting from Makefile via parse) would remove the drift-detection property — if either half goes stale, both fail loud right now. The constant's comment block documents this trade-off. Keeping as-is.

**Stale (already fixed in 8e04941e):**

- POSIX `\s`/`\w` → `[[:space:]]`/`[[:alnum:]_]` was Copilot's loop-1 finding; the comment re-surfaced in loop 2 review summary but the code is already fixed on HEAD.

JFK:
Loop 3: addressed 1/1 actionable finding (8fa9327e). Restructured to explicit if/else exit-code branching: rc=0 → fail with legacy-Column message; rc=1 → pass; rc≥2 → fail with distinct `grep failed` message. Verified all 3 paths. Multi-line echo for CI readability also folded in.

JFK:
Loop 4 reply (no code change — clarifying the Makefile comment that Copilot flagged):

The comment is accurate as written. Copilot's claim "`\s`/`\w` aren't special on either GNU or BSD with `grep -E`" is incorrect for GNU grep:

- **POSIX ERE strict** does NOT include `\s`/`\w` — agreed.
- **GNU grep `-E`** recognizes `\s`/`\w` as **non-POSIX extensions** (documented in GNU grep manpage). This is the reason the original `^\s+\w+\s*=\s*Column\(` regex passed the 5 smoke tests on Linux CI before commit 8e04941e flipped to POSIX classes — if GNU grep treated `\s` as literal, the smoke tests would have all failed (`^s+w+s*=s*Column(` doesn't match `id = Column(...)`).
- **BSD grep `-E`** does NOT recognize them (POSIX-strict ERE).
- **PCRE / `grep -P`** also recognizes them (and many more) — but `-P` is GNU-only and not available on BSD grep, so it's not a portable solution either.

So the Makefile comment's framing — "POSIX character classes used instead of the GNU \s / \w shorthand" — is correct: they ARE a GNU shorthand (and a PCRE shorthand), absent from POSIX strict, absent from BSD. The fix in 8e04941e remains the right portability choice.

No code change for this loop. Ready for merge from my side.

PR #615, opened 2026-05-12
fix(auth,db): narrow DCR_DEFAULT_SCOPE to exclude memory:admin (#608)

### LINKED ISSUE #608: feat(auth): enforce memory:admin + memory:delete + drop OIDC claim (gate for #592)
## Context (updated 2026-05-12)

#592 (v0.15.4 hotfix) consolidated the OAuth metadata + DCR scope set
behind `auth.mcp_scopes.ALL_ADVERTISED_SCOPES`. To avoid a backward-
incompatible scope removal in the hotfix, the canonical set is the UNION
of every scope each source previously advertised:

```
openid memory:read memory:write memory:admin memory:delete offline_access
```

The drift is fixed (every source matches), but three of these are
currently advertised without backing semantics:

1. **`memory:admin`** — no route enforces it. A DCR client that holds
   `memory:admin` believes it has been granted admin privileges by the
   authorization server, which it has not.
2. **`memory:delete`** — same situation. Was previously on
   `/oauth-protected-resource` only; kept on all sources now for compat.
3. **`openid`** — this server does not issue `id_token` (no OIDC subject
   claims path). The OIDC discovery endpoint advertises `openid` for
   compat with clients that pre-flight check the scope, but proceeding
   past discovery to the authorization step does not produce an
   `id_token`.

## This issue: paired enforcement / honest-deprecation gate

Pick one of the following per scope and ship it. Each can be its own
sub-PR.

### `memory:admin`
- Decide semantics (own-context admin, workspace admin, etc.).
- Add `requires_scope("memory:admin")` decorator/dependency.
- Apply to the routes that should require admin scope (audit log,
  `delete_context` via MCP, workspace member mutation via MCP tools).
- Decide DCR auto-grant policy: keep including `memory:admin` in
  `DCR_DEFAULT_SCOPE` (every DCR client gets admin), or filter it out
  of DCR responses unless explicitly requested by the client AND a
  workspace owner has approved the elevation.

### `memory:delete`
- Decide whether it is a real scope distinct from `memory:admin` or
  whether the latter implies the former.
- If distinct: add enforcement to the `delete_memory` / context-delete
  paths.
- If implied by admin: drop from advertisement once enough time has
  passed for clients to migrate.

### `openid`
- Path A: implement actual OIDC (issue `id_token`, add `userinfo`
  endpoint, add JWKS).
- Path B: drop `openid` from advertisement once enough time has passed
  for clients to migrate.

## Required before #592 can be considered fully resolved

This issue is the **enforcement gate** for #592. Filing it is the gate;
landing the actual work can happen on its own timeline. The hotfix on
#592 is shippable without it.

## Related

- #592 fix(oauth): scope mismatch across 3 metadata sources (this issue's parent)
- #513 fix(oauth): support loopback redirects in DCR + RFC 6749 error format
- #519 fix(auth): DB constraint misalignment — oauth_clients.owner_id NOT NULL

---

## Gate1 design clarifications (added 2026-05-12, CSO review)

CSO gate1 review on `/gh-issue-driven:start 608` surfaced 5 design
decisions that must be settled before implementation begins. These
supersede / clarify the per-scope sections above.

### D1. Narrowing-first ordering (privilege-escalation prevention)

Land in this exact order:

1. Drop `memory:admin` from `DCR_DEFAULT_SCOPE` in
   `backend/src/auth/mcp_scopes.py` + DB migration to backfill
   existing `oauth_clients.scope` rows (downgrade rows that received
   `memory:admin` via the union default).
2. Introduce the `require_scope` FastAPI dependency (see D2).
3. Apply `require_scope("memory:admin")` to the workspace-admin
   routes identified in D3.

**Rationale**: if step 3 lands before step 1, every already-issued DCR
token automatically gains admin against the newly-protected routes — a
backward-compat privilege escalation. The narrowing PR is the
gate-keeper, not the route decorator.

### D2. `require_scope` is new infrastructure, not existing utility

`rg "requires_scope|require_scope" backend/src` returns zero matches —
there is no scope-based authorization layer today (only role/permission
services). Land as:

- Location: `backend/src/auth/dependencies.py:require_scope(scope: str)`
- Return: FastAPI `Depends(...)` callable
- Rejection: RFC 6750 §3.1 compliant — `403 Forbidden` +
  `WWW-Authenticate: Bearer error="insufficient_scope" scope="<scope>"`
- Interaction with non-OAuth auth modes:
  - **API keys**: no scope → `require_scope` does NOT apply; API key
    routes use existing role/permission RBAC
  - **Resource tokens**: capability-typed, not scope-typed → no overlap
  - **OAuth Bearer (DCR + IdP)**: scope check applies

### D3. `memory:admin` semantic scope = workspace admin

Not own-context admin (covered by normal RBAC), not system admin
(`/api/v1/system/*` stays on the SystemAdminService lane and never
accepts `memory:admin` as sufficient).

Protected routes (initial set):
- audit_log read endpoints (workspace-scoped)
- workspace member mutation (add / remove / role-change)
- `delete_context` via MCP when the context has ≥2 members

### D4. `memory:delete` is distinct (least-privilege)

Not implied by `memory:admin`. Enforce on:
- `delete_memory` (MCP)
- `delete_context` (REST + MCP single-member case)

Rationale: a delete-only worker (backup pruner, retention job) should
not need admin. Coarsening the permission grain reverses the
"fictional-scope debt that compounds" pattern this issue is closing.

### D5. `openid` = Path B (drop, no deprecation period)

- Remove `"openid"` from `ALL_ADVERTISED_SCOPES` in `mcp_scopes.py`
- Set `id_token_signing_alg_values_supported: []` in
  `/.well-known/openid-configuration`
- Update the existing drift-guard test
  (`test(api): pin OAuth metadata scopes_supported drift guard (#592)`)
  for the new expected set

No deprecation cycle: the server never issued `id_token`, so a strict
OIDC client cannot already be relying on the discovery advertisement.
Discovery-only `openid` consumers were never functional in steady-state.

### Test coverage required

- `test(auth)`: `require_scope` rejects token without scope →
  403 + `insufficient_scope` `WWW-Authenticate`
- `test(auth)`: DCR registration response excludes `memory:admin`
  from default scope after narrowing
- `test(db)`: backfill migration downgrades existing
  `oauth_clients.scope` rows (removes `memory:admin` from clients
  that received it via the union default)
- `test(api)`: drift-guard test green after `openid` removal
  (narrowing = expected change, drift = different problem)

### Scope-cut policy for v0.16.0

Land in order. Items 1+2+3 (`memory:admin` narrowing + `require_scope`
infrastructure + workspace-admin route gates) are the v0.16.0
commitment that honors the #592 forward-compat anchor. Items 4
(`memory:delete` distinct enforcement) and 5 (`openid` drop) may
defer to v0.16.1 if PR count compresses the release. Splitting items
1–3 weakens the "advertise-without-enforcement" closing and is not
permitted.

JFK:
Closes #608

## Summary
- Narrows the DCR fallback default scope to exclude `memory:admin`, closing the silent auto-grant of admin to newly-registered DCR clients (the "advertise-without-enforcement" gap left open by #592).
- `ALL_ADVERTISED_SCOPES` is unchanged — `memory:admin` remains discovery-advertised and explicitly requestable via the DCR `scope` parameter. Only the silent auto-grant in the fallback default is removed.
- Alembic e09 backfills `oauth_clients.scope` on the exact pre-narrow canonical string only; rows with custom scope sets (explicit grants, custom subsets) are preserved (R1 policy from gate1 review).
- This is sub-PR D1 of #608's enforcement gate. Subsequent sub-PRs land `require_scope` infrastructure (D2) and apply it to workspace-admin routes (D3); `memory:admin` enforcement is NOT in this PR.

## Implementation notes
- `backend/src/auth/mcp_scopes.py`: new `DCR_DEFAULT_SCOPES: tuple[str, ...]` derived as `ALL_ADVERTISED_SCOPES` minus `memory:admin`; `DCR_DEFAULT_SCOPE` is space-joined from it. Module docstring expanded with SEP-835 alignment and rollback note.
- `backend/alembic/versions/e09_608_dcr_default_narrow.py`: new migration revising e08. Uses `op.execute(sa.text(...).bindparams(...))` with hardcoded pre/post canonical strings (migration-snapshot discipline, matches e08 convention). Operator rollback runbook embedded in docstring (3 steps if Claude Code MCP DCR breaks post-deploy).
- `backend/tests/api/test_oauth_metadata_drift.py`: 3 new tests pinning the narrowing contract — DCR fallback excludes admin, explicit `scope=memory:admin` request preserves admin, server-side `get_allowed_scope` returns `requested ∩ registered` (SEP-835 least-privilege).
- `backend/tests/test_e09_608_dcr_default_narrow_migration.py`: new file with 5 string-level pin tests linking migration constants to e08's canonical string and runtime `DCR_DEFAULT_SCOPE`. Pins "only memory:admin removed" to prevent silent scope leakage into the D4/D5 PRs.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: green via /ask (CSO lens, re-run after body refinement)
- review provider: copilot

- ~/.claude/cache/gh-issue-driven/608-feat-feat-auth-enforce-memory-admin-memory-de.gate1.md
- ~/.claude/cache/gh-issue-driven/608-feat-feat-auth-enforce-memory-admin-memory-de.gate2.md

Operator note (from gate1 review):
- Deploy app code BEFORE running this migration (or atomic deploy). A DCR registration that lands in the window between migration and deploy uses the old app's union default and is not touched by `upgrade()`.

### DIFF
--- a/backend/alembic/versions/e09_608_dcr_default_narrow.py
+++ b/backend/alembic/versions/e09_608_dcr_default_narrow.py
@@ -13,28 +13,47 @@ FIRST, then enforce on routes.
 
 Scope of the data fix (R1 policy from gate1 review):
 
-- Rows whose ``scope`` is exactly the post-#592 canonical string get
-  ``memory:admin`` stripped. This is the "got admin by default" cohort
-  — clients that did not explicitly opt into admin and would have
-  received it merely because the DCR fallback included it.
+- Only DCR-registered clients are touched. The WHERE clause includes
+  ``owner_id IS NULL`` — DCR clients have ``owner_id=NULL`` per
+  ``d04_519_oauth_owner_nullable``; admin-managed clients have a
+  non-null ``owner_id`` and are never touched by this migration even
+  if their scope string happens to match the canonical.
+- Among DCR clients, rows whose ``scope`` is exactly the post-#592
+  canonical string get ``memory:admin`` stripped. This is the "got
+  admin by default" cohort — clients that did not explicitly opt
+  into admin and would have received it merely because the DCR
+  fallback included it.
 - Rows whose ``scope`` differs from the exact post-#592 canonical
-  string are left untouched. This preserves any client that
-  EXPLICITLY requested ``memory:admin`` (their stored scope is a
-  custom subset / superset that does not match the full default), as
-  well as any other custom-scoped client.
+  string are left untouched. This preserves any DCR client that
+  registered with a custom scope set (subset, superset, or any
+  ordering that doesn't match the canonical exactly).
 
-This guarantees that no client that explicitly requested admin loses
-it. Clients that got admin by accident lose it, which is the intended
-behaviour — they never needed admin to begin with.
+Residual limitation (Copilot review on PR #615): the migration cannot
+distinguish a DCR client that EXPLICITLY requested the full canonical
+string (``scope="openid memory:read memory:write memory:admin
+memory:delete offline_access"`` in this exact ordering) from a client
+that received the same string via the fallback default. Such a client
+would have ``memory:admin`` stripped despite having explicitly
+requested it. In practice this collision is rare — it requires the
+client to type out all six scopes in the server's canonical ordering,
+which only matches clients that derived their scope list from the
+server's discovery endpoint. Affected clients can recover by
+re-registering with a distinguishing custom scope ordering, or by
+operator-driven re-grant.
 
 ``memory:delete`` is intentionally NOT narrowed in this migration. The
 distinct-vs-implied policy for ``memory:delete`` lands in #608 (D4) in
 a separate migration; coarsening admin first, then handling delete as
 its own least-privilege decision keeps the two concerns independent.
 
-Downgrade restores ``memory:admin`` on the exact rows we touched. We
-can identify them precisely because the pre-narrow and post-narrow
-canonical strings differ.
+Downgrade applies the inverse: re-adds ``memory:admin`` on DCR clients
+whose scope matches the post-narrow canonical. Same DCR-only filter
+(``owner_id IS NULL``) and the same canonical-string residual
+limitation apply in reverse — a DCR client that registered AFTER this
+migration with the exact post-narrow canonical string would have
+``memory:admin`` ADDED on downgrade (symmetric privilege escalation).
+Operators running the downgrade should re-issue narrower scopes to
+affected clients via the OAuth admin endpoints if this matters.
 
 Rollback (operator runbook): if Claude Code MCP DCR breaks post-deploy
 because a pre-SEP-835 SDK invalidates the narrowed token (the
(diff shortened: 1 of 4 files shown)


JFK:
Thanks for the catch — addressed in 8307d75. The two findings are linked: both surface that `WHERE scope = :canonical` alone over-claims the R1 preservation guarantee.

**Fix applied** (suggestion *b* — tighten selection criteria, partial; plus suggestion *a* for the residual):

1. **`owner_id IS NULL` filter** added to both upgrade and downgrade WHERE clauses. This restricts the migration to DCR-registered clients only — admin-managed clients have non-null `owner_id` per `d04_519_oauth_owner_nullable` and are now never touched even if their scope string happens to match the canonical. This eliminates the bulk of the over-reach you identified.

2. **Residual limitation documented honestly** in the module docstring: among DCR clients, the migration still cannot distinguish a client that explicitly requested the full canonical string *in the server's exact ordering* from a defaulted client. In practice this collision is rare (requires exact-ordering match against the server's canonical string), and affected clients can recover by re-registering or via operator re-grant. The docstring no longer claims a blanket "guarantees" — it states the limitation and the workaround.

3. **Downgrade asymmetry** acknowledged in the downgrade docstring: the inverse `WHERE scope = :new_scope` has the symmetric privilege-escalation corner (a DCR client registered post-upgrade with the exact post-narrow canonical would be re-granted admin on downgrade). Same `owner_id IS NULL` filter applied; same exact-ordering caveat documented.

4. **Pin test** `test_where_clause_restricts_to_dcr_clients` added — asserts the literal `owner_id IS NULL` fragment is present in both upgrade and downgrade source via `inspect.getsource`. Cheap regression guard against future refactor accidents.

The per-DCR-client exact-ordering edge case isn't fully solvable without a schema change to record "explicit-vs-defaulted" at registration time (your suggestion *b* in full form), which is out of scope for this PR. Recorded for the #608 follow-up sweep.

PR #617, opened 2026-05-12
feat(mcp,api): list_tags tool for tag discovery per context (#614)

### LINKED ISSUE #614: feat(mcp,api): list_tags tool for tag discovery per context
## Context

When LLM clients call `remember`, they have no way to discover which tags
already exist in a context. This causes tag drift — semantically identical
tags appear under different spellings (`troubleshoot` / `troubleshooting`
/ `trouble-shoot`), which silently degrades the precision of
`recall(filters={"tags": [...]})` over time.

A lightweight `list_tags` MCP tool + REST endpoint would let clients
inspect the existing tag vocabulary before writing or filtering, without
imposing a rigid taxonomy (we keep the free-form tag model).

## Proposed shape

**MCP tool**: `list_tags`
**REST**: `GET /api/v1/contexts/{context_id}/tags`

Parameters:
- `context_id` (required)
- `limit` (default 50, cap ~500)
- `min_count` (default 2 — hide one-off tags)
- `sort` — `"count"` (default) or `"recent"`

Return:
```json
[
  { "tag": "troubleshooting", "count": 42, "last_used_at": "2026-05-10T..." },
  { "tag": "pattern",         "count": 31, "last_used_at": "2026-05-12T..." }
]
```

## Use cases

1. LLM calls `list_tags` before `remember` → reuses existing tag spellings
2. LLM calls `list_tags` before `recall` → builds accurate `filters.tags`
3. UI surface: tag cloud / autocomplete on context detail page

## Out of scope for v1

- Tag co-occurrence (`tags often used together with X`) — defer to v2
- Tag normalization / merging — explicit non-goal; keep tags free-form
- Inclusion inside `get_context_info` — would bloat that payload; keep standalone

## Implementation notes

- Naive `SELECT tag, count(*), max(created_at) FROM memory_tags WHERE context_id=? GROUP BY tag` should be fine up to ~10^4 unique tags per context
- Add GIN index on `memories.tags` if not present (verify)
- Consider a materialized view or cached aggregate only if profiling shows it; don't pre-optimize

## Acceptance

- [ ] MCP tool `list_tags` registered and documented
- [ ] REST endpoint with same semantics
- [ ] OpenAPI spec updated
- [ ] Tests: empty context, large context (>1k tags), min_count filtering, both sort modes
- [ ] Pagination decision (limit-only vs cursor) recorded

## Related

- Design discussion on tag drift / observability vs management (2026-05-12)

JFK:
Closes #614

## Summary
- Adds a read-only `list_tags` MCP tool + REST endpoint so LLM clients can discover existing tag spellings before `remember()` / `recall()`, mitigating tag drift (`troubleshoot` vs `troubleshooting` silently degrading filter precision).
- Service `ContextService.aggregate_tags()` mirrors the canonical CTE pattern in `sleep_tasks._refresh_hub_tag_cache` — DISTINCT-per-memory dedup, soft-delete and workspace boundary filters.
- `RelatedTagItem` schema extended with optional `last_used_at` so `recall.related_tags` and `list_tags` share one client-side type.

## Implementation notes
- **Service** (`backend/src/services/context_service.py`): new `aggregate_tags(user_id, context_id, *, limit, min_count, sort, prefix)` returning `{"context_name": str, "rows": [...]}`. Module-level `TagSortMode = Literal["count","recent","alpha"]` is the single source of truth shared with the REST `Query(...)` and service signature (same pattern as `SleepMode`, `SearchMode`). Access check runs once via `self.get_context()`; the MCP handler and REST route both delegate without pre-resolving.
- **REST route** (`backend/src/api/routes/contexts.py`): `GET /contexts/{context_id}/tags` mirrors `get_context_stats`; `APIKeyOrSessionUser` auth matches the `get_memory_usage_stats` analytics precedent. Bounded query params (`limit` 1–500, `min_count` 1–10000, `prefix` max 200 chars).
- **MCP tool** (`backend/src/mcp_server/tools/context.py`): `handle_list_tags` mirrors `handle_get_context_info` structurally. Added to `_RATE_LIMIT_EXEMPT_TOOLS` (read-only), `TOOL_TIMEOUTS` (10s), `_definitions.py` (with `readOnly: True` + "Call this BEFORE remember/recall" guidance), and the registry. Uniform `context_not_found` on `NotFoundException` (CWE-639 — never leak deny reason); `db.rollback()` on every error path.
- **Schema** (`backend/src/models/schemas.py`): `RelatedTagItem` switched `BaseModel → TZAwareBaseModel` (UTC-Z serialization for the new optional `last_used_at`); existing `recall.related_tags` callers unaffected (additive, defaults `None`).
- **SQL**: `WITH scope AS (… WHERE workspace_id=… AND context_id=… AND deleted_at IS NULL AND cardinality(tags)>0) → SELECT DISTINCT id, unnest(tags)::text AS tag, last_at → GROUP BY tag with MAX(last_at) → ORDER BY <sort_clause> LIMIT`. `sort` allow-list validated before f-string interpolation; `prefix` `%`/`_`/`#` escaped (`ESCAPE '#'`) before binding as `:prefix_pattern`.
- **Pagination decision recorded**: limit-only (cap 500), no cursor — discovery use case doesn't warrant the API-contract surface.
- **`min_count` default = 1**: chosen over issue body's `2` so drift typos are visible by default (DX pit of success).

## Pre-PR review summary
- gate2 mode: advisor-only (no binary gate configured)
- audit: skipped
- binary_gate: (none)
- /claude-c-suite:cso: green
- /claude-c-suite:qa-lead: green
- /claude-c-suite:cto: green
- gate1: yellow via /ask (CIO lens) → DX Lead consult → all yellow concerns addressed pre-implementation
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/614-feat-feat-mcp-api-list-tags-tool-for-tag-disc.gate1.md`
- `~/.claude/cache/gh-issue-driven/614-feat-feat-mcp-api-list-tags-tool-for-tag-disc.gate2.md`

## Test plan
- [x] 19 unit tests (REST 5 + MCP 14): happy / empty / 404 / 422 / args forwarding / 7 bound-violation cases
- [x] 12 integration tests (PG-required): DISTINCT-per-memory dedup, soft-delete exclusion, cross-workspace boundary, `min_count`, `prefix`-as-literal, 3 sort modes, `last_used_at`-MAX
- [x] ruff clean, pyright clean
- [x] 157 surrounding tests pass (no regressions)
- [ ] CI integration lane runs the 12 PG tests
- [ ] Manual smoke: `list_tags` via Claude Code MCP after deploy

## Out of scope (separate follow-ups)
- SDK `client.list_tags(...)` in `kagura-memory-python-sdk`
- Frontend tag-cloud / autocomplete UI on context detail page
- Compound `(workspace_id, context_id)` index — defer until profiling shows need

### DIFF
--- a/backend/src/api/routes/contexts.py
+++ b/backend/src/api/routes/contexts.py
@@ -22,8 +22,9 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from auth.dependencies import APIKeyOrSessionUser, SessionUser, get_current_user
 from db.base import get_db
 from models.api_base import TZAwareBaseModel
+from models.schemas import RelatedTagItem
 from models.sleep import SleepMode
-from services.context_service import ContextService
+from services.context_service import ContextService, TagSortMode
 from utils.datetime import to_utc_iso
 from utils.exceptions import NotFoundException, ValidationError
 from utils.logger import get_logger
@@ -256,6 +257,21 @@ class ContextStatsResponse(BaseModel):
     context_name: str = Field(..., description="Context name")
     memory_count: int = Field(..., description="Number of memories in context")
     status: str = Field(..., description="Context status")
+
+
+class ContextTagsResponse(BaseModel):
+    """Response model for ``GET /contexts/{context_id}/tags`` (Issue #614).
+
+    Envelope wrapper consistent with ``ContextListResponse``. ``RelatedTagItem``
+    inherits ``TZAwareBaseModel`` so the per-item ``last_used_at`` datetime
+    serializes with an explicit UTC ``Z`` suffix when populated.
+    """
+
+    context_id: UUID = Field(..., description="Context UUID")
+    tags: list[RelatedTagItem] = Field(
+        ..., description="Aggregated tag info, sorted per the `sort` parameter."
+    )
+    total: int = Field(..., description="Number of tags returned (post-filter, post-limit).")
 
 
 # ============================================================================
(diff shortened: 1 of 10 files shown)


JFK:
## Loop 1: Copilot review addressed

**Applied fixes** (commit 8b… on this branch):

1. `_LIST_TAGS_SORT_MODES` now derived from `TagSortMode` via `typing.get_args()` — `TagSortMode` is the sole source of truth, the tuple cannot drift. (`context_service.py:1086`)
2. `ContextService.aggregate_tags()` validates `limit` (1-500) and `min_count` (1-10000) bounds itself with `ValidationError`. Public method, defensive. (`context_service.py:1095`)
3. MCP handler uses `type(x) is int` / `type(x) is str` instead of `isinstance` for `limit` / `min_count` / `prefix`, so `bool` (which subclasses `int`) is rejected and the `or ""` shorthand no longer silently coerces `0`/`False`. (`context.py:856`)
4. `test_cross_workspace_does_not_leak` renamed to `test_workspace_id_filter_blocks_cross_workspace_leak` and rewritten to seed a memory rooted in ws_b with `context_id=ctx_a.id` — this now genuinely exercises the `workspace_id` predicate (the CTE's `WHERE workspace_id = …` is what blocks leakage; the schema has no compound `(ws_id, ctx_id)` FK). (`test_list_tags_aggregation.py:201`)

Added parametrized test cases for bool / wrong-type rejection on `limit` / `min_count` / `prefix` to lock the strictness. Local: 23 unit tests pass (was 19), ruff clean, pyright 0 errors.

---

**Non-actionable** (`test_list_tags_aggregation.py:37`, naive datetime concern):

This codebase's DB columns are intentionally `TIMESTAMP WITHOUT TIME ZONE` — naive UTC by convention. See `.claude/rules/backend.md`:

> *"DB columns are TIMESTAMP WITHOUT TIME ZONE — naive UTC by convention. Until #490 migrates them, the API layer is the only place ensuring wire-format datetimes are unambiguous (#489)."*

The API layer's `TZAwareBaseModel` serializes naive datetimes with a `Z` suffix on the wire, but the storage convention is naive UTC. Using `datetime(..., tzinfo=timezone.utc)` in the integration test would actually be **inconsistent** with how `Memory.created_at` / `updated_at` are written everywhere else in the codebase (e.g., `tasks/sleep_tasks.py`, `services/memory_service.py`). The "recent" sort and `last_used_at` invariants in this test depend on the seeded values being directly comparable with values inserted via the ORM in production paths — both are naive.

When #490 lands and the columns migrate to `timestamptz`, all tests across the suite (not just these) will need to be updated together; this is tracked as a separate migration concern.

🤖 Generated via /gh-issue-driven:review (manual)

Issue #475, opened 2026-04-28
feat(observability): full-pipeline embedding token tracking (recall + sleep phases 1/2)


JFK:
## Summary
Track embedding API token usage across **all** call sites — not just Sleep Maintenance reindex phase. Current state has 3 gaps that combine to under-report embedding cost:

1. `EmbeddingService.embed()` returns only the vector (`list[float]`), discarding the OpenAI `response.usage.total_tokens` field at `backend/src/services/embedding_service.py:221`.
2. Sleep Phase 1 (`edge_discovery._find_candidates`) and Phase 2 (`dedup_merge._find_similar_pairs`) call `embed()` extensively but **never increment** `result.embedding_calls_used` — the counter is only updated in `reindex.py:93`.
3. Recall path (`mcp_server/tools/memory.py:236`) also calls `embed()` for query vectors with no usage tracking.

The result: today's `sleep_reports.embedding_calls_made` is "reindex-phase API calls only", and #471's new `embedding_tokens` column inherits the same gap.

Sister to #471 (cost-grade schema, v0.15.0) and #474 (comprehensive ledger, v0.16.0). #471 ships with reindex-only embedding tracking by design — this issue closes the gap.

## Why this is a separate issue
#471's primary scope is **schema design + LLM cost tracking**. Pulling embedding instrumentation across the entire pipeline into #471 would:
- Require a breaking change to `EmbeddingService.embed()` (or add a parallel method) + audit every caller
- Add 4-5 phases × N embedding call sites × tests
- Multiply #471's PR size and review surface

Cleaner to ship #471 with reindex-only embedding tracking (matches current `embedding_calls_made` semantics, no behavior regression) and close the full-pipeline gap here.

### 1. Extend `EmbeddingService` to surface token counts
- Add a new method `embed_with_usage(text: str) -> tuple[list[float], int]` that returns vector + token count from `response.usage.total_tokens`. Keep `embed()` as-is for callers that don't care.
- Cache hit case: token count is `0` (cache hits don't consume API tokens).
- All callers that want cost tracking switch to `embed_with_usage()`.

### 2. Instrument Sleep Phase 1 and Phase 2
- `edge_discovery._find_candidates`: capture embedding tokens, increment `self._embedding_calls_used` and `self._embedding_tokens_used` (new instance var).
- `dedup_merge._find_similar_pairs`: same.
[...]

PR #648, opened 2026-05-14
feat(db): add llm_call_log event store + writer skeleton (#474 partial)

JFK:
Partially addresses #474

## Summary

- Adds `llm_call_log` append-only event store for non-Sleep LLM/embedding/rerank API calls — companion to `sleep_reports` (run-shaped). #472 cost aggregation UNION ALLs both tables and groups by `caller`.
- `LlmCallLogWriter` service skeleton mirroring `SleepReporter` pattern (session-injected, `db.add` + `flush`, no commit, raise on error).
- Pricing seed rows for Voyage `rerank-2`/`rerank-2-lite` + Cohere `rerank-3.5` + Cohere `embed-multilingual-v3` (separate alembic revision for rollback granularity).
- Pilot call sites (`reranker_service` / `embedding_service` / recall path) **deferred to #475 PR-3** per the 3-PR split decision.

## 3-PR split (memory `45d4b9fe`)

| PR | Scope | Status |
|----|-------|--------|
| PR-1 | #475 Sleep Phase 1/2 embedding tracking | MERGED (#647, `1ba90db4`) |
| **PR-2** | **#474 schema-design** | **this PR** |
| PR-3 | #475 Task 4 recall path instrumentation | deferred (gated on this PR) |

### Schema (`backend/src/models/llm_call_log.py` + `e12_474_llm_call_log`)

- BigInt PK, append-only, no FK on `workspace_id`/`context_id` (audit preservation across entity deletion).
- 5 CHECK constraints: `caller` / `call_type` / `paid_by` / `cost_usd >= 0` / **`call_metadata` 4 KB octet_length** (first JSONB size cap in this codebase, defends against prompt-body storage).
- 2 indexes: `(occurred_at)` (admin all-tenant date range) + `(workspace_id, occurred_at)` (per-tenant aggregation; mirrors `idx_sleep_reports_workspace_source_started`).
- `paid_by` mirrors `sleep_reports.paid_by` for the #472 UNION ALL leg.

### Writer (`backend/src/services/llm_call_log_writer.py`)

- `_FORBIDDEN_CALLERS = {"sleep"}` — Sleep cost stays in `sleep_reports` + `sleep_report_llm_usage` (memory `9db43e36` canonical pattern). CHECK is a superset for forward-compat.
- `_FULL_IDENTITY_CALLERS = {"recall", "rerank", "ask"}` — Python-side nullability matrix enforcement (DB columns stay nullable to support `admin` backfill jobs).
- `cost_usd` = **write-time snapshot** from `LLMPricingService.compute_cost_usd`; pricing miss → `cost_usd=0` + `call_metadata.pricing_miss=true` (never raises on hot path).
- Negative usage values rejected with `ValueError` (defense against silent under-count).
- Import-time assert: `_USAGE_TO_UNIT_TYPE` keys ⊆ `LLMCallLog.__table__.columns` (catches future column rename).

### Alembic split

- `e12_474_llm_call_log` — schema only.
- `e13_474_pricing_seeds` — 4 seed rows at `effective_from=2026-04-28`. Targeted downgrade by `(effective_from, provider)` so c03 seeds (same date, different providers) survive partial rollback.

## Gate1 yellow points → Acceptance (CIO lens, 2026-05-14)

| # | Point | Resolution |
|---|---|---|
| 1 | `cost_usd` write-time vs lazy compute | write-time snapshot; pricing miss flag in `call_metadata` |
| 2 | partition/retention deferral | 100M-row threshold documented as escalation trigger |
| 3 | Sleep dual-write decision | v1: Sleep → `sleep_reports` only; #472 UNION ALL aggregates |
| 4 | caller nullability matrix | docstring table + writer Python asserts |
| 5 | `call_metadata` PII defense | 4 KB CHECK constraint (first JSONB size cap) |
| 6 | seed-row alembic split | e12 (schema) + e13 (seeds) separate revisions |
| 7 | UNION query shape index match | `(workspace_id, occurred_at)` mirrors sleep_reports leg |

## Out of scope (deferred)

- Pilot hooks on `reranker_service` / `embedding_service` → **#475 PR-3** (recall path embedding/rerank instrumentation, gated on this PR's schema)
- `#472` aggregation API extension → separate PR
- Partitioning / retention implementation → follow-up issue at 100M-row threshold

## Test coverage

- 11 schema integration tests (CHECK reject paths, DDL audit, server defaults, indexes, round-trip)
- 17 writer unit tests (validation, cost computation, pricing miss propagation, negative-value rejection, UUID coercion)
- 78 schema_drift tests pass (auto-detects new model)
- 4 alembic forward/rollback tests pass
- ruff: clean

## Follow-ups (non-blocking)

- e13 seed-row presence verification test (`SELECT count(*)` against the 4 rows landed) — gap shared with `c03_471_seed_pricing`
- `call_metadata` 4 KB CHECK boundary tests (exactly-4096, multibyte unicode, empty-dict) — pin recall-path #475 PR-3 Japanese-content behavior
- Application-layer allowed-key whitelist for `call_metadata` → ship in #475 PR-3 when call sites materialize
- Doc nit: extend the 4 KB CHECK comment with "apply only when JSONB is PII-sensitive AND has no application-layer key whitelist" so the pattern doesn't proliferate

## Pre-PR review summary

- gate2 mode: advisor-only (no binary gate configured)
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: yellow via /ask (CIO lens, 7 yellow points addressed at design time)
- review provider: copilot

- `~/.claude/cache/gh-issue-driven/474-feat-feat-observability-comprehensive-llm-cos.gate1.md`
- `~/.claude/cache/gh-issue-driven/474-feat-feat-observability-comprehensive-llm-cos.gate2.md`

### DIFF
--- a/backend/alembic/versions/e12_474_llm_call_log.py
+++ b/backend/alembic/versions/e12_474_llm_call_log.py
@@ -38,9 +38,10 @@ Out of scope (deferred to follow-ups):
 Revision ID: e12_474_llm_call_log
 Revises: e11_merge_e10_heads
 
-NOTE: Revision ID is 20 chars (alembic_version.version_num is VARCHAR(32),
-asyncpg raises StringDataRightTruncationError above 25 — see c02_471
-header for the canonical note).
+NOTE: Revision ID is 20 chars, well within the VARCHAR(32) limit on
+``alembic_version.version_num``. Revision IDs above 32 chars trigger
+``asyncpg.exceptions.StringDataRightTruncationError`` at upgrade time —
+see ``c02_471`` header for the canonical note describing the limit.
 """
 
 from collections.abc import Sequence
--- a/backend/alembic/versions/e12_474_llm_call_log.py
+++ b/backend/alembic/versions/e12_474_llm_call_log.py
@@ -1,0 +1,144 @@
+"""Add llm_call_log: append-only event log for non-Sleep LLM/embedding/rerank calls (#474).
+
+Issue #474: Companion to ``sleep_reports`` (run-shaped) — captures
+per-call cost events from recall reranking, recall embedding (wired up
+in #475 PR-3), future ``ask()`` synthesis, and one-off admin actions.
+Cost aggregation in #472 UNION ALLs ``sleep_reports`` and ``llm_call_log``
+on ``(workspace_id, occurred_at)`` and groups by ``caller`` for the
+dashboard breakdown.
+
+This revision adds schema only. Pricing seed rows for Voyage rerank-2 /
+Voyage rerank-2-lite / Cohere rerank-3.5 / Cohere embed-multilingual-v3
+land in the next revision (``e13_474_pricing_seeds``) so schema and
+data fixes have independent rollback units.
+
+Design notes (mirror module docstring of ``models/llm_call_log.py``):
+
+- ``cost_usd`` is computed at write time from ``llm_pricing`` and stored
+  as a snapshot (NOT NULL DEFAULT 0). Pricing misses store 0 + a flag in
+  ``call_metadata``.
+- ``call_metadata`` is JSONB with a 4 KB CHECK constraint defending
+  against accidental storage of prompt bodies. This is the codebase's
+  first schema-level JSONB size cap.
+- ``paid_by`` mirrors ``sleep_reports.paid_by`` so the #472 UNION ALL
+  groups both legs on the same axis.
+- ``caller`` CHECK includes 'sleep' for forward-compat (future migration
+  may unify) but the writer service refuses to emit it today — Sleep
+  cost stays in ``sleep_reports`` + ``sleep_report_llm_usage``.
+- Two indexes: ``(occurred_at)`` for the admin-wide query,
+  ``(workspace_id, occurred_at)`` for per-tenant aggregation. Mirrors
+  ``idx_sleep_reports_workspace_source_started`` so the UNION ALL has
+  matching index support on both legs.
+
+Out of scope (deferred to follow-ups):
+- Partitioning / retention (escalation trigger: 100M rows).
+- Recall path instrumentation itself (#475 PR-3).
+- Aggregation API extension (#472).
+
+Revision ID: e12_474_llm_call_log
+Revises: e11_merge_e10_heads
+
+NOTE: Revision ID is 20 chars (alembic_version.version_num is VARCHAR(32),
+asyncpg raises StringDataRightTruncationError above 25 — see c02_471
+header for the canonical note).
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "e12_474_llm_call_log"
+down_revision: str | Sequence[str] | None = "e11_merge_e10_heads"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Create llm_call_log with CHECK constraints + 2 indexes."""
+    op.create_table(
+        "llm_call_log",
+        sa.Column("id", sa.BigInteger, primary_key=True, autoincrement=True),
+        # Naive DateTime to match ``sleep_reports.started_at`` and
+        # ``llm_pricing.effective_from`` (the cost-snapshot join target).
+        sa.Column("occurred_at", sa.DateTime, nullable=False),
+        sa.Column("user_id", sa.String(255), nullable=True),
+        sa.Column("workspace_id", UUID(as_uuid=True), nullable=True),
+        sa.Column("context_id", UUID(as_uuid=True), nullable=True),
+        sa.Column("caller", sa.String(20), nullable=False),
+        sa.Column("call_type", sa.String(20), nullable=False),
+        sa.Column("provider", sa.String(50), nullable=False),
+        sa.Column("model", sa.String(100), nullable=False),
+        # Per-call usage columns. Any subset may be NULL per ``call_type``;
+        # the writer service enforces the correct field is set.
+        sa.Column("input_tokens", sa.Integer, nullable=True),
+        sa.Column("output_tokens", sa.Integer, nullable=True),
+        sa.Column("cached_input_tokens", sa.Integer, nullable=True),
+        sa.Column("cache_write_tokens", sa.Integer, nullable=True),
+        sa.Column("embedding_tokens", sa.Integer, nullable=True),
+        sa.Column("rerank_tokens", sa.Integer, nullable=True),
+        sa.Column("rerank_search_units", sa.Integer, nullable=True),
+        # Write-time cost snapshot. NOT NULL DEFAULT 0 so aggregation
+        # queries don't need COALESCE; pricing misses store 0 + a flag
+        # in ``call_metadata``.
+        sa.Column(
+            "cost_usd",
+            sa.Numeric(12, 6),
+            nullable=False,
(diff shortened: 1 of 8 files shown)


JFK:
**Re: github-code-quality "Unused import" finding on `backend/tests/conftest.py:20`**

This is an intentional side-effect import for SQLAlchemy table registration with `Base.metadata` (matching the existing `models.llm_pricing` / `models.sleep` / `models.analysis` imports on the surrounding lines). The `# noqa: F401  isort: skip` comment opts out of the standard linter unused-import / import-ordering warnings. The same pattern is used in `backend/src/models/__init__.py:11-15`.

The bot's check doesn't appear to honor the `# noqa: F401` directive — false positive, no action needed. Verified ruff lint passes against the file (`make lint` clean).

JFK:
**Copilot loop 1 — 3 actionable findings addressed**

1. **`models/llm_call_log.py:227` — `server_default` form** → fixed. Now uses `text("'platform'")` for `paid_by` and `text("0")` for `cost_usd`, matching `SleepReport.paid_by` (`models/sleep.py:153-158`). Migration form (`e12_474_llm_call_log.py:97-101`) was already correct; this brings the model into byte-for-byte parity, eliminating potential `test_schema_drift` ambiguity.

2. **`services/llm_call_log_writer.py:177` — `call_type` contract not enforced** → fixed. Added `_CALL_TYPE_ALLOWED_USAGE` matrix and inline validation in `record()`. Cross-axis mixing (e.g., `call_type="completion"` + `embedding_tokens > 0`) now raises `ValueError` at the writer boundary instead of silently producing under-counted rows. Rerank additionally rejects "both `rerank_tokens` AND `rerank_search_units`" since Voyage and Cohere are mutually exclusive providers. Degenerate all-zero rows are still allowed (documented audit contract preserved). Added 5 new unit tests for the cross-axis contract.

3. **`alembic/versions/e13_474_pricing_seeds.py:162` — `downgrade()` too broad** → fixed. Replaced `WHERE provider IN ('voyage', 'cohere') AND effective_from = ...` with exact `(provider, model, unit_type, effective_from)` tuple matching via a shared `_SEED_ROWS` authoritative list. Both `upgrade()` and `downgrade()` now iterate the same list, so a fifth seed row added later only requires editing one place. Defends against collateral damage if a future migration adds an unrelated voyage/cohere row at the same `2026-04-28` snapshot.

All checks green: 115 tests pass (11 schema integration + 22 writer unit + 78 schema-drift + 4 alembic forward/rollback), ruff clean.

JFK:
**Copilot loop 2 — 3 actionable findings addressed**

1. **`models/llm_call_log.py:221` + `e12_474_llm_call_log.py:89` — `cost_usd` precision** → fixed. `NUMERIC(12, 6)` → `NUMERIC(14, 10)`, mirroring `llm_pricing.price_per_unit`. Sub-microcent costs (e.g., 20 embedding tokens × $0.02/1M = $0.0000004) now survive the row instead of rounding to 0 and silently undercounting SUM aggregation. 4 digits before the decimal still caps per-call cost at ~$9,999.99 — orders of magnitude above any plausible single API call. Both column comments updated to document the precision rationale.

2. **`services/llm_call_log_writer.py:223` — axis guard `value > 0` blind spot** → fixed. The loop-1 cross-axis validation only ran the `column_name not in allowed_axes` check inside the `if value > 0:` branch, so a caller could pass `embedding_tokens=0` on a completion row and the row would store with `embedding_tokens IS NOT NULL` (dirty signal for downstream filters that treat non-NULL as "this axis was relevant"). The check now runs on **any non-None value** — `embedding_tokens=0` on completion raises `ValueError`. Added `test_completion_rejects_zero_value_on_disallowed_axis` to pin the edge case.

All checks green: 116 tests pass (11 schema + 23 writer unit + 78 schema-drift + 4 alembic), ruff clean.

JFK:
**Copilot loop 3 — 2 actionable findings addressed**

1. **`services/llm_call_log_writer.py:247` — rerank exclusivity blind spot for value=0** → fixed. The exclusivity check now tracks `rerank_axes_populated` on any non-None value (including 0), not just `value > 0`. This is the mirror of the loop-2 `_CALL_TYPE_ALLOWED_USAGE` fix applied to the rerank-specific exclusivity rule. `rerank_tokens=0` + `rerank_search_units=10` now raises `ValueError` at the writer boundary instead of writing both columns as non-NULL. Added `test_rerank_rejects_both_axes_with_zero_value` to pin the case.

2. **`services/llm_call_log_writer.py:131` — class name initialism inconsistency** → fixed. Renamed `LlmCallLogWriter` → `LLMCallLogWriter` across the source (class definition) and tests (import + fixture + docstring). Now consistent with `LLMPricingService` and `LLMCallLog`. Done before #475 PR-3 starts importing the class at recall call sites.

All checks green: 117 tests pass (11 schema + 24 writer unit + 78 schema-drift + 4 alembic), ruff clean.

JFK:
**Copilot loop 4 — 4 actionable findings addressed**

1. **`services/llm_call_log_writer.py:380` — short-circuit on first pricing miss** → fixed. `_compute_cost_usd` now returns `(Decimal("0"), True)` immediately on the first miss instead of continuing to query remaining axes. The result was always collapsing to `cost_usd=0 + pricing_miss=true` regardless of how many axes missed, so the extra lookups added no signal — saves DB round-trips on the recall hot path's "new model not yet seeded" common case. Comment updated.

2. **`services/llm_call_log_writer.py:273` — pricing_miss log level** → fixed. Dropped from `logger.warning` to `logger.debug`, matching `LLMPricingService.lookup()`'s lower-level "no row" log path. Misses are expected (newly-deployed model before seeds land; exotic provider) and `call_metadata.pricing_miss=true` on the row carries the signal to dashboards — WARNING on every miss would create alert fatigue on a recall hot path. Comment block explains the choice.

3. **`alembic/versions/e12_474_llm_call_log.py:43` — header revision-ID comment** → fixed. The "asyncpg above 25 chars" wording was a misread of `c02_471`'s note (which says revision ID is 25 chars long *AND* the VARCHAR(32) limit is the true constraint). e12's header now aligns with the codebase's established "VARCHAR(32) limit" wording.

4. **`alembic/versions/e13_474_pricing_seeds.py:37`** — same fix as #3 (pair finding).

All checks green: 117 tests pass, ruff clean.

JFK:
**Copilot loop 5 — 1 actionable finding addressed (max_loops reached)**

**`services/llm_call_log_writer.py:252` — rerank "exactly one" under-enforced** → fixed. The loop-3 fix rejected `len > 1` (both populated) but accepted `len == 0` (neither populated), contradicting the writer/docstring contract that says rerank must populate exactly one axis. Since `reranker_service` always has a usage figure from the provider (Voyage returns `rerank_tokens`, Cohere returns search units), a zero-axis rerank row is a caller-side bug, not a degenerate audit case (this asymmetry is intentional — completion DOES allow all-None as a documented audit path because LLM error returns might land that way). Check now raises on `len != 1`. Added `test_rerank_rejects_no_axis_populated` to pin the new path.

All checks green: 118 tests pass, ruff clean.

---

**Loop status**: this is loop 5 of the default `copilot.max_loops=5`. Each prior loop converged on real-but-progressively-smaller findings (loop 1 = 3 findings, loop 2 = 3, loop 3 = 2, loop 4 = 4, loop 5 = 1). The shape is healthy — Copilot is moving from architectural concerns (cost_usd precision, dual-write boundary) to API/log polish (class name initialism, log level) to contract-tightening edge cases (rerank exactly-one). Further iterations would likely target documentation nits rather than semantics.

Per the `/gh-issue-driven:ship` workflow, the loop exits here with `exit_reason=max_loops`. Re-entry via `/gh-issue-driven:review` is available if you want another pass.

PR #657, opened 2026-05-15
fix(auth): apply signup gate to Google OAuth (#655)

### LINKED ISSUE #655: feat(auth): apply signup gate to Google OAuth (follow-up to #358)
## Summary

The admin-configurable signup gate (`signup_gate_service`, Issue #358 Phase 1) currently has an explicit Google pass-through: `provider == "google"` returns `None` without consulting the allowlist. The `signup_allowlist` table itself is also GitHub-only (`github_user_id` NOT NULL, no Google column). The original Phase 1 design assumed Google's OAuth Consent Screen would enforce signup access via the test-users list.

**The assumption does not hold for non-sensitive scopes.** With `openid` / `userinfo.email` / `userinfo.profile` (what this app uses), Google in "Testing" status shows the "未確認のアプリ" warning but lets any user click through `詳細 > 安全ではないページに移動` and complete OAuth. The test-users list is only strictly enforced for sensitive / restricted scopes.

A production audit confirmed the impact: Google accounts outside the configured test-user list have completed OAuth and now hold workspaces. The alternative remediations (publishing status = Internal; switching to sensitive scopes) are not available — Internal requires a Google Workspace domain (not held), and adding sensitive scopes degrades the consent UX without functional benefit. The only durable fix is to gate Google OAuth on the backend, just as GitHub already is. **GitHub side is intentionally unchanged in this issue** — it already works as designed.

## Context

**Priority**: high (production has an active inflow channel for out-of-scope Google signups; today every hour the gate is open is more cleanup work later)
**Suggested labels**: bug
**Suggested milestone**: v0.16.3 (Code Health & Cleanup) — closest near-term tactical release; the v0.16.1/v0.16.2 themes (cost/token & TZ migration) are off-topic, and waiting for v1.0.0 leaves the gap open through the launch window.

### Schema migration (`signup_allowlist`)

- Add `provider VARCHAR(20) NOT NULL DEFAULT 'github'` (CHECK `provider IN ('github','google')`). Existing rows backfill to `'github'`.
- Add `subject_id VARCHAR(255) NOT NULL`; copy `github_user_id` into it for the `provider='github'` rows.
- Add `subject_label VARCHAR(255) NOT NULL` for UI display (GitHub: `github_username`; Google: `email`). Not used for matching.
- Replace UNIQUE `(github_user_id, source)` with `(provider, subject_id, source)`.
- Replace INDEX `ix_signup_allowlist_github_user_id` with `ix_signup_allowlist_provider_subject` on `(provider, subject_id)`.
- `github_user_id` stays NOT NULL during the migration window (deprecated comment) — physical drop is a separate follow-up.

### Gate logic (`signup_gate_service.py`)

- Remove the `provider == "google"` pass-through (`signup_gate_service.py:98-100`).
- For Google: match on `subject_id = oauth_sub` (Google's `sub` claim — immutable). **Do not match on email** — email-change attacks must not bypass the gate.
- `mode == "manual"`: same allowlist requirement for both providers.
- `mode == "github_sponsors"` / `"both"` (Phase 2 modes): when `provider == "google"`, behave as `mode == "manual"` (sponsorship is GitHub-specific; document this fallback explicitly).

### Admin API / CLI

- Extend allowlist endpoints (`create / list / revoke`) to accept `provider`.
- CLI: add `--provider {github,google}` to the existing allowlist commands; for Google add `--sub <google_sub>` and `--email <label>`.
- Add a helper `utils/google_user.py` to resolve an existing `users.user_id` (= Google `sub`) by email so admins don't need to dig the sub out of the IdP. Pre-OAuth invitation (email-only allowlist entry, sub filled at first callback) is deferred to Phase 2.

### Existing-user semantics (side-effect callout)

- The migration **does not** auto-add existing `users` rows to the allowlist. Existing users continue to log in via the "Existing user → allowed" branch in `check_access` (`signup_gate_service.py:82-84`). The gate only affects **new** OAuth signups.
- Removing already-admitted out-of-scope users is a human/admin decision and lives in a separate follow-up issue. This issue closes the inflow only.

### Tests

- Unit: `SignupGateService.check_access` for `provider='google'` + `mode='manual'` with and without allowlist entry.
- Unit: `provider='google'` + existing `users` row → allowed (existing-user branch).
- Integration: `/api/v1/auth/google/callback` with mocked OAuth — allowlist miss returns 302 redirect with a reason; allowlist hit proceeds.
- Integration: existing GitHub allowlist path must not regress.
- Schema drift test + up/down migration test (`make test-integration`).

## Out of scope (deferred follow-ups)

- Email-domain allowlist (e.g. wildcard `@example.com`). Phase 2.
- Pre-OAuth invitation flow (email-only entry → sub auto-filled at first callback). Phase 2.
- Cleanup of already-admitted out-of-scope users (data retention / workspace transfer decisions). Separate issue.
- Web UI for allowlist management. Separate issue.
- Physical drop of the `github_user_id` column. Separate issue once all admin tooling has switched to `subject_id`.

## Acceptance Criteria

- Migration up/down passes `make test-integration`; existing rows correctly carry `provider='github'` and `subject_id = old github_user_id`.
- `provider='google'` + `mode='manual'`: allowlist miss → 302 with reason; allowlist hit → `ensure_user` runs.
- Existing Google users (rows already in `users`) continue to log in successfully even without an allowlist entry.
- GitHub login flow is unchanged (no regression in unit + E2E suites).
- Admin allowlist API accepts `provider`; CRUD works for both `github` and `google` entries.
- Documentation under `docs/` is updated to describe the dual-provider gate (the `#358` text and any related runbook).

## References

- Parent issue: `#358` — admin-configurable signup gate Phase 1.
- `backend/src/services/signup_gate_service.py:98-100` — the Google pass-through that this issue removes.
- `backend/src/models/signup_gate.py` — `SignupAllowlistEntry` model.
- `backend/src/api/routes/auth.py:481-487` — Google callback's `check_signup_access` call site.

JFK:
Closes #655

## Summary
- Removes the Google pass-through in `signup_gate_service.py` and gates Google OAuth uniformly with GitHub (the prior #358 Phase 1 design trusted Google's Consent Screen test-user list, which is advisory for non-sensitive scopes)
- Widens `signup_allowlist` from a GitHub-only key to `(provider, subject_id)` so Google rows match on the immutable OIDC `sub` claim (never email, to keep email-change attacks closed)
- Writes an `audit_logs` row on every blocked signup attempt — email HMAC'd via `hmac_sha256_hex`, plus IP/UA captured for admin triage
- Extends `POST /api/v1/admin/signup-gate/allowlist` to accept `{provider: "google", email}` alongside the legacy `{github_username}` shape (backward-compat, no flag day)

## Implementation notes
- **Migration `e14_655`**: adds `provider`/`subject_id`/`subject_label` with transitional `server_default`s, backfills GitHub rows (`subject_id=github_user_id`, `subject_label=github_username`), drops the server_defaults, swaps UNIQUE/INDEX onto the new key, then tightens to NOT NULL. Step 4.5 belt-and-suspenders re-UPDATE catches any stragglers from the deploy window. Legacy `github_user_id`/`github_username` columns stay NOT NULL during the migration window — physical drop deferred to a follow-up issue.
- **`utils/google_user.py`** (new): `resolve_google_sub_by_email` looks up an existing `users` row with `auth_provider='google'` to get the immutable `sub`. The invitee must complete Google OAuth at least once before they can be allowlisted (pre-OAuth invitation by email is a Phase 2 follow-up).
- **`mode=github_sponsors`/`both` + Google → manual fallback**: sponsorship is GitHub-specific, so Google falls back to manual allowlist semantics rather than raising NotImplementedError.
- **Existing users unchanged**: `_is_existing_user` branch still passes through; the gate only affects NEW signups. Removing already-admitted out-of-scope users is a separate operator decision and a separate follow-up.
- GitHub side intentionally untouched.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped (binary_gate = null)
- binary_gate: (none)
- cso: yellow — 2 findings, see below
- qa-lead: green
- cto: green
- gate1: green via /claude-c-suite:ask (PM lens)
- review provider: copilot

### CSO yellow findings (address in this PR or as follow-ups)

1. **`audit_logs.user_metadata.subject_label` stores plaintext email for Google rows** — defeats the HMAC design intent of `new_value_hash`. The codebase precedent (`backend/src/auth/roles.py:439-446`) HMACs the email and keeps `user_metadata` to provider-type only. Suggested fix: drop `subject_label` from `user_metadata` in `_record_blocked_signup` (2-line change).
2. **`User.email.ilike(email)` allows wildcard overmatch** when admin-supplied email contains `%` (RFC 5321 permits it in the local part). Admin-only path so bounded blast radius, but a correctness bug. Suggested fix: escape `%`/`_` before passing to `ilike`, or switch to `func.lower(...) == email.lower()` (1-line change).

- ~/.claude/cache/gh-issue-driven/655-fix-google-signup-gate.gate2.md

## Test plan
- [x] `make lint` (passes locally)
- [x] `pytest tests/services/test_signup_gate_service.py tests/utils/test_google_user.py` — 39/39 pass
- [ ] `make test-integration` (Docker — runs in CI): admin API + migration round-trip
- [ ] Verify migration upgrade + downgrade idempotency in CI
- [ ] Address CSO yellow findings (or accept as follow-up)
- [ ] Smoke test: Google user not in allowlist → blocked redirect with `?provider=google&sub=<first8>`
- [ ] Smoke test: Existing Google user (in `users` table) still logs in normally

### DIFF
--- a/backend/alembic/versions/e14_655_signup_allowlist_provider.py
+++ b/backend/alembic/versions/e14_655_signup_allowlist_provider.py
@@ -76,34 +76,30 @@ def upgrade() -> None:
     # the end of Step 2 (DROP DEFAULT) so post-migration INSERTs from new
     # code must supply real values — the columns having a permanent default
     # is exactly what we are NOT trying to bake in.
-    op.add_column(
-        "signup_allowlist",
-        sa.Column(
-            "provider",
-            sa.String(20),
-            nullable=True,
-            server_default=sa.text("'github'"),
-        ),
+    #
+    # Each ADD COLUMN goes through raw SQL with ``IF NOT EXISTS`` so a
+    # partial-rerun under autocommit mode self-heals (PR #657 Copilot loop
+    # 4 finding #3). PG 9.6+ has native ``ADD COLUMN IF NOT EXISTS`` so this
+    # is one statement, not a DO block.
+    op.execute(
+        sa.text(
+            "ALTER TABLE signup_allowlist "
+            "ADD COLUMN IF NOT EXISTS provider VARCHAR(20) DEFAULT 'github'"
+        )
     )
-    op.add_column(
-        "signup_allowlist",
-        sa.Column(
-            "subject_id",
-            sa.String(255),
-            nullable=True,
+    op.execute(
+        sa.text(
+            "ALTER TABLE signup_allowlist "
             # Empty-string sentinel. Step 2's backfill overwrites both
             # legacy NULL rows and any '' rows from the deploy window.
-            server_default=sa.text("''"),
-        ),
+            "ADD COLUMN IF NOT EXISTS subject_id VARCHAR(255) DEFAULT ''"
+        )
     )
-    op.add_column(
-        "signup_allowlist",
-        sa.Column(
-            "subject_label",
-            sa.String(255),
-            nullable=True,
-            server_default=sa.text("''"),
-        ),
+    op.execute(
+        sa.text(
+            "ALTER TABLE signup_allowlist "
+            "ADD COLUMN IF NOT EXISTS subject_label VARCHAR(255) DEFAULT ''"
+        )
     )
 
     # Step 2: backfill from existing GitHub-only columns.
@@ -136,26 +132,36 @@ def upgrade() -> None:
     op.alter_column("signup_allowlist", "subject_label", server_default=None)
 
     # Step 3: drop old UNIQUE + INDEX before tightening to NOT NULL, so the
-    # new constraints can sit on the new columns. ``type_="unique"`` on
-    # drop_constraint matches the codebase convention from a99 / b03.
-    op.drop_constraint(
-        "uq_allowlist_user_source",
-        "signup_allowlist",
-        type_="unique",
+    # new constraints can sit on the new columns. Raw SQL with ``IF EXISTS``
+    # guards (PR #657 Copilot loop 4 finding #3): self-heals if a prior run
+    # already dropped these and was interrupted before reaching the version
+    # bump, OR if an operator runs the migration manually with
+    # ``transaction_per_migration=False`` and the alembic_version row
+    # wasn't updated. Under the default transactional mode the guards are
+    # idempotency insurance; under autocommit mode they're actually load-
+    # bearing.
+    op.execute(
+        sa.text("ALTER TABLE signup_allowlist DROP CONSTRAINT IF EXISTS uq_allowlist_user_source")
     )
-    op.drop_index(
-        "ix_signup_allowlist_github_user_id",
-        table_name="signup_allowlist",
-    )
+    op.execute(sa.text("DROP INDEX IF EXISTS ix_signup_allowlist_github_user_id"))
 
     # Step 4: add CHECK constraint on provider. For this small table we
     # don't need the NOT VALID + VALIDATE dance from b03 (which is for
     # large tables where the scan would block writes); a direct ADD
     # CONSTRAINT is fine.
-    op.create_check_constraint(
-        "valid_signup_allowlist_provider",
-        "signup_allowlist",
-        "provider IN ('github', 'google')",
+    #
+    # Wrapped in a DO block that catches ``duplicate_object`` so a partial-
+    # rerun under autocommit mode self-heals. PG has no native
+    # ``ADD CONSTRAINT IF NOT EXISTS`` for CHECK constraints.
+    op.execute(
+        sa.text(
+            "DO $$ BEGIN "
+            "  ALTER TABLE signup_allowlist "
+            "    ADD CONSTRAINT valid_signup_allowlist_provider "
+            "    CHECK (provider IN ('github', 'google')); "
+            "EXCEPTION WHEN duplicate_object THEN NULL; "
+            "END $$"
+        )
     )
 
     # Step 4.5: belt-and-suspenders re-backfill. Step 2 caught the rows
@@ -206,15 +212,27 @@ def upgrade() -> None:
     # includes ``source`` for the same reason the old one did — a single
     # subject can legitimately have one ``manual`` row and one
     # ``github_sponsors`` row simultaneously.
-    op.create_unique_constraint(
-        "uq_allowlist_provider_subject_source",
-        "signup_allowlist",
-        ["provider", "subject_id", "source"],
+    #
+    # Both wrapped for idempotency (PR #657 Copilot loop 4 finding #3):
+    # UNIQUE goes through a ``DO $$ ... EXCEPTION WHEN duplicate_table``
+    # block because PG has no native ``ADD CONSTRAINT IF NOT EXISTS``
+    # for unique constraints. INDEX uses native ``CREATE INDEX IF NOT
+    # EXISTS`` which is the cleaner path.
+    op.execute(
+        sa.text(
+            "DO $$ BEGIN "
+            "  ALTER TABLE signup_allowlist "
+            "    ADD CONSTRAINT uq_allowlist_provider_subject_source "
+            "    UNIQUE (provider, subject_id, source); "
+            "EXCEPTION WHEN duplicate_table THEN NULL; "
+            "END $$"
+        )
     )
-    op.create_index(
-        "ix_signup_allowlist_provider_subject",
-        "signup_allowlist",
-        ["provider", "subject_id"],
+    op.execute(
+        sa.text(
+            "CREATE INDEX IF NOT EXISTS ix_signup_allowlist_provider_subject "
+            "ON signup_allowlist (provider, subject_id)"
+        )
     )
 
 
--- a/backend/alembic/versions/e14_655_signup_allowlist_provider.py
+++ b/backend/alembic/versions/e14_655_signup_allowlist_provider.py
@@ -154,18 +154,27 @@ def upgrade() -> None:
 
     # Step 4.5: belt-and-suspenders re-backfill. Step 2 caught the rows
     # that existed when the migration started; this catches any row that
-    # may have slipped in between Step 2 and now via the server_default
-    # path (e.g. a stray INSERT during a non-quiesced deploy window).
-    # Without this pass, a row inserted between Step 2 and Step 5 would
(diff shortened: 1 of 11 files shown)


JFK:
**Loop 1: addressed all 5 Copilot comments + 1 github-code-quality finding.**

| Finding | Resolution |
| --- | --- |
| Copilot #1 — `subject_label` plaintext PII in `audit_logs.user_metadata` | Dropped from metadata. HMAC in `new_value_hash` + `user_id` + `created_at` is now the sole correlation key. Defense-in-depth test added: assert plaintext email absent from row. |
| Copilot #2 — `ilike()` wildcard overmatch via admin-supplied `%`/`_` | Switched to `func.lower(User.email) == email.lower()`. Wildcards now treated as literals. |
| Copilot #3 — `AllowlistAddRequest` silently accepts mismatched fields | Added `@model_validator(mode="after")` that rejects with explicit 422 + actionable message. Two new tests cover both mismatched directions. |
| Copilot #4 — Inaccurate `# pragma: no cover` on `_record_blocked_signup`'s outer except | Removed. `TestRecordBlockedSignup.test_audit_failure_is_swallowed` already exercises this path. |
| Copilot #5 — Manual `f"..."` URL concatenation skips encoding | Switched to `urllib.parse.urlencode`. Provider/sub-prefix are URL-safe today but the change is correctness-by-default for future providers. |
| github-code-quality — Empty `except Exception: pass` in rollback handler | Now `except Exception as rollback_exc:` + `logger.debug(...)`. Preserves "never raise from audit path" contract. |

`make lint` clean. 39/39 unit tests pass locally.

JFK:
**Loop 2: addressed 4 of 8 Copilot comments on commit 9f6e0dc.**

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | `ilike()` wildcard injection (re-flag) | **Stale** — already fixed in commit 9f6e0dc (uses `func.lower(...) == lower(email)`). |
| 2 | `AllowlistAddRequest` mismatched-field acceptance (re-flag) | **Stale** — already fixed in commit 9f6e0dc (`_enforce_provider_field_match` validator present). |
| 3 | Migration Step 4.5 missing `subject_id IS NULL` predicate | **Fixed**. Mirrors Step 2's predicate so the zero-downtime-deploy edge case is covered: any old-code INSERT between Step 2.5 and Step 5 now gets re-backfilled before NOT NULL tightening. |
| 4 | Dead-code 422 raises in `POST /allowlist` handler | **Fixed**. The `_enforce_provider_field_match` validator catches missing fields at parse time, making the inline `raise HTTPException(422)` branches unreachable. Replaced with `assert ... is not None` for type narrowing. |
| 5 | Test docstring still mentions `.ilike(...)` | **Fixed**. Updated comment to describe the `func.lower(...) == lower(...)` shape the production code actually emits. |
| 6 | `_OAUTH_CALLBACK_ACTOR` underscore-prefix import | **Deferred**. Pre-existing codebase convention (cf. `_check_registration_allowed` lazy import); not regressed by this PR. |
| 7 | Double commit on shared session | **Documented**. Added invariant docstring to `_record_blocked_signup` noting the gate runs at OAuth-callback step 3.5 (before any other DB mutation), so the early commit is currently safe. The docstring also flags the switch to `begin_nested()` for any future refactor that moves DB work in front of the gate. |
| 8 | `func.lower(User.email)` misses B-tree index | **Deferred**. Admin-only path with bounded N today; warrants a separate perf issue if/when `users` table grows. |

`make lint` clean. 39/39 unit tests pass locally.

**Note on loop discipline**: comments #1 and #2 appear to be re-flags of findings already addressed in the parent commit (9f6e0dc) — Copilot's analysis may have keyed off the original PR-level context rather than the latest diff. Verified the current code on `main`-target branch already has the fixes (`func.lower(...)` at `utils/google_user.py:71`, `_enforce_provider_field_match` at `admin_signup_gate.py:131`).

JFK:
**Loop 3: addressed 4 of 9 Copilot comments on commit 290c076 (the critical one was a real production bug).**

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | `ilike()` wildcard injection (re-flag) | **Stale** — already fixed in commit 9f6e0dc (uses `func.lower(...)`). |
| 2 | `AllowlistAddRequest` mismatched-field acceptance (re-flag) | **Stale** — `_enforce_provider_field_match` already added. |
| 3 | `_OAUTH_CALLBACK_ACTOR` underscore import | **Deferred** — pre-existing codebase convention. |
| 4 | Double-commit on shared session | **Documented in loop 2** — see service docstring. |
| 5 | `func.lower(...)` no functional index | **Deferred** — admin-only path, bounded N, separate perf issue. |
| 6 | **CRITICAL — Alembic revision ID `e14_655_signup_allowlist_provider` is 33 chars (`version_num` is `VARCHAR(32)`)** | **Fixed**. Renamed to `e14_655_allowlist_provider` (26 chars). On asyncpg this would have raised `StringDataRightTruncationError` on `alembic upgrade`; some backends silently truncate, leaving the DB in an unknown revision state. Updated migration `revision = ...` and integration test reference. The migration docstring now records the old ID as a forensic breadcrumb. Excellent catch — this would have broken the very first production deploy. |
| 7 | Step 4.5 should also re-coalesce `provider` | **Docs only**. Added a code comment explaining why `provider` is intentionally absent from the WHERE predicate: its `server_default` ('github') persists permanently (unlike `subject_id`/`subject_label` whose defaults drop at Step 2.5), so the only path to `provider IS NULL` requires a writer that knows about the column — which would also explicitly set `subject_id`. |
| 8 | Sentinel `<provider>:<subject>` collision risk on downgrade | **Docs only**. Expanded `add_to_allowlist_entry` docstring to document why the colon prefix cannot collide with real GitHub numeric IDs (no colons in real IDs) and why the prefix is also useful for human readability. |
| 9 | Test docstring drift ("EmailStr") | **Fixed**. Updated the `test_422_on_malformed_google_email` docstring to describe the regex pattern strategy (we deliberately don't depend on `email-validator`). |

`make lint` clean. 39/39 unit tests pass locally.

Issue #487, opened 2026-04-29
fix(db): c03_471_seed_pricing migration TIMESTAMP type mismatch (asyncpg VARCHAR bind)


JFK:
## Symptom

`./scripts/deploy.sh` (blue-green) on production VM fails during `alembic upgrade head` on the green container with:

\`\`\`
sqlalchemy.exc.ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.DatatypeMismatchError'>: column "effective_from" is of type timestamp without time zone but expression is of type character varying
HINT:  You will need to rewrite or cast the expression.
[SQL: INSERT INTO llm_pricing (..., effective_from, ...) VALUES (..., \$4::VARCHAR, ...)]
[parameters: ('anthropic', 'claude-opus-4-7', 'input_tokens', '2026-04-28 00:00:00', '5')]
\`\`\`

## Root cause

`backend/alembic/versions/c03_471_seed_pricing.py:64`:

\`\`\`python
_SEED_EFFECTIVE_FROM = "2026-04-28 00:00:00"   # str
\`\`\`

asyncpg infers bind types from Python value runtime types. A `str` value becomes `\$4::VARCHAR` in the rendered SQL, which PostgreSQL refuses to coerce into `TIMESTAMP WITHOUT TIME ZONE` without an explicit cast.

The migration's docstring even claims "Naive UTC to match \`llm_pricing.effective_from\` (DateTime, naive)" — the intent was a `datetime`, not a string.

The bug also affects `downgrade()` at line 203-205 (the targeted DELETE) — the same `:effective_from` bind would fail.

## Fix

Use a `datetime` object so asyncpg infers `TIMESTAMP`:

\`\`\`python
from datetime import datetime
_SEED_EFFECTIVE_FROM = datetime(2026, 4, 28, 0, 0, 0)
\`\`\`

## Why CI didn't catch it

Need to verify whether CI runs alembic migrations against a fresh test DB. If \`make test-integration\` runs migrations, this would have failed there too — the bug class is "first-fresh-DB-only" since the migration only fails on INSERT (the schema migration c02 is fine).

If CI doesn't currently exercise migrations against a fresh DB, that's a follow-up.

## Production impact

**Production unaffected** — \`deploy.sh\` aborted before Caddy switch. \`kagura-api-blue\` (old container) still serves traffic, DB still on \`b06_406_embedding_calibration\`.

[...]

PR #703, opened 2026-05-17
refactor(db): pin postgres TZ to UTC + add naive-datetime lint guard (#490)

### LINKED ISSUE #490: feat(db): complete DateTime(timezone=True) migration for ~85 naive columns (Option B from #489)

JFK:
Closes #490

## Summary

- Pin postgres session timezone to UTC at three independent layers (container env, SQLAlchemy engine, Python writes) so the DB-stores-UTC invariant no longer depends on the postgres:15-alpine image default.
- Add `ruff DTZ` (flake8-datetimez) to block bare `datetime.utcnow()` / `datetime.now()` without `tz` / naive `datetime(YYYY,...)` literals in `backend/src/`, preventing the third-layer guarantee from regressing as new code lands.
- Add layer-2 regression-guard integration tests (async + sync engine) so future refactors of `db.base.create_*_engine` can't silently drop the connect_args.

## Scope pivot from the issue body

The original #490 scoped a full `mapped_column(DateTime, ...) -> (DateTime(timezone=True), ...)` migration across 113 columns + 188 `utcnow()` caller flip + alembic batches. After gate1 review (`/claude-c-suite:ask` -> CTO lens), an end-to-end audit confirmed that the "DB stores UTC + frontend converts" chain already works today:

| Layer | Implementation | State pre-PR |
|---|---|---|
| DB writes | `utils.datetime.utcnow()` (naive UTC) / `func.now()` | UTC by container default |
| Wire format Z-suffix | `TZAwareBaseModel` covers 100% of datetime-returning response schemas | working |
| Frontend parse | `formatDateTime` / `new Date("...Z")` | working |

Since predecessor #489 (closed 2026-04-29) already eliminated the user-visible JST shift bug, the column-type migration offered hygiene only - with significant production risk (asyncpg bind-type interop per #487, 188 `utcnow()` caller migration, 31 test fixture files). Operator chose to pivot to a 1-PR defense-in-depth alternative.

This PR delivers the equivalent correctness guarantees by **making the implicit UTC assumption explicit**, without touching column types. The "do not 'fix' the naive columns without re-litigating the trade-off" note in `.claude/rules/backend.md` records this decision durably so future contributors see it before re-opening the migration discussion.

## Implementation notes

1. **Container layer** - `TZ=UTC` + `PGTZ=UTC` added to the postgres service env in both `docker-compose.yml` (dev) and `terraform/single-server/docker-compose.prod.yml` (prod). Removes the silent dependency on the postgres:15-alpine image default.

2. **Engine layer** - `connect_args` added to both engines in `backend/src/db/base.py`:
   - async (asyncpg): `connect_args={"server_settings": {"timezone": "UTC"}}` - sent in the libpq startup packet, zero per-query overhead
   - sync (psycopg2, OAuth2 server only): `connect_args={"options": "-c timezone=utc"}` - psycopg2's libpq equivalent

3. **Python layer** - `ruff DTZ` enabled in `backend/pyproject.toml`. `backend/src/cli/setup_env.py:83` (`datetime.now()` without tz) and `backend/src/mcp_server/tools/context.py:553` (`datetime.min` sentinel) were the only existing src violations:
   - `setup_env.py`: switched to `datetime.now(UTC)` (intentionally keeps stdlib because this bootstrap script runs before `utils.*` is on `sys.path`)
   - `context.py`: added `# noqa: DTZ901` with a WHY comment explaining the sentinel is intentionally naive because `Memory.last_used_at` is `TIMESTAMP WITHOUT TIME ZONE`
   - `tests/**` exempted via `[tool.ruff.lint.per-file-ignores]` - 193 pre-existing fixture violations are deferred to a follow-up

4. **Test layer** - `backend/tests/integration/test_db_engine_timezone.py` adds two parallel regression guards (async + sync engines) that go through the production `_get_engine()` / `_get_sync_engine()` path and assert `SHOW timezone == 'UTC'`. The async test was added during self-review; the sync test was added after gate2 (QA Lead caught the asymmetry).

5. **Docs layer** - `.claude/rules/backend.md` Datetime/UTC section rewritten around the three-layer policy. The line previously saying "Until #490 migrates them" is replaced with the explicit statement that the migration was evaluated and deferred.

6. **utcnow() docstring** - updated in `backend/src/utils/datetime.py` to reference the three-layer policy and remove the "when the database is migrated..." comment that anticipated the deferred migration. **Behavior unchanged** (still returns naive UTC).

## Pre-PR review summary

- **gate2 mode**: advisor-only (`gate2.binary_gate=null` per v0.1.1 default)
- **audit**: skipped (advisor-only mode)
- **binary_gate**: (none)
- **cso**: green
- **qa-lead**: green (initial yellow on sync-engine asymmetry, closed in-flight by commit e54b5eda)
- **cto**: green
- **gate1**: green (scope-shrunk; original full-scope verdict was yellow - pivot via `/claude-c-suite:ask` -> CTO lens)
- **review provider**: code-review (per repo config)

- `~/.claude/cache/gh-issue-driven/490-refactor-explicit-utc-and-lint.gate1.md`
- `~/.claude/cache/gh-issue-driven/490-refactor-explicit-utc-and-lint.gate2.md`

## Verification

Live-verified against the running `kagura-postgres` container after recreating with the new env vars:

```
$ docker compose exec postgres sh -c 'echo TZ=$TZ; date'
TZ=UTC
Sun May 17 ... UTC 2026

$ docker compose exec postgres psql -U kagura -d kagura -c "SHOW timezone;"
 UTC

$ PYTHONPATH=src uv run python -c "...async session SHOW timezone..."
session timezone: UTC
now() is UTC: True
```

Plus 2 new integration tests (`test_async_engine_session_timezone_is_utc`, `test_sync_engine_session_timezone_is_utc`) - both pass.

## Test plan

- [x] `make lint` clean (ruff DTZ active, no src/ violations)
- [x] `tests/utils/` (307 pass) - `utcnow()` behavior unchanged
- [x] `tests/api/` (594 pass) - no regression
- [x] `tests/db/` (47 pass) - no regression
- [x] `tests/integration/test_db_engine_timezone.py` (2 pass) - new regression guards
- [x] `tests/integration/` baseline parity confirmed: 211 pass / 16 fail, identical to `main` (all 16 failures are pre-existing test-isolation flakes, unrelated to this diff)
- [ ] After merge: file follow-up for `tests/**` DTZ sweep (193 fixture violations deferred)

🤖 Generated via `/gh-issue-driven:ship`

### DIFF
--- a/.claude/rules/backend.md
+++ b/.claude/rules/backend.md
@@ -26,7 +26,7 @@ paths:
 
 ## Datetime / UTC
 
-**Storage (DB) — mixed-awareness, predominantly naive UTC.** Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); 11 columns across 5 models are `TIMESTAMP WITH TIME ZONE` (aware): `Context.last_used_at` (auth.py), and timestamps in `bm25_drift.py`, `erasure.py`, `hub_tag.py`, and `neural.py`. **Always check the model's `mapped_column(DateTime[, timezone=True])` declaration before constructing comparison sentinels or default values**, never assume naive across the board.
+**Storage (DB) — mixed-awareness, predominantly naive UTC.** Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); **13 columns across 6 models** are `TIMESTAMP WITH TIME ZONE` (aware): `Context.last_used_at` (auth.py), 2 timestamps in `config.py` (declared with `TIMESTAMP(timezone=True)` rather than `DateTime(timezone=True)`), and timestamps in `bm25_drift.py`, `erasure.py`, `hub_tag.py`, and `neural.py`. **Always check the model's `mapped_column(...)` declaration before constructing comparison sentinels or default values** — and grep both `DateTime(timezone=True)` and `TIMESTAMP(timezone=True)` since the codebase mixes both styles. Never assume naive across the board.
 
 UTC is enforced for every SQLAlchemy session by the engine, regardless of postgres server defaults:
 
--- a/.claude/rules/backend.md
+++ b/.claude/rules/backend.md
@@ -26,7 +26,7 @@ paths:
 
 ## Datetime / UTC
 
-**Storage (DB) — mixed-awareness, predominantly naive UTC.** Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); about a dozen columns are `TIMESTAMP WITH TIME ZONE` (aware) — for example `Context.last_used_at`, several `signup_gate.py` / erasure / neural timestamps. **Always check the model's `mapped_column(DateTime[, timezone=True])` declaration before constructing comparison sentinels or default values**, never assume naive across the board.
+**Storage (DB) — mixed-awareness, predominantly naive UTC.** Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); 11 columns across 5 models are `TIMESTAMP WITH TIME ZONE` (aware): `Context.last_used_at` (auth.py), and timestamps in `bm25_drift.py`, `erasure.py`, `hub_tag.py`, and `neural.py`. **Always check the model's `mapped_column(DateTime[, timezone=True])` declaration before constructing comparison sentinels or default values**, never assume naive across the board.
 
 UTC is enforced for every SQLAlchemy session by the engine, regardless of postgres server defaults:
 
--- a/.claude/rules/backend.md
+++ b/.claude/rules/backend.md
@@ -25,9 +25,19 @@ paths:
 - Pool config: `pool_size=5, max_overflow=10, pool_pre_ping=True`
 
 ## Datetime / UTC
-- DB columns are TIMESTAMP WITHOUT TIME ZONE — naive UTC by convention. Until #490 migrates them, the API layer is the only place ensuring wire-format datetimes are unambiguous (#489).
-- Response Pydantic schemas with any `datetime` field MUST inherit from `models.api_base.TZAwareBaseModel` (not `BaseModel`). The base class serializes naive datetimes with a `Z` suffix; without it, JS clients parse the string as local time and JST users see times shifted by their UTC offset.
-- For `str`-typed datetime fields in handler/service code (manual ISO formatting), use `to_utc_iso(dt)` from `utils.datetime`. Never call `.isoformat()` directly on a `datetime` field that ends up in a JSON response or cached payload — `to_utc_iso` is idempotent (handles None / naive / aware UTC / aware non-UTC) so it can be applied unconditionally.
+
+**Storage (DB)** — columns are `TIMESTAMP WITHOUT TIME ZONE`, holding naive UTC. UTC is enforced explicitly at three layers, so a naive value here is unambiguous:
+1. Postgres container env: `TZ=UTC` + `PGTZ=UTC` (`docker-compose.yml`, `terraform/single-server/docker-compose.prod.yml`)
+2. SQLAlchemy engine: async `connect_args={"server_settings": {"timezone": "UTC"}}` / sync `connect_args={"options": "-c timezone=utc"}` (`db/base.py`)
+3. Python writes via `utils.datetime.utcnow()` — never bare `datetime.utcnow()` or `datetime.now()` without `tz`
+
+`ruff DTZ` blocks naive-datetime patterns in `backend/src/`. Tests are exempt via `[tool.ruff.lint.per-file-ignores]` while the fixture sweep remains a follow-up.
+
+**Wire format (API JSON)** — response Pydantic schemas with any `datetime` field MUST inherit from `models.api_base.TZAwareBaseModel` (not `BaseModel`). The base class serializes naive datetimes with a `Z` suffix; without it, JS clients parse the string as local time and JST users see times shifted by their UTC offset.
+
+For `str`-typed datetime fields in handler/service code (manual ISO formatting), use `to_utc_iso(dt)` from `utils.datetime`. Never call `.isoformat()` directly on a `datetime` field that ends up in a JSON response or cached payload — `to_utc_iso` is idempotent (handles None / naive / aware UTC / aware non-UTC) so it can be applied unconditionally.
+
+**Column-type migration was evaluated in #490 and explicitly deferred.** The full `Column(DateTime, ...) → Column(DateTime(timezone=True), ...)` sweep was scoped at 113 columns + 188 caller sites and judged hygiene-only after #489 closed the user-visible bug. The three layers above provide equivalent correctness guarantees; do not "fix" the naive columns without re-litigating the trade-off.
 
 ## Testing
 - Test files: `backend/tests/{module}/test_{name}.py`
--- a/.claude/rules/backend.md
+++ b/.claude/rules/backend.md
@@ -26,18 +26,21 @@ paths:
 
 ## Datetime / UTC
 
-**Storage (DB)** — columns are `TIMESTAMP WITHOUT TIME ZONE`, holding naive UTC. UTC is enforced explicitly at three layers, so a naive value here is unambiguous:
-1. Postgres container env: `TZ=UTC` + `PGTZ=UTC` (`docker-compose.yml`, `terraform/single-server/docker-compose.prod.yml`)
-2. SQLAlchemy engine: async `connect_args={"server_settings": {"timezone": "UTC"}}` / sync `connect_args={"options": "-c timezone=utc"}` (`db/base.py`)
-3. Python writes via `utils.datetime.utcnow()` — never bare `datetime.utcnow()` or `datetime.now()` without `tz`
+**Storage (DB) — mixed-awareness, predominantly naive UTC.** Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); about a dozen columns are `TIMESTAMP WITH TIME ZONE` (aware) — for example `Context.last_used_at`, several `signup_gate.py` / erasure / neural timestamps. **Always check the model's `mapped_column(DateTime[, timezone=True])` declaration before constructing comparison sentinels or default values**, never assume naive across the board.
 
-`ruff DTZ` blocks naive-datetime patterns in `backend/src/`. Tests are exempt via `[tool.ruff.lint.per-file-ignores]` while the fixture sweep remains a follow-up.
+UTC is enforced for every SQLAlchemy session by the engine, regardless of postgres server defaults:
+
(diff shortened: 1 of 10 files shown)


JFK:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

JFK:
### Copilot review response — all 3 findings addressed

Thank you @copilot — all three comments were valid. Commit 0a324a01 addresses each:

**1. `context.py:554` — real TypeError bug uncovered (CRITICAL)**

You were right that `Context.last_used_at` is `DateTime(timezone=True)`. My previous noqa comment misidentified the field as `Memory.last_used_at`. Confirmed the bug:

```python
>>> from datetime import datetime, UTC
>>> sorted([datetime.min, datetime.now(UTC)])
TypeError: can't compare offset-naive and offset-aware datetimes
```

Fixed by switching the sentinel to `datetime.min.replace(tzinfo=UTC)` (aware), removing the `# noqa: DTZ901` since the suppression is no longer needed. This bug pre-existed PR #703 but was masked by the rarity of contexts with mixed `None`/non-`None` values — my noqa with the wrong rationale made it harder to spot. Good catch.

**2. `backend.md:40` — storage rule too broad**

Correct — `Context.last_used_at` plus several `signup_gate.py` / erasure / neural timestamps are already `DateTime(timezone=True)`. Rewrote the storage section to flag the mixed-awareness reality:

> Most columns are `TIMESTAMP WITHOUT TIME ZONE` (naive UTC by convention); about a dozen columns are `TIMESTAMP WITH TIME ZONE` (aware) ... Always check the model's `mapped_column(DateTime[, timezone=True])` declaration before constructing comparison sentinels or default values, never assume naive across the board.

Plus an explicit note that aware-column sentinels must use `datetime.min.replace(tzinfo=UTC)` — addresses the same class of bug as #1.

**3. `docker-compose.prod.yml:70` — TZ/PGTZ over-promised**

Correct that `PGTZ` only affects libpq clients (`psql`) and that neither env var rewrites the postgres server's `timezone` GUC in an already-initialized `postgres_data` volume (that's set at `initdb`). Reframed the three-layer narrative in backend.md so the engine `connect_args` is described as the **primary** guarantee for application sessions, and the container env layer is OS-level alignment (cron, logs, `psql`) rather than a server-default rewrite. The compose changes still pull their weight for the OS / interactive-session story, but the doc no longer overstates what they accomplish for postgres itself.

Diff: +18 / -11 in 2 files (`.claude/rules/backend.md` + `backend/src/mcp_server/tools/context.py`).

🤖 Generated via `/gh-issue-driven:review` loop

JFK:
### Copilot review response (round 2) — both findings addressed

Commit d0d6418e addresses both:

**1. `test_db_engine_timezone.py:34` — false-negative risk closed**

You were right that the original tests only ran `SHOW timezone` against a real session, and because the container is now pinned to `TZ=UTC` (and the postgres image default is also UTC), `SHOW timezone` would still return `UTC` even if `connect_args` were dropped from `_get_engine`. That defeated the test's stated purpose as a layer-2 regression guard.

Refactored each test to spy on `create_async_engine` / `create_engine` via `unittest.mock.patch` while the lazy getter rebuilds the engine, then assert the exact `connect_args` shape:

```python
with patch.object(db_base, "create_async_engine", wraps=db_base.create_async_engine) as spy:
    engine = db_base._get_engine()
    kwargs = spy.call_args.kwargs
    assert kwargs["connect_args"]["server_settings"]["timezone"] == "UTC"
```

If anyone removes `connect_args` from `_get_engine`, the spy assertion fires immediately — independent of postgres server defaults or container env. The end-to-end `SHOW timezone` check is retained per engine as a secondary signal (catches a malformed `connect_args` dict where the spy alone would not). Two complementary checks per engine, both pass locally.

**2. `backend.md:29` — aware-column example list corrected**

Confirmed via `grep -rn 'DateTime(timezone=True)' backend/src/models/`:

| File | Aware columns |
|---|---|
| `auth.py` | `Context.last_used_at` (1) |
| `bm25_drift.py` | (1) |
| `erasure.py` | `requested_at`, `confirmed_at`, `scheduled_for`, `started_at`, `completed_at`, `cancelled_at` (6) |
| `hub_tag.py` | (1) |
| `neural.py` | `created_at`, `valid_until` (2) |

Total: **11 columns across 5 files**. `signup_gate.py` is 100% naive — removed from the example list, replaced with the verified set above.

Diff: +79/-12 across 2 files.

🤖 Generated via `/gh-issue-driven:review` loop

JFK:
### Copilot review response (round 3) — finding addressed

Commit 0f525894 addresses the one finding:

**`test_db_engine_timezone.py:54` — session_factory cache also cleared**

You were right. `_reset_cached_engines` disposed the `engine` / `sync_engine` and cleared those globals, but `async_session_factory` / `sync_session_factory` are cached separately and would have stayed bound to the just-disposed engine. Any subsequent test calling `get_db()` or `get_sync_session()` would have received a stale, disposed pool — exactly the order-dependent failure mode the integration `conftest._reset_db_base_state` already guards against.

The helper now mirrors that pattern (engines disposed, both engine globals AND both factory globals set to `None`):

```python
for factory_name in ("async_session_factory", "sync_session_factory"):
    if hasattr(db_base, factory_name):
        setattr(db_base, factory_name, None)
```

Verified by cross-test run with `test_oauth_dcr_integration.TestDcrLoopbackPersistsNullOwnerId` (which uses real session factories) — 3 passed, no stale-pool failure.

Diff: +9 / -0 in 1 file.

🤖 Generated via `/gh-issue-driven:review` loop

JFK:
### Copilot review response (round 4) — 1 of 2 findings addressed

**`backend.md:29` — aware-column inventory still stale (ACCEPTED, fixed in 0aa61f1b)**

Correct catch. My earlier `grep -rn 'DateTime(timezone=True)'` missed `backend/src/models/config.py:103,106`, which uses the alternate `TIMESTAMP(timezone=True)` spelling. Combined with the previously-listed 11 `DateTime(timezone=True)` columns, the total is **13 aware columns across 6 models**. Updated the inventory and added a note that contributors must grep both spellings since the codebase mixes them.

**`context.py:555` — declined as false positive (with evidence)**

The claim is that `datetime.min.replace(tzinfo=UTC)` will still trigger DTZ901 because the value is only made aware after the `datetime.min` access. Verified this against the actual ruff implementation:

```
$ uv run ruff check --no-cache --select DTZ901 src/mcp_server/tools/context.py
All checks passed!

$ cat /tmp/repro.py
from datetime import UTC, datetime
x = datetime.min.replace(tzinfo=UTC)

$ uv run ruff check --no-cache --select DTZ901 /tmp/repro.py
All checks passed!
```

Ruff's DTZ901 AST checker recognizes the chained `.replace(tzinfo=...)` immediately following `datetime.min` / `datetime.max` and correctly does not fire. The full DTZ rule set on `src/` is clean (no `noqa` needed), and CI lint passed on the previous commit.

Not adding a `# noqa: DTZ901` since the lint genuinely does not trigger and an unnecessary suppression would be misleading documentation.

Diff: +1 / -1.

🤖 Generated via `/gh-issue-driven:review` loop

JFK:
### Copilot review response (round 5) — finding addressed

Commit 2435ab8f addresses the one finding:

**`db/base.py:50` (effective scope `backend/alembic/env.py`) — alembic engine pinned to UTC**

You're right. `backend/alembic/env.py:101` creates its own async engine via `async_engine_from_config` and was bypassing the layer-2 guarantee. Migrations relying on `server_default=func.now()` for timestamp columns (e.g., `signup_gate_config.created_at`) would have inherited the postgres server default during `alembic upgrade`, not the explicit UTC pin from `db.base._get_engine`.

This depended only on layer 1 (container TZ/PGTZ) — and as you correctly flagged in round 1, container env vars do NOT rewrite the postgres server default in an already-initialized `postgres_data` volume. So if a deploy ever dropped TZ/PGTZ before this commit, fresh migrations could have silently written non-UTC timestamps. Real gap.

Fix: pass the same `connect_args={"server_settings": {"timezone": "UTC"}}` to `async_engine_from_config`. Verified end-to-end:

```
$ alembic engine session tz: UTC
$ make test-integration (test_db_engine_timezone + test_alembic_migrations) → 14 passed
```

This brings layer 2's coverage from "application sessions only" to "application + migration sessions" — closing the round-1 finding's tail risk.

Diff: +6 / -0 in 1 file.

🤖 Generated via `/gh-issue-driven:review` loop

PR #711, opened 2026-05-18
feat(api,services): per-workspace embedding spend cap (#709)

JFK:
Closes #709

## Summary
- Adds per-workspace BYOK embedding spend cap (USD, daily + monthly) — hard prereq for #708 Option A (shared-context reads charge the shared-source workspace owner's API key)
- Mitigates the Embedding Drain Attack threat: a hostile shared context can no longer burn the owner's API budget unbounded
- Schema: `Workspace.embedding_*_cap_usd` NUMERIC(10,6) NULL columns + CHECK >= 0 (migration `e16_709`). NULL = inherit-tier-default; positive = admin override. Tier defaults: Free $0.50/$15, Basic $2/$60, Pro $10/$300 (env-overridable)
- Service: `EmbeddingSpendCapService` (~400 LOC) — Redis micro-USD counter + cap check + 80%/100% Resend alert with SETNX dedup
- Enforcement chokepoint: `EmbeddingService.embed_with_usage` + `embed_batch` via new `_prepare_spend_cap_gate` helper (covers search_service + 3 Sleep + embed_batch sites)
- Admin endpoint: `PUT /api/v1/admin/plans/workspaces/{id}/spend-cap` with tier-bounded validation

## Implementation notes
- **Override semantics** (not addon-style): admin-set value REPLACES tier default rather than stacking. Distinct from `addon_sleep_contexts_bonus` (#560) which is additive.
- **Email is `await`-ed inline** (not `asyncio.create_task`) — avoids GC orphan + dangling AsyncSession risk identified in `/simplify` Agent 1 H2 review.
- **`paid_by` writer fix**: `search_service.py:208` no longer hardcodes `"platform"`. New `EmbeddingService.resolve_paid_by` centralizes byok/platform resolution via `ExternalAPIKey` presence check.
- **Period type-narrowed**: `EmbeddingSpendCapPeriod = Literal["daily", "monthly"]` + `EMBEDDING_SPEND_CAP_PERIODS` tuple constant (mirrors `LLM_CALL_LOG_PAID_BY_VALUES` pattern).
- **HYBRID architecture**: Phase 4 cherry-pick from 3 architects — chokepoint enforcement (Minimal) + Redis precision (Clean) + member_credentials paid_by fix (Pragmatic).

## Tests (35 new, all pass locally)
- 18 unit tests — `EmbeddingSpendCapService` (cap check, record, alert dedup, fail-open Redis, key format, TTL)
[...]

PR #524, opened 2026-05-01
feat(db): cost-grade source + paid_by dimensions on sleep_reports (#523)

JFK:
Closes #523

## Summary
- Adds `source` (`sleep`/`analysis`) and `paid_by` (`platform`/`byok`) NOT NULL columns to `sleep_reports` so #472 cost aggregation can split by run type and billing party, and #495 broadlistening pipeline can emit cost rows in the same persist transaction.
- Migration uses `ADD CONSTRAINT NOT VALID` + `VALIDATE CONSTRAINT` (mirrors the `b03_396` precedent) so the CHECK validation runs under SHARE UPDATE EXCLUSIVE rather than ACCESS EXCLUSIVE on populated tables.
- New composite index `(workspace_id, source, started_at DESC)` covers the anticipated #472 query shape; reporter constructor explicitly sets the new columns to mirror defaults and document the `#495` divergence point.

## Implementation notes
- ORM (`backend/src/models/sleep.py`): both columns carry Python `default=` AND `server_default=text("'X'")`. Module-level `SLEEP_REPORT_SOURCES` and `SLEEP_REPORT_PAID_BY_VALUES` tuples mirror the `LLM_PRICING_UNIT_TYPES` pattern for service-layer validation; the DB CHECK is the source of truth.
- Migration (`backend/alembic/versions/d05_523_costgrade_source_paid_by.py`): revision `d05_523_source_paid_by` (22 chars, ≤32 cap), down_revision `d04_519_oauth_owner_nullable`. PG 15 fast-path: NOT NULL + constant DEFAULT is metadata-only, no table rewrite.
- Tests (`backend/tests/integration/test_sleep_report_schema.py`): 5 integration tests cover CHECK rejection × 2, ORM in-memory default population, server_default DDL clause via `information_schema.columns`, and BYOK round-trip.
- Verified manually: `alembic upgrade head` → `downgrade -1` → `upgrade head` round-trip clean on `kagura` DB; resulting schema has `idx_sleep_reports_workspace_source_started btree (workspace_id, source, started_at DESC)` plus both CHECK constraints.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped (binary_gate=null)
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: not run (ship-only mode — issue had no /start state)
- review provider: copilot

Full reviews are saved in the plugin cache:
[...]

Issue #81, opened 2026-04-01
feat(search): cross-context recall


JFK:
## Summary

Allow recall queries to search across multiple contexts or all user-owned contexts at once.

## Motivation

Currently `context_id` is required for recall — you can only search one context at a time. Users with multiple contexts (per-project, per-topic) can't do a unified search.

Use cases:
- "Search everything I know about authentication" across all projects
- A central agent that draws from multiple knowledge domains
- Deduplication detection across contexts

## Proposed Design

Option A: Multi-context recall
```json
{
  "query": "authentication flow",
  "context_ids": ["ctx-1", "ctx-2", "ctx-3"],
  "k": 10
}
```

Option B: Global recall (all owned contexts)
```json
{
  "query": "authentication flow",
  "context_id": "*",
  "k": 10
}
```

Response includes `context_id` per result so the caller knows the source:
```json
{
  "memories": [
    {"memory_id": "mem-1", "context_id": "ctx-1", "summary": "..."},
    {"memory_id": "mem-2", "context_id": "ctx-3", "summary": "..."}
  ]
}
```

## Considerations

- Performance: searching multiple vector indexes vs one unified index
- Ranking: how to merge scores across contexts with different embedding models?
- Same-model contexts could be searched together; mixed models need separate searches + score normalization
- Permission: only search contexts the user owns or has access to
- Public contexts: should they be included in global search?

## SDK Impact

- `KaguraClient.recall()` accepts `context_ids` list or `"*"`
- Useful for future `KaguraAgent` auto-context selection (currently "auto" picks one)

PR #715, opened 2026-05-18
fix(api): shared-context recall pays source workspace (#708)

### LINKED ISSUE #708: Shared-context MCP reads fail when caller workspace has no API key
## Summary

When an admin (or any user who is not the workspace owner) reads a shared context via MCP (`recall`, `explore`, `analyze_context`, `reference`), the embedding-key lookup is keyed on the **caller's** `User.current_workspace_id` rather than the **context's** `workspace_id`. If the caller's own workspace has no `ExternalAPIKey` configured, the request fails with:

```
OpenAI API key not configured for workspace {workspace_id}.
Configure a workspace OpenAI API key in settings...
```

— even when the shared-source workspace owner has configured a key.

## Context

**Priority**: high
**Suggested labels**: bug, backlog
**Suggested milestone**: (none — blocked on `paid_by` policy decision; anchor after policy resolution)

## Root cause

Embedding key lookup chain:

1. `backend/src/mcp_server/transport.py:420` — `_get_user_workspace_id(user_id)` returns `User.current_workspace_id` (caller's active workspace).
2. `backend/src/mcp_server/tools/memory.py:288` — `recall` handler passes that `workspace_id` straight through to `MemoryService.recall`.
3. `backend/src/services/memory_service.py:1221` — passes `workspace_id=str(current_workspace_id)` to `SearchService.hybrid_search`.
4. `backend/src/services/embedding_service.py:114-164` — `_get_user_api_key` filters `ExternalAPIKey` by that `workspace_id` and raises `ConfigurationError` when no row is found.

`Context.workspace_id` (the shared-source workspace, `backend/src/models/auth.py:1199-1203`) is never consulted, even though the model holds it.

## Acceptance criteria

- [ ] Decide `paid_by` policy first (see "Open policy question" below). Block implementation on this.
- [ ] When `caller_workspace_id != context.workspace_id` (shared-context read), embedding/LLM key lookup uses `context.workspace_id`.
- [ ] When `caller_workspace_id == context.workspace_id` (self-workspace read), behavior is unchanged.
- [ ] Error message in `embedding_service.py` is reworded to point operators at the correct owner when the lookup fails (the current message tells the admin to set their own key, which is misleading after this change).
- [ ] Unit + integration tests cover both branches (self vs shared) and the failure path (shared-source workspace also has no key).
- [ ] `paid_by` / cost-grade source attribution reflects the policy decision (e.g. shared-source workspace owner is the `paid_by` party when their key is used).

## Open policy question (must decide before implementation)

Who absorbs embedding/LLM cost when a non-owner reads a shared context?

- **Option A — shared-source workspace owner pays**: natural for "open the door with the home owner's key". Risk: hostile shared context could be used to burn the owner's API budget.
- **Option B — caller's workspace pays**: requires caller-side key. Equivalent to today's behavior — does not solve the reported failure.
- **Option C — explicit per-context billing setting**: most flexible, more UI/DB surface.

This must align with the v0.15.0 cost-grade `source` / `paid_by` mechanism (PR #524). Implementing the fallback without resolving this risks a double-implementation conflict with cost-grade.

## Suggested implementation direction (after policy decision)

**B-1 (recommended)**: resolve the target `Context` first, then if the caller's workspace differs from `context.workspace_id`, override `workspace_id` to `context.workspace_id` for downstream embedding/LLM resolution. Self-workspace reads retain the current path. Single-point change in the MCP handler / transport layer.

**B-2**: always use `context.workspace_id` for key lookup, regardless of caller. Simpler, but loses the symmetry where each user's own context reads use their own configured keys.

## Touch points (B-1)

- `backend/src/mcp_server/transport.py:420` — extend `_get_user_workspace_id` (or its callers) to accept a context hint and prefer `context.workspace_id` when set.
- `backend/src/mcp_server/tools/memory.py:288` — after `_resolve_context`, override `current_workspace_id` for the embedding/LLM path.
- `backend/src/services/embedding_service.py:114-164` — reword the `ConfigurationError` message to "ask the workspace owner to configure an API key" so admins of caller-side workspaces are not misled.

## Related

- #146 — Workspace-scoped API keys support (merged; established the keying scheme this issue extends).
- #165 — Privacy control for shared contexts (sibling concern; coordinate scope).
- PR #524 — cost-grade source/paid_by (v0.15.0; the mechanism `paid_by` policy must integrate with).

## Reporter context

Reported via Claude Desktop MCP: an admin (not workspace owner) calling a shared-context recall received the "OpenAI API key not configured" prompt despite never having had OpenAI configured personally.

JFK:
Closes #708

## Summary
- Implements Option A billing for shared-context MCP reads: when the caller workspace differs from the context's owner workspace, embedding cost (BYOK key + spend cap + paid_by + Qdrant collection + graph mutations) routes to the source workspace.
- Service-layer aggregation in `MemoryService.recall`: new `context_workspace_id` parameter, derives `effective_workspace_id`, threads through 4 downstream call sites.
- H1 BYOK-required deny (OWASP A04): source workspace without BYOK → `NotFoundException`. Closes the platform-key fallback uncapped path that PR #711's BYOK-only spend cap intentionally does not cover.
- H2 uniform error (OWASP A01 / CWE-639): handler re-casts `NotFoundException` to `_ContextNotFoundError`. Callers cannot distinguish "source missing BYOK" from "context does not exist for you" via the response body.
- H3 cross-context invariant: `context_ids[]` enforced same-workspace at API boundary (early-break before further permission SELECTs).
- Bundle: migrates `_resolve_context` → `_resolve_context_for_read` for read paths (recall + reference) to unify CWE-639 disclosure across all deny reasons (paid down issue #614 anti-pattern surface).

## Implementation notes
- **Rate-limit invariant preserved**: `mcp_server.tools.__init__._check_rate_limit` keeps the caller workspace as the rate-limit subject. The override scope is bounded to embedding-cost surfaces only, so attackers cannot bypass their own RPS via shared-context reads.
- **Defense-in-depth**: 3-layer model (upstream rate-limit → service H1 BYOK probe → downstream `EmbeddingSpendCapService` BYOK-only cap from PR #711).
- **Server-side audit log**: `paid_by_workspace_id` + `caller_user_id` + `reason="missing_byok"` for admin diagnosis without leaking to client.
- **Behavioral change**: `_resolve_context_for_read` migration broadens the deny-reason range surfaced as `context_not_found` (private non-creator, not-a-member, etc. now uniformly disclosed). Intentional.
- **Defensive bind**: `current_context_id: UUID | None = None` at top of handler `try` block with unreachable-branch fallback in the `NotFoundException` handler — closes a static-analysis edge case the test suite cannot exercise.

## Out of scope (intentional)
- `service.forget()` internal `self.recall(...)` at memory_service.py:1582 — pre-existing missing `current_workspace_id` (fails validation regardless of #708). Follow-up worth filing.
- REST `/api/v1/memory/recall` route — stub since #246 (passes `current_context_id=None`).
- `default_embedding_model` heterogeneity (gate1 R1) — codebase uses single global model from settings; column not yet added. Forward-looking hook: `effective_workspace_id` is the natural insertion point.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: yellow (handler-layer integration test gap — acknowledged, follow-up issue worth filing)
- cto: green
- gate1: green via /ask → CTO routing + follow-up /cso

- ~/.claude/cache/gh-issue-driven/708-fix-shared-context-mcp-reads-fail-when-calle.gate1.md
- ~/.claude/cache/gh-issue-driven/708-fix-shared-context-mcp-reads-fail-when-calle.gate2.md

## Tests
- 7 new service-layer unit tests covering self-workspace regression, shared-context routing, BYOK probe targeting, no-BYOK deny, uniform-error workspace-leak fence, and self-workspace BYOK-probe skip optimization.
- 230 existing tests pass (memory_service, mcp_server, recall_analysis_cluster_filter, embedding_service, embedding_spend_cap_service).

🤖 Generated via `/gh-issue-driven:ship`

### DIFF
--- a/backend/src/mcp_server/tools/memory.py
+++ b/backend/src/mcp_server/tools/memory.py
@@ -263,6 +263,17 @@ async def handle_recall(
                 # Option A paid_by routing has a single source-of-truth
                 # workspace. Check inline so a mismatch short-circuits
                 # before paying further permission lookups.
+                #
+                # #708 loop 6 (Copilot): also reject MIXED privacy across
+                # the list. ``SearchService`` derives a single
+                # ``is_shared_context`` value from the primary and applies
+                # it to every context's Qdrant filter. If primary is
+                # shared and a secondary is private (which the handler
+                # permits when the caller is that private context's
+                # creator / ContextMember), dropping the ``user_id`` filter
+                # would leak memories authored by other users in the
+                # private secondary. Rejecting at API boundary mirrors
+                # the same-embedding-model invariant pattern below.
                 for cid in cross_context_ids[1:]:
                     cross_ctx = await _resolve_context_for_read(db, user_id, cid)
                     if cross_ctx.workspace_id != current_context.workspace_id:
@@ -270,6 +281,12 @@ async def handle_recall(
                             "workspace_mismatch",
                             "All contexts in cross-context recall must belong to "
                             "the same workspace.",
+                        )
+                    if cross_ctx.is_private != current_context.is_private:
+                        return _error_response(
+                            "context_privacy_mismatch",
+                            "All contexts in cross-context recall must share the "
+                            "same privacy setting (all shared or all private).",
                         )
 
                 # Validate all contexts use the same embedding model
--- a/backend/src/mcp_server/tools/memory.py
+++ b/backend/src/mcp_server/tools/memory.py
@@ -18,10 +18,12 @@ from mcp_server.tools._helpers import (
     _error_response,
     _log_tool_usage,
     _resolve_context,
+    _resolve_context_for_read,
     _resolve_context_id,
     _validate_memory_id,
     execute_with_timeout,
 )
+from utils.exceptions import NotFoundException
 
 logger = logging.getLogger(__name__)
 
@@ -241,19 +243,34 @@ async def handle_recall(
 
     start_time = time.time()
     async for db in get_db():
+        # Bind defensively so the NotFoundException handler below can
+        # always reference a concrete context_id. In normal flow this is
+        # re-bound at lines below before any path that raises.
+        current_context_id: UUID | None = None
         try:
             # Issue #81: Cross-context recall — context_ids overrides context_id
             context_ids_arg = args.get("context_ids")
             cross_context_ids: list[UUID] | None = None
 
             if isinstance(context_ids_arg, list) and context_ids_arg:
-                # Multi-context mode
+                # Multi-context mode. Use _resolve_context_for_read so deny
+                # reasons (private non-creator, not a workspace member, etc.)
+                # surface as a uniform context_not_found (CWE-639 / OWASP A01).
                 cross_context_ids = [_resolve_context_id(cid) for cid in context_ids_arg]
                 current_context_id = cross_context_ids[0]
-                # Validate access to all contexts, keep primary context object
-                current_context = await _resolve_context(db, user_id, current_context_id)
+                current_context = await _resolve_context_for_read(db, user_id, current_context_id)
+                # #708 H3: all contexts must belong to the same workspace —
+                # Option A paid_by routing has a single source-of-truth
+                # workspace. Check inline so a mismatch short-circuits
+                # before paying further permission lookups.
                 for cid in cross_context_ids[1:]:
-                    await _resolve_context(db, user_id, cid)
+                    cross_ctx = await _resolve_context_for_read(db, user_id, cid)
+                    if cross_ctx.workspace_id != current_context.workspace_id:
+                        return _error_response(
+                            "workspace_mismatch",
+                            "All contexts in cross-context recall must belong to "
+                            "the same workspace.",
+                        )
 
                 # Validate all contexts use the same embedding model
                 from repositories.config_repository import ContextSearchConfigRepository
@@ -277,7 +294,7 @@ async def handle_recall(
                         "Missing required field: context_id (or provide context_ids with 2+ UUIDs).",
                     )
                 current_context_id = _resolve_context_id(args["context_id"])
-                current_context = await _resolve_context(db, user_id, current_context_id)
+                current_context = await _resolve_context_for_read(db, user_id, current_context_id)
 
             service = MemoryService(db)
             result = await execute_with_timeout(
@@ -286,6 +303,8 @@ async def handle_recall(
                     user_id=user_id,
                     current_context_id=current_context_id,
                     current_workspace_id=workspace_id,
+                    # #708 Option A: source workspace pays for shared-context reads.
+                    context_workspace_id=current_context.workspace_id,
                     context_ids=cross_context_ids,
                 ),
                 operation_name="recall",
@@ -338,6 +357,31 @@ async def handle_recall(
         except _ContextNotFoundError as e:
             await db.rollback()
             return e.to_response()
+        except NotFoundException:
+            # #708 H2: service raises NotFoundException("Context", ...) for
+            # shared-context reads that cannot proceed (e.g. source has no
+            # BYOK). Re-cast through _ContextNotFoundError so the response
+            # body is byte-identical to a regular deny — callers cannot
+            # distinguish "source missing BYOK" from "context does not
+            # exist for you" (CWE-639 / OWASP A01).
+            await db.rollback()
+            await _log_tool_usage(
+                db, user_id, "recall", start_time, 404, current_context_id, workspace_id
+            )
+            if current_context_id is None:
+                # Unreachable in practice — current_context_id is bound
+                # before any path that can raise NotFoundException — but
+                # static analysis cannot prove the invariant.
+                return _error_response(
+                    "context_not_found",
+                    "Context not found or you don't have access to it.",
+                    context_id=args.get("context_id"),
+                    help="Use list_contexts() to see contexts you have access to.",
+                )
+            return _ContextNotFoundError(
+                current_context_id,
+                "Context not found or you don't have access to it.",
+            ).to_response()
         except Exception:
             await db.rollback()
             await _log_tool_usage(
@@ -438,7 +482,10 @@ async def handle_reference(
     async for db in get_db():
         try:
             current_context_id = _resolve_context_id(args["context_id"])
-            await _resolve_context(db, user_id, current_context_id)
+            # Issue #708 bundle: migrate read paths to the CWE-639-uniform
+            # resolver so deny reasons (private non-creator, not a workspace
+            # member, etc.) cannot be distinguished by callers.
+            await _resolve_context_for_read(db, user_id, current_context_id)
 
             service = MemoryService(db)
             try:
@@ -446,16 +493,12 @@ async def handle_reference(
                     service.reference(request.memory_id, user_id=user_id),
                     operation_name="reference",
                 )
-            except Exception as e:
-                from utils.exceptions import NotFoundException
-
-                if isinstance(e, NotFoundException):
-                    return _error_response(
-                        "memory_not_found",
-                        f"Memory not found or you don't have access: {request.memory_id}",
-                        help="Use recall() to find memories you have access to.",
-                    )
-                raise
+            except NotFoundException:
+                return _error_response(
+                    "memory_not_found",
+                    f"Memory not found or you don't have access: {request.memory_id}",
+                    help="Use recall() to find memories you have access to.",
+                )
 
             reference_data = {
                 "memory_id": str(result.memory_id),
--- a/backend/src/services/embedding_service.py
+++ b/backend/src/services/embedding_service.py
@@ -84,6 +84,7 @@ class EmbeddingService:
         user_id: str,
         context_id: str | None = None,
         workspace_id: str | None = None,  # Issue #146
+        disallow_env_fallback: bool = False,
     ) -> str:
         """Resolve the OpenAI API key for the calling user's workspace context.
 
@@ -104,12 +105,19 @@ class EmbeddingService:
             context_id: Optional context UUID (#82: context-scoped keys take priority).
             workspace_id: Workspace UUID (#146); when omitted the DB lookup is skipped
                 and only the env-var fallback can satisfy the request.
+            disallow_env_fallback: Issue #708 loop 7. When True, skip the
+                ``OPENAI_API_KEY`` env-var fallback at priority 3 and raise
+                ``ConfigurationError`` instead. Set by the Option A shared-context
+                read path so a TOCTOU race between the preflight ``has_byok_key``
+                probe and this resolution cannot silently route to the uncapped
+                platform key. Has no effect when a DB key is found.
 
         Returns:
             Decrypted OpenAI API key.
 
         Raises:
-            ConfigurationError: If neither a DB key nor an env var is available.
+            ConfigurationError: If neither a DB key nor an env var is available
+                (or when ``disallow_env_fallback`` is True and no DB key was found).
         """
         from uuid import UUID
 
@@ -155,14 +163,27 @@ class EmbeddingService:
             )
             return api_key
 
-        # Fallback to environment variable (development / env-only deployments).
-        env_key = os.getenv("OPENAI_API_KEY")
-        if env_key:
-            logger.debug(
-                "openai_api_key_from_env",
+        # Fallback to environment variable (development / env-only deployments),
+        # gated by the Option A shared-context contract: when the caller has
+        # explicitly required a BYOK key (disallow_env_fallback=True), do NOT
+        # silently route to the platform key — that would defeat PR #711's
+        # BYOK-only spend cap and bypass the H1 preflight guard via TOCTOU.
+        if not disallow_env_fallback:
+            env_key = os.getenv("OPENAI_API_KEY")
+            if env_key:
+                logger.debug(
+                    "openai_api_key_from_env",
+                    user_id=user_id,
+                )
+                return env_key
+        else:
+            logger.warning(
+                "openai_api_key_env_fallback_disallowed",
                 user_id=user_id,
+                workspace_id=workspace_id,
+                context_id=context_id,
+                reason="option_a_shared_context_read",
             )
-            return env_key
 
         if workspace_id:
             raise ConfigurationError(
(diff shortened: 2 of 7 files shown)


JFK:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

JFK:
## Addressed all 3 findings — `3175c344`

Thanks @Copilot. All three findings were real and have been fixed.

### F1 (line 1184) — BYOK gate scoped to embed-cost paths

`search_mode="keyword"` skips `embed_with_usage` (BM25-only) and Ollama is free/local — neither charges the source workspace, so the H1 deny no longer fires for them. Scope is now:

```python
will_charge_embedding_cost = (
    request.search_mode != "keyword" and embed_svc.provider != "ollama"
)
if will_charge_embedding_cost and not await embed_svc.has_byok_key(...):
    raise NotFoundException("Context", ...)
```

Fenced by `test_shared_context_keyword_mode_skips_byok_gate` and `test_shared_context_ollama_provider_skips_byok_gate`.

### F2 (line 1269) — Cross-workspace shared reads broken end-to-end

You called out the list-context case, but I traced it further and the single-context case was also breaking under Option A: `SearchService.hybrid_search` at line 128 was running `is_workspace_member(workspace_id=source)`, which fails for any cross-workspace caller (system_admin, ContextMember-only) — including the original #708 reporter scenario.

Fix threads `is_shared_context_read` from `MemoryService.recall` (where the cross-workspace condition is unambiguously known) through to `SearchService.hybrid_search`. When set, the search service:
1. Skips the redundant `is_workspace_member` check — handler-layer `_resolve_context_for_read` is already authoritative
2. Forces `is_shared_context=True` so the Qdrant filter drops `user_id == caller` — applied uniformly to single-context AND list-context_id paths

This unblocks the headline use case end-to-end. Fenced by `test_shared_context_recall_propagates_is_shared_context_read` (positive) and `test_self_workspace_recall_does_not_propagate_is_shared_context_read` (negative).

### F3 (line 1182) — BYOK probe now mirrors `_get_user_api_key` priority

`has_byok_key(workspace_id, context_id=None)` now accepts an optional `context_id`. When supplied, the probe matches the same OR clause `_get_user_api_key` uses: `(context_id == ctx OR context_id IS NULL)`. A workspace whose only BYOK row is scoped to a different context no longer falsely passes the H1 gate.

The existing test `test_shared_context_recall_calls_has_byok_with_source_workspace_id` was updated to assert the new `context_id` kwarg.

### Verification

- 219 tests pass (208 regression + 11 in the new test file).
- `ruff check` clean.
- Diff: +204/-22 across 4 files.

JFK:
## Loop 2 fixes — `757a7479`

Both new findings addressed. Thanks @Copilot.

### F4 (line 1181) — Per-context embedding model

The H1 probe now loads `ContextSearchConfig` for the current context and constructs `EmbeddingService(self.db, model=ctx_config.embedding_model)` — same source of truth that `SearchService.hybrid_search` uses to pick its embed client (`resolve_routing_from_config`). Both failure modes you flagged are closed:

- Ollama-configured context on an OpenAI-default platform: gate now sees `provider == "ollama"` (from the context's model) and skips correctly.
- OpenAI-configured context on an Ollama-default platform: gate now sees `provider == "openai"` and the BYOK probe fires correctly (no more silent bypass).

The structured deny log gains `embedding_model` + `embedding_provider` fields so admins can distinguish provider-mismatch from genuine no-BYOK denials.

Fenced by `test_shared_context_byok_gate_uses_per_context_embedding_model` — asserts `EmbeddingService.__init__` is called with `model="voyage-3-large"` (fixture).

### F5 (line 1185) — Gate deferred past `analysis_cluster` short-circuit

The gate logic is split from the workspace-id override:

1. `effective_workspace_id` + `is_shared_context_read` are computed up front (downstream call sites — cluster filter, hybrid_search, GraphService, explore_hints — all get the right value).
2. The actual H1 probe + raise is deferred to immediately before `hybrid_search`, AFTER the cluster filter's empty/unknown short-circuit.

Empty/unknown cluster paths now return `results=[]` without ever calling `has_byok_key` or `hybrid_search`. No-cost filtered reads are no longer wrongly 404'd.

Fenced by `test_shared_context_byok_gate_deferred_past_empty_cluster_short_circuit` — asserts both `has_byok_key.assert_not_called()` and `hybrid_search.assert_not_called()` when the cluster pre-resolves to empty.

### Verification
- 221 tests pass (208 regression + 13 in `test_recall_shared_context_paid_by.py`).
- `ruff check` clean.
- Diff: +159/-56 in 2 files (this commit only — earlier commits stand).

JFK:
## Loop 3 — SearchService-level test fence — `1cbb41fe`

Good catch @Copilot. The flag's semantics ARE authorization-sensitive, and verifying them only at the caller (`MemoryService.recall`) leaves the bypass undefended at its actual implementation site.

Added `TestSharedContextReadFlag` class to `tests/services/test_search_service.py` with 3 direct fences:

1. **`test_flag_true_skips_context_and_permission_probes`** — `is_shared_context_read=True` MUST skip BOTH the `is_context_shared` probe AND the `is_workspace_member` probe. Fences the dual concern: the bypass exists because handler-layer `_resolve_context_for_read` already verified access, but the SearchService probes would (a) waste a round-trip and (b) raise `AuthorizationError` for legitimate cross-workspace callers (system_admin, ContextMember-only).

2. **`test_flag_true_propagates_to_both_qdrant_calls`** — `search_memories_qdrant` (semantic) AND `search_memories_fulltext` (keyword/BM25) BOTH receive `is_shared_context=True`. Either site missing the propagation would silently filter out source-workspace memories authored by other users — exactly the Option A failure mode you flagged.

3. **`test_flag_false_runs_legacy_probe`** — regression fence for the opt-in semantics. Default (flag=False) MUST still invoke `is_context_shared`. Prevents an operator flipping the default to True from silently disabling the legacy authorization probe for every recall.

### Verification
- 32 tests pass (19 existing in `test_search_service.py` + 13 in `test_recall_shared_context_paid_by.py`).
- `ruff check` clean.
- Diff: +153 / -0 (test-only, no production code changes).

JFK:
## Loop 4 fix — context_id threading through accounting paths — `<sha-after-push>`

Real bug. The context-aware `has_byok_key` from loop 1 was wired only into the H1 preflight gate — `_prepare_spend_cap_gate` and `resolve_paid_by` still called the legacy form. The accounting paths were therefore out of sync with what `_get_user_api_key` actually selects.

### Threading change

- `EmbeddingService._prepare_spend_cap_gate(workspace_id, context_id=None)` — forwards `context_id` to `has_byok_key`.
- `EmbeddingService.resolve_paid_by(workspace_id, context_id=None)` — same.
- `embedding_service.py:450` (`embed_with_usage`) and `:572` (`embed_batch`) pass the `context_id` they already receive as a parameter.
- `search_service.py:227` `resolve_paid_by(workspace_id)` → `resolve_paid_by(workspace_id, context_id=primary_context_id)` — same source as `embed_with_usage` two lines above.

Both `context_id` parameters default to `None` for backward-compat; legacy callers still get the any-key probe.

### Symptoms closed
- `EmbeddingSpendCapService.check_cap_or_raise` no longer fires on env-fallback (platform-billed) calls in mixed-BYOK workspaces.
- `LLMCallLog.paid_by` accurately matches the actual credential used — protects cost-grade attribution (#524) and billing analytics from silent corruption.

### Test fences (4 new, in `TestContextAwareBYOKThreading`)

1. `test_prepare_spend_cap_gate_passes_context_id_to_has_byok_key`
2. `test_resolve_paid_by_passes_context_id_to_has_byok_key`
3. `test_resolve_paid_by_returns_platform_when_byok_not_applicable` — context-aware probe miss → "platform" (matches env fallback)
4. `test_prepare_spend_cap_gate_context_id_optional_for_backward_compat` — None default preserved

248 tests pass (244 previous + 4 new). `ruff check` clean.

JFK:
## Loop 5 fix — orthogonalize the two flags — `<sha-after-push>`

Real security issue. Thanks for catching it @Copilot.

### Root cause

Loop 1's fix conflated two orthogonal concepts:

- `is_shared_context_read` (handler-verified cross-workspace) — legitimate `is_workspace_member` probe bypass under Option A.
- `is_shared_context` (privacy state) — controls Qdrant `user_id == caller` filter.

I had unconditionally set `is_shared_context = is_shared_context_read` at the top. For a creator reading their own **private** context from a different active workspace, this dropped the user_id filter and returned memories authored by other users (e.g. memories written before the context was made private) — bypassing the private-context access rule you described.

### Fix

Decoupled. `is_shared_context` is now always derived from `ContextService.is_context_shared(primary_context_id)`. The cross-workspace flag now only controls the workspace-member probe bypass:

```python
is_shared_context = False
should_probe_sharing = not isinstance(context_id, list) or is_shared_context_read
if should_probe_sharing:
    is_shared_context = await context_service.is_context_shared(UUID(primary_ctx))

# Redundant workspace-member probe — only skipped under cross-workspace bypass.
if (is_shared_context and not is_shared_context_read
    and not isinstance(context_id, list)):
    if not await perm_service.is_workspace_member(user_id, UUID(workspace_id)):
        raise AuthorizationError(...)
```

### Behavior matrix

| Context privacy | Caller's workspace | `is_shared_context_read` | Filter dropped? |
|---|---|---|---|
| Shared | Same | False | Yes (legacy) |
| Private | Same | False | No (legacy) |
| Shared | Cross | True | Yes (Option A — correct) |
| **Private** | **Cross** | **True** | **No (Option A — keeps filter, fixes loop 5)** |

For list-context, same-workspace legacy behavior (`is_shared_context=False`) preserved per Issue #81; cross-workspace Option A probes the primary and follows its sharing state.

### Tests

3 fences in `TestSharedContextReadFlag`:

1. `test_flag_true_skips_permission_probe_but_still_checks_sharing` — `is_context_shared` IS called (drives filter), `is_workspace_member` is NOT called (legitimate bypass). Tightened from previous "both not called" assertion.
2. `test_flag_true_shared_context_propagates_to_both_qdrant_calls` — shared + cross-workspace → filter dropped on both Qdrant paths.
3. **`test_flag_true_private_context_keeps_user_id_filter` (NEW)** — direct fence for your loop 5 finding: private + cross-workspace → filter kept on both Qdrant paths.

249 tests pass. `ruff check` clean.

JFK:
## Loop 6 fix — reject mixed-privacy in cross-context recall — `<sha-after-push>`

Real issue again. Thanks @Copilot.

### Root cause

Loop 5 fix derived `is_shared_context` from the **primary** context's `is_context_shared`. That's correct for uniform-privacy lists, but breaks when the list mixes privacy: primary shared + secondary private = filter dropped for both, leaking other users' memories in the private secondary (e.g. memories written before that context was made private).

### Fix

Reject mixed privacy at the API boundary (your option 1 — matches the existing H3 + same-embedding-model pattern in the same handler block):

```python
for cid in cross_context_ids[1:]:
    cross_ctx = await _resolve_context_for_read(db, user_id, cid)
    if cross_ctx.workspace_id != current_context.workspace_id:
        return _error_response("workspace_mismatch", ...)
    if cross_ctx.is_private != current_context.is_private:
        return _error_response("context_privacy_mismatch", ...)
```

Per-context Qdrant filters (your option 2) would have required restructuring `_build_search_filter` to emit per-context conditions — more invasive, and the cross-context recall API doesn't yet support per-context predicates elsewhere, so the hard-reject keeps the contract consistent.

### Tests

Added handler-level fence `test_cross_context_recall_rejects_mixed_privacy`:

- Constructs two `Mock` contexts: same workspace, `is_private=False` for primary, `is_private=True` for secondary.
- Patches lazy-imported `_resolve_context_for_read` (at both `_helpers` source and `memory` re-export) and `db.base.get_db`.
- Asserts response `{"status": "error", "error": "context_privacy_mismatch"}`.

This is the first handler-level fence for the cross-context invariants in this PR. H3 (workspace_mismatch) + same-embedding-model still rely on integration tests for handler-level coverage; the new test pattern can be extended to fence those if needed.

250 tests pass (249 + 1 new). `ruff check` clean.

JFK:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

JFK:
## Loop 7 fixes — all 3 findings addressed — `5bc67081`

Real issues, especially F7-3. Thanks @Copilot.

### F7-1 — `has_byok_key(context_id=None)` aligned with `_get_user_api_key`

The legacy default branch in `has_byok_key` was matching ANY workspace+provider row regardless of `context_id` scope. `_get_user_api_key(..., context_id=None)` requires `context_id IS NULL` (workspace-wide only). Fixed by adding the symmetric `ExternalAPIKey.context_id.is_(None)` filter when `context_id` is None. Test docstring updated to reflect the correct legacy-priority semantics.

### F7-2 — Stale comment removed

The inline comment at `memory_service.py:1307-1313` claimed `is_shared_context_read=True` "treats the request as shared (drops user_id == caller from the Qdrant filter)". Post-loop-5 this is incorrect: the flag only bypasses the redundant `is_workspace_member` probe; the user_id filter is dropped only when `ContextService.is_context_shared(...)` returns True. Comment now reflects the orthogonal-flag reality.

### F7-3 — TOCTOU fail-closed

Real race window between preflight `has_byok_key` probe in `MemoryService.recall` and `_get_user_api_key` in `EmbeddingService.embed_with_usage`. Under READ COMMITTED, the key can be disabled between the two — and the existing env fallback would silently route to `OPENAI_API_KEY`, bypassing PR #711's BYOK-only spend cap and the H1 guard.

Fix: threaded `disallow_env_fallback: bool` through the chain:

```
search_service.hybrid_search(is_shared_context_read=True)
  → embed_svc.embed_with_usage(disallow_env_fallback=True)
    → _get_client(disallow_env_fallback=True)
      → _get_user_api_key(disallow_env_fallback=True)
        → raises ConfigurationError instead of env fallback when no DB key
```

A structured `logger.warning("openai_api_key_env_fallback_disallowed", reason="option_a_shared_context_read")` fires whenever the disallow kicks in, so the race condition becomes observable in production.

### Tests — 5 new fences

- `test_disallow_env_fallback_raises_instead_of_env` — env set but flag=True + no DB key → ConfigurationError
- `test_default_allows_env_fallback` — regression: default=False still uses env
- `test_disallow_env_fallback_does_not_block_db_key` — DB BYOK still resolves when flag=True
- `test_embed_with_usage_propagates_disallow_env_fallback` — threading verified
- `test_flag_true_propagates_disallow_env_fallback_to_embed` (SearchService) — `is_shared_context_read=True` → `disallow_env_fallback=True`

255 tests pass. `ruff check` clean. Diff: +294/-32 across 5 files.

JFK:
## Loop 8 fix — close CWE-639 leak in TOCTOU deny path — `8215f36d`

Real CWE-639 issue caught. My loop-7 `disallow_env_fallback` fix introduced the very leak the H2 contract is meant to hide.

### Root cause

When the TOCTOU race fires (BYOK row disabled between preflight `has_byok_key` and `_get_user_api_key`), my code only logged the warning and fell through to:

```python
raise ConfigurationError(
    f"OpenAI API key not configured for workspace {workspace_id}. ..."
)
```

`workspace_id` is the SOURCE workspace under Option A. MCP exception serializer surfaces this message verbatim → workspace UUID leaks to the caller.

### Fix

Restructured the disallow branch to raise `NotFoundException("Context")` BEFORE the env-fallback / ConfigurationError logic:

```python
if disallow_env_fallback:
    logger.warning("openai_api_key_env_fallback_disallowed", ...)
    raise NotFoundException("Context")

env_key = os.getenv("OPENAI_API_KEY")
...
```

The handler's existing `except NotFoundException` arm produces the uniform `context_not_found` response — same shape as the H1 deny. No handler change needed.

The `NotFoundException` is constructed WITHOUT a resource_id, so even if its `.message` ever surfaces accidentally it contains only "Context not found" — no workspace UUID, no context UUID.

### Test

Updated `test_disallow_env_fallback_raises_notfound_not_configerror` (renamed):
- Asserts `NotFoundException` raised (not `ConfigurationError`).
- Asserts source `workspace_id` NOT in `.message` or `.details` — defense-in-depth fence.
- Asserts NOT a `ConfigurationError` subclass.

255 tests pass. `ruff check` clean. Diff: +47/-16 in 2 files.

Issue #613, opened 2026-05-12
test(infra): schema-drift detector for Base.metadata.create_all() vs alembic upgrade head


JFK:
## Context

Optional acceptance criterion from #600 (closed via PR #610) — explicitly deferred to keep the mini-batch tight. PR #610 fixed 16 specific `server_default` drift cases on `models/sleep.py` (15 columns + SleepReport.id from Copilot loop 3) by manually mirroring alembic a91/c02/e01.

A drift detector test would catch future drift of this shape automatically.

## Problem statement

`Base.metadata.create_all()` (used by integration test fixtures + dev DB bootstrap) generates DDL from the **model file**. Alembic migrations generate DDL from the **migration file**. When the two diverge:

- Production (alembic-driven) is correct
- Tests / fresh dev DB (create_all-driven) silently lacks DEFAULT clauses
- Raw INSERT paths (test fixtures, bulk loaders, future migrations) fail NOT NULL on omitted columns instead of getting documented defaults

PR #610's #600 audit was manual — easy to miss columns (Copilot loop 3 caught SleepReport.id that the audit overlooked).

## Proposed solution

Write a test that:

1. Creates a clean DB with `Base.metadata.create_all()`
2. Captures the DDL via `pg_dump --schema-only`
3. Tears down, creates another clean DB, runs `alembic upgrade head`
4. Captures DDL via `pg_dump --schema-only`
5. Diffs the two
6. Asserts the diff is empty (or only contains expected exceptions per allowlist)

## Acceptance criteria

- [ ] Test file: `backend/tests/integration/test_schema_drift.py`
- [ ] Uses ephemeral test DB (existing `db_session` fixture pattern)
- [ ] Captures DDL via `pg_dump --schema-only` or pyscopg `pg_get_tabledef` equivalent
- [ ] Diffs CREATE TABLE statements + DEFAULT clauses + CHECK constraints
- [ ] Asserts diff is empty
- [ ] Optional: small allowlist for known-divergent edges (e.g. comment ordering, whitespace)
- [ ] Runs in `make test-integration` (Docker-required)
- [ ] Documented in `backend/tests/README.md` or similar

## Sister precedent

`#587` introduced an ORM CHECK constraint vs migration drift detector. This is the same idea applied to the broader DDL surface (DEFAULT clauses + column types + constraints).

## Out of scope

- Reverting / fixing existing drift (PR #610 fixed sleep.py; other models might have drift but that's a separate audit)
[...]

PR #610, opened 2026-05-11
chore(api,models,types): pyright tier-A re-enable + DDL drift cleanup (#599, #600)

JFK:
Closes #599
Closes #600

## Summary

Two #596 (Mapped[T] migration epic #370) follow-up cleanup issues bundled as a mini-batch (file-area disjoint):

- **#599 Tier A only** — re-enable two pyright rules suppressed since #596 (`reportAssignmentType` + `reportReturnType`) by fixing 9 caller-side type narrowings. Tier B (~58 fixes), Tier C (~163 fixes), Tier D (4 broader rules) remain deferred.
- **#600** — fix DDL drift on `sleep.py` by adding `server_default` to 15 columns (matches alembic a91/c02/e01); apply 3 cosmetic consistency cleanups (B1 kwarg order, B2 `uuid.uuid4` qualified, B3 `ARRAY` dialect import).

## Implementation notes

**#599 Tier A — 9 caller-side fixes**:
- `api/routes/contexts.py` (×2): widen `creator_names` dict value to `str | None` to match nullable `User.name`; rename `label("count")` → `label("memory_count")` to avoid `tuple.count` collision in pyright; explicit `int(row.memory_count)` cast.
- `services/workspace_service.py` (×2): insert `if member is None: raise NotFoundException(...)` after `get_member` calls in `update_member_role` + `update_member_context_access`. `get_member` raises by default — this is forward-compat narrowing for any future call site that passes `raise_if_not_found=False`. Caller audit (`api/routes/workspaces.py`) confirmed no None-check dead branches to remove.
- `auth/roles.py`: add `return []` after `async for db in get_db():` loop in `list_users` for the no-iteration fall-through.
- `auth/oauth2.py`: `cast(Credentials, flow.credentials)` on web-flow code-exchange return; new `from typing import cast` import.
- `db/qdrant.py`: switch `_admin_scroll_context_points` offset variable to `qdrant_client.conversions.common_types.PointId | None` (function-local import).
- `pyproject.toml`: remove `reportAssignmentType = false` and `reportReturnType = false`. Audit comment rewritten as Tier A/B/C/D status table.

**#600 — DDL drift + cosmetic**:
[...]

PR #716, opened 2026-05-18
test(infra): create_all vs alembic head schema-drift detector (#613)

JFK:
Partially addresses #613

## Summary

- Adds a new integration test that compares the DDL produced by `Base.metadata.create_all()` against the DDL produced by `alembic upgrade head` and asserts no actionable drift.
- Catches the class of bug fixed in PR #610 (16 `server_default` columns on `models/sleep.py` that the ORM missed). Complementary to sister #587 / PR #590 (AST-based CHECK constraint detector).
- Discovers and registers 34 baseline drift entries across 5 categories with reason comments + follow-up plans — turning the detector into an immediate defence line while the underlying drift is paid down via separate PRs.

## Implementation notes

- **Single file**: `backend/tests/integration/test_create_all_vs_alembic_drift.py` (836 LOC, no production code changes).
- **DDL capture**: `information_schema.columns` + `information_schema.table_constraints` + `information_schema.key_column_usage` + `information_schema.constraint_column_usage` + `pg_catalog.pg_index/pg_class/pg_namespace/pg_attribute`. `pg_dump` (unavailable in the test image's Dockerfile) and `pg_get_tabledef` (not present on vanilla Postgres 15) were ruled out by exploration.
- **Fixture**: reuses `_reset_alembic_state`, `_alembic_at_test_db`, `_get_alembic_config`, `_sync_engine` from `backend/tests/integration/test_alembic_migrations.py` via the established import precedent in `test_resources_foundation_migration.py:33-37`. Two sequential `DROP SCHEMA public CASCADE; CREATE SCHEMA public;` resets — one before each capture. Test ends with the DB at alembic head so subsequent integration tests are unaffected.
- **Single test method** with aggregated `pytest.fail()` reporting (1 pytest item, N drift cases listed together). Parametrize was rejected because it would require module-import-time DB queries — brittle under `--collect-only`, risks running in unit-test collection contexts.
- **Allowlist** as `frozenset[str]` keyed by stable_ids of the form `<table>.<name>.<kind>.<side>`. Each entry has a category comment and a follow-up plan.

## Renamed from issue's proposed filename

Issue proposed `backend/tests/integration/test_schema_drift.py`. Renamed to `test_create_all_vs_alembic_drift.py` because the existing `backend/tests/test_schema_drift.py` (sister #587 detector) would otherwise collide on grep / IDE jump.

## Why "Partially addresses" instead of "Closes"

The issue AC explicitly lists "CHECK constraints" as part of the diff target. This detector delegates CHECK to sister `backend/tests/test_schema_drift.py` (#587 / PR #590) — that is the established pattern in this repo, and replicating CHECK detection here would duplicate the sister. The detector covers every other AC item (CREATE TABLE / DEFAULT clauses / column types / non-CHECK constraints / indexes), but a literal reading of the AC makes this a subset PR. Marking accordingly to avoid GitHub auto-close on a subset.

## Baseline drift — 34 entries registered for follow-up

First live run on `kagura_test` against alembic chain at `e15_675_workspace_slot_bonus` found 34 real drift entries. Each is registered in `_KNOWN_DRIFT` with category comment + follow-up plan. As each follow-up PR lands, the corresponding entries MUST be deleted from `_KNOWN_DRIFT` to turn the detector into a real defence line.

**Cat A — `server_default` drift (8 columns, high priority)**: ORM declares the column without `server_default=...` while the alembic migration emits a literal default. Same pattern as PR #610. Affects `erasure_requests.is_self_service`, `erasure_requests.status`, `file_objects.id`, `llm_pricing.context_min_tokens`, `llm_pricing.currency`, `llm_pricing.unit_denominator`, `users.auth_method`, `users.totp_enabled`.

**Cat B — FK naming-convention drift (14 entries / 7 pairs)**: SQLAlchemy auto-names FK constraints `<table>_<col>_fkey` while alembic migrations assigned explicit `fk_<table>_*` names. Same physical FKs on both sides. Root fix: configure `Base.metadata.naming_convention` so SQLAlchemy and alembic produce identical names.

**Cat C — UK naming drift (2 entries)**: Same root as Cat B for UNIQUE constraints on `oauth_device_codes`. Likely folds into Cat B once the naming convention is configured.

**Cat D — Migration-only indexes (7 entries)**: Partial WHERE indexes, GIN access method, composite indexes, partial UNIQUE — all created by alembic but not declared via `Index(...)` in the ORM. Resolution: declare matching `Index` / `UniqueConstraint` entries in the ORM models.

**Cat E — Uniqueness mismatch (3 entries, potentially data-integrity bug)**: `ix_oauth_device_codes_device_code` / `ix_oauth_device_codes_user_code` `is_unique` flag differs between ORM and alembic. **`uq_file_objects_workspace_sha256_active`** — ORM declares unique on `(workspace_id, sha256)` but alembic produces only `(workspace_id,)`. Single-column unique on `workspace_id` would mean "one file per workspace", which is clearly wrong — needs investigation as a possible data-integrity bug.

## Known limitations (documented in module docstring)

- **CHECK constraints**: delegated to sister `backend/tests/test_schema_drift.py` (#587 / PR #590) via AST diff.
- **Sequence ownership** (`ALTER SEQUENCE … OWNED BY`): invisible to `information_schema`. Functional equivalence preserved via matching `nextval(...)` column default on both sides.
- **ENUM body** (value-set additions / removals): `information_schema.columns.data_type` returns only the literal `USER-DEFINED` token; value-set drift is invisible.
- **Multi-column FK column ordering**: `information_schema.constraint_column_usage` produces a Cartesian product; the dedup-then-sort aggregation loses positional information. No multi-column FKs exist in the current schema. Fix plan (switch to `pg_constraint.conkey/confkey` array join) documented in `_introspect_fk` docstring.
- **`_strip_type_cast` empty-string collapse**: `column_default = ''` is returned as `None`. No current trigger in this schema; fix plan documented.
- **`_diff_*` 3-function structural clone (~80%)**: deferred refactor when a 4th diff category is added.

## CI coverage

`tests/integration/` is currently not run in CI (`backend-unit` job uses `--ignore=tests/integration`). The detector runs locally via `make test-integration`. Adding a `backend-integration` CI job with path-based triggering for `backend/src/models/**` and `backend/alembic/versions/**` is **deliberately deferred to a follow-up issue** so this PR stays scope-clean.

## Verification

- `ruff check` / `ruff format` clean on the new file.
- `pytest --collect-only` collects exactly 1 test.
- Live run against the test DB after `_KNOWN_DRIFT` registration: **PASSED in 2.06s**.

## Pre-PR review summary

- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- gate1: green via /claude-c-suite:ask (QA Lead lens)

- ~/.claude/cache/gh-issue-driven/613-feat-test-infra-schema-drift-detector-for-bas.gate1.md
- ~/.claude/cache/gh-issue-driven/613-feat-test-infra-schema-drift-detector-for-bas.gate2.md

### DIFF
--- a/backend/tests/integration/test_create_all_vs_alembic_drift.py
+++ b/backend/tests/integration/test_create_all_vs_alembic_drift.py
@@ -46,11 +46,13 @@ For every user-defined table in the ``public`` schema (excluding the
   alembic for ``BIGSERIAL`` surrogates but not surfaced through
   ``information_schema``. Both snapshots see the same ``nextval(...)``
   column default, so functional equivalence is preserved.
-- **ENUM type body** — ``information_schema.columns.data_type`` returns
-  the literal string ``USER-DEFINED`` for every ENUM regardless of the
-  enum name. Both ``create_all`` and alembic produce the same string, so
-  ENUM column existence drift IS caught, but value-set drift (adding /
-  removing labels) is NOT. That belongs to its own audit.
+- **ENUM value set (label additions / removals)** —
+  ``information_schema.columns.data_type`` returns the literal string
+  ``USER-DEFINED`` for every ENUM, but ``udt_name`` carries the actual
+  ENUM type name, so type-rename drift IS now caught. **Value set
+  drift** (e.g. adding ``'pending'`` to an existing ``status_enum``)
+  requires ``pg_enum`` introspection and remains out of scope; it
+  belongs to its own audit.
 - **Function-call defaults whose Postgres normalization is non-stable**
   (e.g. ``now()`` vs ``CURRENT_TIMESTAMP``). When both paths use the
   same SQL, Postgres normalizes both to the same stored representation,
@@ -273,6 +275,13 @@ class ColumnRow(TypedDict):
     ``numeric_precision`` / ``numeric_scale``, and ``datetime_precision``
     so that drift in fully-qualified types like ``NUMERIC(14, 10)`` or
     ``TIMESTAMP(6)`` is caught — ``data_type`` alone would compare equal.
+
+    ``udt_name`` / ``udt_schema`` carry the **underlying type name** for
+    ``data_type`` values that lose information (notably ``ARRAY`` —
+    ``udt_name`` becomes ``_int4`` / ``_text`` etc. with a leading
+    underscore; and ``USER-DEFINED`` — ``udt_name`` becomes the ENUM /
+    domain type name). Comparing these catches ARRAY element-type drift
+    and ENUM type-name drift that ``data_type`` alone misses.
     """
 
     table: str
@@ -284,6 +293,8 @@ class ColumnRow(TypedDict):
     numeric_precision: int | None
     numeric_scale: int | None
     datetime_precision: int | None
+    udt_name: str
+    udt_schema: str
 
 
 class ConstraintRow(TypedDict):
@@ -318,13 +329,21 @@ class ConstraintRow(TypedDict):
 
 
 class IndexRow(TypedDict):
-    """One index excluding PK / UK auto-indexes."""
+    """One index excluding PK auto-indexes.
+
+    ``access_method`` is the index access method (``btree`` / ``gin`` /
+    ``gist`` / ``hash`` / ``brin`` / ``spgist``) from ``pg_am.amname``.
+    Comparing it catches drift where an ORM-declared ``Index`` defaults
+    to ``btree`` while alembic uses ``postgresql_using='gin'`` for the
+    same name + columns.
+    """
 
     table: str
     name: str
     columns: tuple[str, ...]
     is_unique: bool
     predicate: str | None  # WHERE clause for partial indexes
+    access_method: str  # btree | gin | gist | hash | brin | spgist
 
 
 class SchemaSnapshot(TypedDict):
@@ -396,7 +415,9 @@ _COLUMNS_SQL = """
         character_maximum_length,
         numeric_precision,
         numeric_scale,
-        datetime_precision
+        datetime_precision,
+        udt_name,
+        udt_schema
     FROM information_schema.columns
     WHERE table_schema = 'public'
     ORDER BY table_name, ordinal_position
@@ -453,6 +474,7 @@ _INDEXES_SQL = """
         i.relname AS index_name,
         ix.indisunique AS is_unique,
         ix.indisprimary AS is_primary,
+        am.amname AS access_method,
         ARRAY(
             SELECT pg_get_indexdef(ix.indexrelid, k.ord::int, true)
             FROM generate_series(1, ix.indnatts) WITH ORDINALITY AS k(_, ord)
@@ -463,6 +485,7 @@ _INDEXES_SQL = """
     JOIN pg_class i ON i.oid = ix.indexrelid
     JOIN pg_class c ON c.oid = ix.indrelid
     JOIN pg_namespace n ON n.oid = c.relnamespace
+    JOIN pg_am am ON am.oid = i.relam
     WHERE n.nspname = 'public'
         AND NOT ix.indisprimary
     ORDER BY c.relname, i.relname
@@ -488,6 +511,8 @@ def _introspect_columns(conn: Connection) -> dict[tuple[str, str], ColumnRow]:
             numeric_precision=row.numeric_precision,
             numeric_scale=row.numeric_scale,
             datetime_precision=row.datetime_precision,
+            udt_name=row.udt_name,
+            udt_schema=row.udt_schema,
         )
     return result
 
@@ -609,6 +634,7 @@ def _introspect_indexes(conn: Connection) -> dict[str, IndexRow]:
             columns=tuple(row.columns),  # ordinal order preserved
             is_unique=row.is_unique,
             predicate=row.predicate,
+            access_method=row.access_method,
         )
     return result
 
@@ -711,6 +737,8 @@ def _diff_columns(
             "numeric_precision",
             "numeric_scale",
             "datetime_precision",
+            "udt_name",
+            "udt_schema",
         ):
             if row_a[attr] != row_b[attr]:
                 diffs.append(f"{attr}: create_all={row_a[attr]!r}, alembic={row_b[attr]!r}")
@@ -820,7 +848,7 @@ def _diff_indexes(
         row_a = a[name]
         row_b = b[name]
         diffs: list[str] = []
-        for attr in ("columns", "is_unique", "predicate"):
+        for attr in ("columns", "is_unique", "predicate", "access_method"):
             if row_a[attr] != row_b[attr]:
                 diffs.append(f"{attr}: create_all={row_a[attr]!r}, alembic={row_b[attr]!r}")
         if diffs:
--- a/backend/tests/integration/test_create_all_vs_alembic_drift.py
+++ b/backend/tests/integration/test_create_all_vs_alembic_drift.py
@@ -239,10 +239,18 @@ _KNOWN_DRIFT: frozenset[str] = frozenset(
         #   * ix_oauth_device_codes_device_code (is_unique=True/False)
         #   * ix_oauth_device_codes_user_code   (is_unique=True/False)
         #     → ORM declares ``unique=True, index=True``; alembic emits
-        #       ``op.create_index(unique=False)``. The device/user code
-        #       MUST be unique for OAuth device-code grant correctness,
-        #       so alembic appears to need a fix — but verify against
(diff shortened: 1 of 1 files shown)


JFK:
### Code review

JFK:
## Loop 1 — addressed all 4 Copilot Auto-review findings

All four inline findings on commit `80fe34e6` are addressed in `95c83862`. Live test re-run **PASSED** in 2.26s.

### #L417 — Expression-aware index introspection (correctness fix)

Replaced `unnest(ix.indkey) JOIN pg_attribute` with `pg_get_indexdef(indexrelid, ord, true)` over `generate_series(1, indnatts)`. Now correctly captures functional/expression index elements that Postgres stores in `pg_index.indexprs` (where `indkey.attnum=0`).

**Important downstream impact**: this fix changes the Cat E `uq_file_objects_workspace_sha256_active` finding from the previously-reported `('workspace_id',)` to the actual `('workspace_id', 'lower(sha256::text)')` — confirming this is migration `e07_556_sha256_lowercase_index`'s case-insensitive dedup index. **This is not a data-integrity bug**: production correctly enforces case-insensitive sha256 uniqueness; the drift is that the ORM `UniqueConstraint` declaration didn't follow the migration's case-insensitive change. The Cat E comment in `_KNOWN_DRIFT` is updated accordingly, downgrading the severity to "ORM declaration alignment".

### #L119 — `models.llm_call_log` added to `_MODEL_MODULES`

Trivial 1-line fix in alphabetical position. Closes the isolation-run gap.

### #L358 — NUMERIC precision/scale comparison

Extended `_COLUMNS_SQL` to select `numeric_precision`, `numeric_scale`, `datetime_precision`. Extended `ColumnRow` TypedDict + `_diff_columns` comparison loop. Now catches `NUMERIC(14, 10) vs NUMERIC(10, 2)` drift on pricing/cost columns that `data_type='numeric'` alone would miss.

### #L28 — Module docstring fix

Original claim "PK / UK auto-indexes are filtered out" was wrong (only PK is filtered; UK is intentionally kept). Docstring now matches `_introspect_indexes` and the `_KNOWN_DRIFT` Cat D `oauth_device_codes_*_key.index.alembic_only` entries.

---

Thanks for the thorough review — finding #L417 in particular caught a real detector blindspot that was producing a misleading Cat E narrative.

JFK:
## Loop 2 — FK referential actions comparison

Addressed Copilot Auto-review #L419 on commit `95c83862`. New fixup `000fd096`.

**Change**: extended FK introspection to compare `ON UPDATE` / `ON DELETE` referential actions.

- `_FK_SQL`: JOIN `information_schema.referential_constraints` for `update_rule` / `delete_rule`
- `ConstraintRow` TypedDict: add `update_rule` / `delete_rule` (str | None — populated for FK rows, None for PK/UK)
- `_introspect_pk_uk`: pass None for both on PK/UK
- `_introspect_fk`: capture rules from the first row per constraint (constraint-scoped)
- `_diff_constraints`: include both attributes in the value-mismatch comparison loop

**Verification**: ruff / format clean. Live test **PASSED in 2.10s**. **No new actionable drift surfaced** — the FKs that exist on both sides under matching names already agree on referential actions. Cat B FK pairs (different names per side) remain in the "only on one side" buckets and do not exercise attribute comparison.

Total: 836 → 863 LOC (+27 on this commit). 34 entries in `_KNOWN_DRIFT` (unchanged).

JFK:
## Follow-up issues filed for the 34-entry baseline allowlist

Per the PR body's "follow-up issues" plan, 5 issues are now filed for systematic cleanup of `_KNOWN_DRIFT`. Each subsequent PR should delete its category's entries from the allowlist as the underlying drift is fixed.

| Issue | Category | Entries | Priority |
|---|---|---|---|
| #717 | Cat A — `server_default` audit (8 columns) | 8 | HIGH |
| #718 | Cat B+C — FK/UK naming-convention unification | 16 | MEDIUM |
| #719 | Cat D — ORM `Index(...)` parity for 7 indexes | 7 | MEDIUM |
| #720 | Cat E — ORM uniqueness alignment (file_objects + oauth_device_codes) | 3 | MEDIUM (downgraded from HIGH after loop-1 narrative correction) |
| #721 | CI — `backend-integration` GHA job for `tests/integration/` | (infra) | MEDIUM |

Total: 34 / 34 entries tracked. When all 5 issues land, `_KNOWN_DRIFT` should be `frozenset()` again — at which point the detector becomes a true defence line against new drift.

JFK:
## Loop 3 — Preserve PK/UK column order

Addressed Copilot Auto-review #L506 on commit `000fd096`. New fixup commit pushed.

**Issue**: `_introspect_pk_uk()` used `tuple(sorted(entry["columns"]))` which destroyed the `ORDER BY kcu.ordinal_position` from the SQL. Because `_INDEXES_SQL` filters PK auto-indexes, a composite PRIMARY KEY with different column order between `create_all` and alembic (e.g. `(analysis_id, memory_id)` vs `(memory_id, analysis_id)`) would silently pass the detector. The repo already has composite PKs like `memory_analysis_assignments` that would be exposed by this gap.

**Fix**:
- `_introspect_pk_uk`: drop the `sorted()`; preserve `ordinal_position` order from the SQL
- `ConstraintRow` docstring: clarify that PK/UK column order is now preserved (B-tree prefix-scan plans differ between `(a, b)` and `(b, a)`); FK order remains a documented known limitation
- Module docstring sections 2 + 3: revise the "inherently unordered" wording to reflect that order IS drift-significant for composite PK/UK

**Verification**: ruff/format clean. Live test **PASSED in 2.10s**. No new actionable drift — the existing composite PKs/UKs (notably `memory_analysis_assignments`) already agree on column order. The fix prevents future regressions of that property from silently passing.

JFK:
## Loop 4 — udt_name + index access_method capture

Addressed both Copilot Auto-review findings on commit `e3f2229e`. New fixup `80ab9cd9`.

### #L402 — ARRAY / USER-DEFINED type drift
`data_type` returns literal `ARRAY` / `USER-DEFINED` for these. Two different ENUM types would both look `USER-DEFINED` and compare equal. Added `udt_name` + `udt_schema` to `_COLUMNS_SQL`, `ColumnRow`, `_introspect_columns`, and `_diff_columns`. Module docstring updated — ENUM **type-name** drift is now caught; only ENUM **value-set** drift (requires `pg_enum`) remains out of scope.

### #L466 — Index access method
`_INDEXES_SQL` didn't capture btree/gin/gist/hash/brin/spgist. Same-name same-columns btree-vs-gin would silently match. JOINed `pg_am` on `pg_class.relam`, added `access_method` to `IndexRow` + introspect + diff.

**Verification**: ruff / format clean. Live test **PASSED in 2.10s**. No new actionable drift — existing columns/indexes on both sides already agree on udt_name and access_method. The fixes prevent future regressions from silently passing.

Total: 4 loops, 7 detector enhancements (loop 1: 4, loop 2: 1, loop 3: 1, loop 4: 2). `_KNOWN_DRIFT` still 34 entries (the detector strengthening hasn't unearthed any current schema drift, only future-proofed the defense line).

JFK:
## Loop 5 — Documentation precision fixes

Addressed Copilot Auto-review #L662 + #L245 on commit \`80ab9cd9\`. New fixup \`24c13476\`. Both are docstring / comment fixes; no behavior change.

**#L662** — \`_capture_schema\` docstring rewritten to describe the actual three-map shape (\`columns\` / \`constraints\` (merged PK+UK+FK) / \`indexes\`) and to explain why the merge is safe (postgres constraint-name uniqueness).

**#L245** — Cat E \`oauth_device_codes\` comment rewritten to remove the overstated correctness claim. Migration \`d08_536_device_code_grant\` already declares the columns with \`unique=True\` so uniqueness is enforced via UNIQUE constraint + backing unique index. The drift is index **redundancy** (ORM emits two unique indexes per column; alembic emits one unique-backing + one non-unique lookup), not a correctness gap. Fix path: drop \`unique=True\` from the ORM \`mapped_column\` (keep \`index=True\`).

Live test **PASSED in 2.06s**. \`_KNOWN_DRIFT\` unchanged (34 entries). Ready to merge.

---

Loops 1-5 summary: **5 Copilot Auto-Review loops, 10 findings total, all addressed**. Detector strengthened across columns (server_default + NUMERIC precision + udt_name), constraints (FK referential actions + PK/UK column order), and indexes (expression-aware via pg_get_indexdef + access_method). 34 entries in \`_KNOWN_DRIFT\` will be paid down via the 5 follow-up issues filed (#717-#721).

PR #811, opened 2026-05-27
refactor(api): cache pricing + dedup BYOK probe on recall (#713)

### LINKED ISSUE #713: Hot-path perf bundle for #709 spend cap (pricing TTL cache + api-key source tuple)
## Summary

Performance follow-up to #709 / PR #711 (squash `4b8e3b64`). Two related hot-path optimizations identified by `/simplify` Agent 3 and reaffirmed by gate2 CTO review:

1. **`LLMPricingService` process-local TTL cache** — every `embed_with_usage` cache-miss now runs `LLMPricingService.compute_cost_usd(...)` inside `EmbeddingSpendCapService.record_spend_from_tokens` for cost computation. `LLMCallLogWriter.record()` also runs the same pricing lookup 1 step later in `search_service.py`. Result: pricing-row SELECT runs **twice per recall** for the same `(provider, model, embedding_tokens, started_at)`. Pricing rows change at most a few times per year; in-process TTL cache (60-minute TTL) eliminates the duplicate.

2. **`_get_user_api_key` return `(key, source)`** — `EmbeddingService.has_byok_key()` runs an indexed SELECT on `external_api_keys` to set `LLMCallLog.paid_by` (via `EmbeddingService.resolve_paid_by`). The same SELECT is performed inside `_get_user_api_key` moments later when retrieving the key. Refactor `_get_user_api_key` to return `(api_key, source: Literal["workspace", "context", "env"])` so the caller can derive `paid_by` from `source` without a second SELECT.

Together: saves **2 DB round-trips per recall** on the embed hot path.

## Context

**Priority**: medium
**Suggested labels**: enhancement
**Suggested milestone**: v0.17.0 (UI polish + admin /plans audit) — or carve a v0.16.x perf milestone

### (1) LLMPricingService TTL cache
- `backend/src/services/llm_pricing_service.py:148` — `compute_cost_usd` is the hot-path entry. Cache key = `(provider, model, unit_type, started_at_day)`.
- Implementation candidates: `functools.lru_cache` (no TTL) — risk if pricing changes mid-day; `cachetools.TTLCache` (60-min TTL) — safer. Prefer the latter.
- Cache invalidation: not required for v1; pricing changes are infrequent and a 60-min lag is acceptable.

### (2) `_get_user_api_key` return tuple
- `backend/src/services/embedding_service.py:74-169` — current signature `_get_user_api_key(user_id, context_id, workspace_id) -> str`. Refactor to `-> tuple[str, Literal["workspace", "context", "env"]]`.
- Single caller: `_get_client` (line 210). Update to unpack the tuple.
- `has_byok_key` becomes obsolete: callers compute `paid_by = "byok" if source != "env" else "platform"` from the source tuple element.
- `resolve_paid_by` updates: delegate to `_get_client` to get source, or accept a pre-computed source from the caller.

## Acceptance criteria

- [ ] `LLMPricingService.compute_cost_usd` results cached process-locally with 60-min TTL (`cachetools.TTLCache` or equivalent). Cache key includes `(provider, model, unit_type, started_at.date())`.
- [ ] Cache hit avoids the `llm_pricing` SELECT entirely on the hot path; verified by mocking the underlying `lookup` and asserting call count ≤ 1 for 100 successive `compute_cost_usd` calls with the same key.
- [ ] `_get_user_api_key` returns `(api_key, source)`. `source` is one of `"workspace"`, `"context"`, `"env"`.
- [ ] `_get_client` unpacks the tuple; `has_byok_key` is either removed or rewritten as a thin wrapper around the new return shape (avoid duplicate SELECT).
- [ ] `resolve_paid_by` no longer issues its own SELECT; derives `paid_by` from the source value.
- [ ] Existing 67 cap-service + endpoint + embed/search tests still pass.

## Out of scope

- Workspace-row cache (separate concern, not on the hot path's same fix surface).
- Redis-based pricing cache (process-local is sufficient — pricing rows are read-mostly).
- Provider-level price drift alerts (operational, not perf).

## Related

- #709 / PR #711 (originating PR — deferred follow-up explicitly listed in body's Out-of-scope section)
- `/simplify` Agent 3 F4 (pricing cache) + F6 (has_byok_key piggyback) — both findings filed here as bundle

JFK:
Closes #713

## Summary
- Removes **2 DB round-trips per recall** on the embedding hot path (the last open issue in milestone v0.17.0).
- Collapses the duplicate `llm_pricing` SELECT (`record_spend_from_tokens` + `LLMCallLogWriter.record`) into one via a process-local TTL cache.
- Eliminates the duplicate `external_api_keys` SELECT by deriving `paid_by` from the key tier already resolved during the embed, instead of a second `has_byok_key` probe.

## Implementation notes
- **`LLMPricingService`**: new module-global `cachetools.TTLCache` (60-min TTL, `maxsize=1024`) keyed by `(provider, model, unit_type, started_at.date(), context_tokens)` — `context_tokens` retained so tier-priced models stay correct. `compute_cost_usd` routes through `_cached_price_components` (single `__getitem__` via `try/except KeyError` — race-free, negative-caches misses). `clear_pricing_cache()` exposed for tests + post-seed refresh.
- **`EmbeddingService`**: `_get_user_api_key` now returns `(api_key, source: Literal["workspace","context","env"])`; `_get_client` records `self._last_key_source`; `resolve_paid_by` derives `paid_by` (`env`→`platform`, else `byok`) from it. The source is the actual key used — strictly more accurate than the prior probe — so the #708 loop-4 probe/lookup parity is closed by construction. `has_byok_key` retained as a defensive fallback (Ollama / standalone callers). The memo is reset at each `embed_*` entry so a reused, request-scoped instance can never read a stale tier.
- New dependency `cachetools>=5.3.0` (this repo's lock did not pull it transitively). Autouse `conftest` fixture clears the global price cache between tests to prevent cross-test price bleed.

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- prior review: `/code-review` at high + max effort (3 finder agents × 9 angles) + `/self-review` — all findings addressed (TTL-clock race → `try/except`, lock/dep, stale-memo reset, doc/type tightening); correctness pass found no behavior bugs.

Tests: 122 unit + 32 integration pass; ruff clean; pyright 0 errors on changed `src/`.

Full gate2 review saved in the plugin cache:
- ~/.claude/cache/gh-issue-driven/713-refactor-spend-cap-hotpath-perf.gate2.md

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
 
     # Cache & Queue
     "redis>=5.2.0",
+    "cachetools>=5.3.0",  # Process-local TTL cache for llm_pricing hot path (#713)
 
     # HTTP Client
     "httpx>=0.27.0",
--- a/backend/src/services/embedding_service.py
+++ b/backend/src/services/embedding_service.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 import json
 import os
 import unicodedata
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 import xxhash
 from openai import AsyncOpenAI
@@ -37,6 +37,12 @@ if TYPE_CHECKING:
     from services.embedding_spend_cap_service import EmbeddingSpendCapService
 
 logger = get_logger(__name__)
+
+# The credential tier ``_get_user_api_key`` resolved a key from. ``"workspace"``
+# / ``"context"`` are BYOK rows in ``external_api_keys``; ``"env"`` is the
+# platform ``OPENAI_API_KEY`` fallback. Maps directly to ``LLMCallLog.paid_by``
+# (``"env"`` → ``"platform"``, otherwise → ``"byok"``) — see ``resolve_paid_by``.
+KeySource = Literal["workspace", "context", "env"]
 
 
 class EmbeddingService:
@@ -79,6 +85,12 @@ class EmbeddingService:
             self.provider = settings.embedding_provider
         self.ollama_base_url = settings.ollama_base_url
         self._ollama_verified = False
+        # Issue #713: the credential tier the most recent ``_get_client`` OpenAI
+        # call resolved (set in ``_get_client``). ``resolve_paid_by`` reads it to
+        # derive ``paid_by`` without re-probing ``external_api_keys`` — the
+        # service is request-scoped (one per recall, holding a per-request
+        # session) so this state is never shared across concurrent requests.
+        self._last_key_source: KeySource | None = None
 
     async def _get_user_api_key(
         self,
@@ -86,7 +98,7 @@ class EmbeddingService:
         context_id: str | None = None,
         workspace_id: str | None = None,  # Issue #146
         disallow_env_fallback: bool = False,
-    ) -> str:
+    ) -> tuple[str, KeySource]:
         """Resolve the OpenAI API key for the calling user's workspace context.
 
         Issue #105: DB-first lookup with environment variable fallback.
@@ -114,7 +126,12 @@ class EmbeddingService:
                 platform key. Has no effect when a DB key is found.
 
         Returns:
-            Decrypted OpenAI API key.
+            ``(api_key, source)`` where ``source`` is the credential tier the
+            key came from — ``"context"`` / ``"workspace"`` for a BYOK row,
+            ``"env"`` for the platform ``OPENAI_API_KEY`` fallback. Issue #713:
+            callers derive ``LLMCallLog.paid_by`` from ``source`` (``"env"`` →
+            ``"platform"``, else ``"byok"``) instead of re-running the BYOK
+            probe via ``has_byok_key``.
 
         Raises:
             ConfigurationError: If neither a DB key nor an env var is available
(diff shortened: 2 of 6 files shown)


JFK:
Thanks — addressed in `27249ba4` (documentation, not a key change). Rationale after tracing both call sites:

**Why not full `started_at` in the key:** the two duplicate lookups this cache exists to collapse — `EmbeddingSpendCapService.record_spend_from_tokens` and `LLMCallLogWriter.record` — each call `utcnow()` independently (neither passes `occurred_at` on the live path), so they carry timestamps milliseconds apart. `started_at.date()` is exactly what collapses them into one SELECT; a full-resolution key would never hit between them and would regress the optimization back to 2 SELECTs.

**Why the bound is acceptable today:**
- Live path: `started_at` is always `utcnow()`, and the 60-min TTL already caps staleness to ≤1h regardless of key granularity — so day-bucketing adds no extra live-path staleness over an intra-day `effective_from` change.
- The historical-replay path (`record_spend_from_tokens(occurred_at=...)`) that *would* expose the intra-day collapse has no live caller (the param is designed but unused).

You're right that the day-granularity reproducibility assumption was silent and `effective_from` is an unconstrained `DateTime`. I've made that explicit in the `_cached_price_components` docstring, including the revisit trigger: **if sub-day-precise historical replay is ever wired up, the key needs a finer time component or the replay must bypass the cache.**

Issue #618, opened 2026-05-12
feat(frontend): tag cloud + autocomplete UI on context detail page (#614 follow-up)


JFK:
## Context

`list_tags` MCP tool + REST endpoint shipped in #614 / #617. Use case #3 from the original issue was "UI surface: tag cloud / autocomplete on context detail page" — out-of-scope at the time, tracked here as the frontend follow-up.

The server side already supports `prefix` filtering and three sort modes (`count`, `recent`, `alpha`) specifically so the frontend can build an interactive tag-cloud and autocomplete without round-tripping the full set.

### Tag cloud (context detail page)

- Display top-N tags by frequency (`sort=count`, `limit=50`), size-coded by `count`
- Click a tag → filter the memory list by that tag (existing recall filter)
- "Recently used" tab toggling to `sort=recent`

### Autocomplete (tag input field)

- Trigger `GET /contexts/{context_id}/tags?prefix=<user-input>&limit=10&sort=count` on debounced keystroke
- Show top 10 suggestions ranked by count
- Insert selected tag into the tag input

## Use cases

1. User browsing a context wants a visual sense of "what's in here"
2. User adding a memory wants tag autocomplete to avoid drift
3. User searching wants to discover tags that exist before filtering

## Out of scope for v1 (frontend side)

- Tag rename / merge UI (server-side non-goal)
- Tag co-occurrence visualization (separate feature)

## Implementation notes

- API: `GET /api/v1/contexts/{context_id}/tags` — already shipped in #617
- Component location: context detail page (likely `frontend/src/app/contexts/[id]/page.tsx` and a new `TagCloud.tsx` / `TagAutocomplete.tsx` component)
- Design discussion needed: visual treatment of the tag cloud (font sizing function, color palette tied to count, dark mode)
- i18n: tag labels are user-generated content (no translation needed); UI chrome strings need i18n

## Acceptance

- [ ] Design discussion captured (mockup or written spec for visual treatment)
- [ ] `TagCloud` component on context detail page rendering top tags by count
- [ ] `TagAutocomplete` component for tag input fields (debounced, prefix-filtered)
- [ ] Click-tag-to-filter wired into the existing memory list filter
- [ ] Loading + empty states (200 + `tags=[]` from server)
- [ ] Dark mode coverage, responsive (`sm:` / `md:`)
- [ ] Frontend tests (Vitest) for both components

## Related

- Backend: #614 (issue), #617 (PR merged `e80c4a8`)
[...]

PR #833, opened 2026-05-30
feat(frontend,api): tag-selection facet drill-down on context tag cloud (#830)

### LINKED ISSUE #830: feat(frontend,api): tag-selection facet drill-down on context tag cloud (#618 follow-up)
## Summary

Follow-up to #618 (tag cloud + autocomplete on context detail). Today the tag cloud
**facets on the free-text `q` search** (`TagCloud` `q` prop → `GET /contexts/{id}/tags?q=`)
but **clicking a tag does not facet the cloud** — it only toggles the memory-list filter.
This asymmetry means you can't drill down: selecting a tag should narrow the cloud to the
tags that co-occur with it, the same way `q` already narrows it.

This issue makes tag selection facet the cloud too, as a **multi-tag AND drill-down**.

## Behavior

Symmetric with `q` faceting, AND-combined with it:

1. Click a cloud tag → it moves to a **selected-tags chip row**; the cloud re-facets to
   "tags on memories that have **all** currently-selected tags" (the selected tags
   themselves drop out of the cloud body — the cloud always shows "what else can I add").
2. Click more tags to AND further. Remove individual tags via the chip `×`.
   Show a "clear all" affordance when ≥1 tag is selected.
3. The memory list filters by the same selected set with **ALL-match** semantics, AND-ed
   with `q`.
4. When no co-occurring tags remain, the cloud shows an empty state
   ("no further tags to refine by").
5. URL stays shareable: `?q=...&tags=a&tags=b` (reuse the existing repeatable `tags` param).

## Backend

- **`GET /contexts/{id}/tags`**: add repeatable `with_tags` query param. When set, aggregate
  only over memories whose tags contain **all** of `with_tags` (`tags @> ARRAY[...]`), and
  **exclude** the `with_tags` values from the returned tags. Combine with existing `q` (AND).
  Counts reflect the faceted (co-occurrence) subset.
  - `context_service.aggregate_tags(...)` gains `with_tags: list[str] = []`
    (`backend/src/services/context_service.py:1099`+).
- **`GET /memory/list`**: add `tags_match` param (`any` | `all`), **default `any` to preserve
  the #618 behavior** (`backend/src/api/routes/memory.py`; client comment
  `frontend/src/lib/api/memory.ts:27`). `all` → memories holding every given tag.

## Frontend

- `MemoriesTabPanel`: `activeTag: string | null` → `selectedTags: string[]`; multi-tag URL sync;
  memory-list fetch uses `tags=selectedTags, tags_match="all"`.
- `SelectedTagChips`: **generalize the existing single-chip block**
  (`MemoriesTabPanel.tsx:240-256`, `Badge variant="secondary"` + `X` + `filter.clearTag`) into
  a list — not a net-new pattern (same shape as `FieldDefinitionInput.tsx:255-264`).
- `TagCloud`: accept `selectedTags`, pass as `with_tags`; `onTagClick` is now additive
  (selected tags never appear in the cloud). Add empty state.
- `lib/api/contexts.ts` (`withTags?`), `lib/api/memory.ts` (`tags_match?`); i18n en/ja for
  chips / empty / clear-all.

## UX / accessibility requirements (from /ask CDO review)

These are **part of the spec, not optional polish**:

- **Live region**: announce facet changes via `aria-live="polite"` on chip add/remove
  (e.g. "filtered by 'X', N tags remaining") — the cloud silently re-rendering is invisible
  to screen readers.
- **Focus management**: after removing a chip via `×`, move focus deterministically
  (next chip, else the search input). Define this; the existing single chip doesn't handle it.
- **`aria-pressed` cleanup**: since selected tags leave the cloud, `TagCloud`'s `aria-pressed`
  becomes always-false — remove it or document why it stays.
- **Stable size scale**: `TagCloud.sizeClass` recomputes min/max from the *displayed* set, so
  the same tag's font size jumps on every drill-down, eroding the "size = magnitude" signal.
  Fix the scale to a stable basis (or weaken size encoding during drill-down).
- **State legibility**: show a "filtering N memories" count so the additive (not replace)
  mental model is clear — selecting a tag accumulates, it doesn't swap.

## Testing

- Backend: `aggregate_tags` `with_tags` ALL-match + self-exclusion + `q` combination
  (extend `backend/tests/integration/test_list_tags_aggregation.py`); `/memory/list`
  `tags_match=all` with a `default=any` regression pin (extend
  `backend/tests/api/test_memory_list.py`).
- Frontend: TagCloud facets by `selectedTags`; chip removal; multi-tag URL sync; empty state;
  live-region announcement.

## Out of scope (YAGNI)

- `TagAutocomplete` (memory-edit input surface, not the filter cloud).
- OR-mode toggle; tag hierarchy.

---

Parent: #618. Design brainstormed + CDO-reviewed in session 2026-05-29.

JFK:
Closes #830

## Summary
- Follow-up to #618. Clicking a tag in the cloud now **facets the cloud** (multi-tag AND drill-down) symmetric with the `q` search, instead of only toggling the memory list. Selected tags become a chip row; the cloud re-facets to tags co-occurring on memories that have **all** selected tags, and the selected tags drop out of the cloud ("what else can I add").
- The memory list filters by the same set with **ALL-match**, AND-ed with `q`. URL is shareable: `?q=…&tags=a&tags=b`.

## Backend
- `ContextService.aggregate_tags` gains `with_tags`: narrows the memory set with `tags @> with_tags` (GIN-indexed) and self-excludes those tags from the result, AND-combined with `q`; counts reflect the faceted subset. Bound as a single `varchar[]` param (never interpolated); capped at 50 tags / 200 chars each. **Empty list is a deliberate no-op (== #618).**
- `GET /contexts/{id}/tags` gains repeatable `with_tags`.
- `GET /memory/list` gains `tags_match` (`any|all`, **default `any` preserves #618**). One shared `tag_filter` predicate is applied to **both** the data and count queries (replacing a duplicated predicate).

## Frontend
- `TagCloud`: `activeTag` → `selectedTags` (sent as `with_tags`); click is additive; `aria-pressed` removed (selected tags leave the cloud); **stable absolute-count size buckets** (no font-jump on drill-down); drill-down-aware empty state.
- `MemoriesTabPanel`: single `?tag=` → repeatable `?tags=`; selected-tags chip list with per-chip remove + clear-all; ALL-match list fetch; **a11y**: `aria-live` announcements, deterministic focus management after chip removal, "filtering N" count.
- API clients (`withTags` / `tagsMatch`); i18n en + ja.

## UX / a11y (from the issue's CDO review — treated as spec)
aria-live facet announcements · focus moves to next chip (else search input) on remove · aria-pressed removed · stable size scale · "filtering N memories" count.

## Tests
- Backend: `with_tags` ALL-match + self-exclusion + `q`-combo + no-co-occurrence + 50-cap (real-DB integration); `tags_match=all` (`@>`) + `default=any` (`&&`) regression pin (unit).
- Frontend: `selectedTags` → `with_tags`, `aria-pressed` absent, drill-down empty state, multi-tag URL sync, de-dupe/blank drop, chip removal → reduced URL.
- 26 backend + 23 frontend tests pass. ruff / pyright / tsc clean.

## Pre-PR review summary
- gate2 mode: advisor-only — cso: green · qa-lead: green · cto: green
- review provider: code-review (2 independent review agents → no real bugs; 3 low-severity polish items applied: doc drift, count flicker, untrimmed-q empty-state)

### DIFF
--- a/backend/src/api/routes/contexts.py
+++ b/backend/src/api/routes/contexts.py
@@ -751,6 +751,16 @@ async def list_context_tags(
             "treated as no filter."
         ),
     ),
+    with_tags: list[str] = Query(
+        default_factory=list,
+        description=(
+            "Repeatable tag filter for multi-tag AND drill-down (#830). When "
+            "set, only tags that co-occur on memories carrying ALL of these "
+            "tags are aggregated, and the with_tags values themselves are "
+            "excluded from the result. AND-combined with `q`. Counts reflect "
+            "the faceted (co-occurrence) subset."
+        ),
+    ),
 ) -> ContextTagsResponse:
     """List existing tag vocabulary in a context (Issue #614).
 
@@ -783,6 +793,7 @@ async def list_context_tags(
             sort=sort,
             prefix=prefix,
             q=q,
+            with_tags=with_tags,
         )
     except NotFoundException as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e)) from e
--- a/backend/src/api/routes/memory.py
+++ b/backend/src/api/routes/memory.py
@@ -539,6 +539,13 @@ async def list_memories(
         "Repeat the param to pass several: ?tags=a&tags=b. Combined with other "
         "filters (e.g. q) by AND. Blank / whitespace-only entries are ignored.",
     ),
+    tags_match: str = Query(
+        "any",
+        pattern="^(any|all)$",
+        description="How to combine `tags`: `any` (default — memory has at least "
+        "one of the tags, PG array overlap; preserves #618 behavior) or `all` "
+        "(memory holds every given tag, PG array contains — #830 drill-down).",
+    ),
     limit: int = Query(50, ge=1, le=500),
     offset: int = Query(0, ge=0),
 ):
@@ -642,10 +649,23 @@ async def list_memories(
         if q_pattern is not None:
             query = query.where(Memory.summary.ilike(q_pattern, escape="\\"))
 
-        # ANY-match: row has at least one of the requested tags (PG array
-        # overlap ``&&``). NULL-tags rows never overlap, so they're excluded.
+        # Tag filter (#618 ANY / #830 ALL). Built once and applied to BOTH the
+        # data and count queries so the total always reflects the same filter
+        # (the two used to carry duplicate predicates — easy to let drift).
+        # NULL-tags rows never match either operator, so they're excluded in
+        # both modes.
+        tag_filter = None
         if tags_normalized is not None:
-            query = query.where(Memory.tags.overlap(tags_normalized))
+            if tags_match == "all":
+                # ALL-match (#830): row's tags contain every requested tag
+                # (PG array contains ``@>``).
+                tag_filter = Memory.tags.contains(tags_normalized)
+            else:
+                # ANY-match (default, #618): row has at least one of the
+                # requested tags (PG array overlap ``&&``).
+                tag_filter = Memory.tags.overlap(tags_normalized)
+        if tag_filter is not None:
+            query = query.where(tag_filter)
 
         # Get total count (with same filters as data query)
         count_query = select(func.count(Memory.id)).where(Memory.deleted_at.is_(None))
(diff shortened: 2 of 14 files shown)


JFK:
Both addressed in 9254555c:
1. `aggregate_tags` now binds the **stripped** tag value, so `?with_tags=%20python%20` matches memories tagged `python` and self-excludes it (added a whitespace regression test).
2. Updated `test_context_tags::test_passes_through_query_params` to include the new `with_tags` param in the asserted `aggregate_tags` call (CI would have failed otherwise — thanks for catching).

Issue #1093, opened 2026-06-27
feat(billing): owner-only handoff endpoint minting a short-lived signed token


JFK:
memory-cloud already exposes the Stripe-agnostic internal entitlement endpoint (`PUT /internal/workspaces/{id}/plan`). The external billing service needs an authenticated workspace **owner** to start a billing session without re-implementing user auth on the billing host.

## Proposal
- Add `POST /api/v1/billing/handoff` — **session-auth, owner-only** (reject admin/member and API-key auth) — that mints a short-lived **Ed25519-signed** token bound to `(user_id, workspace_id, role=owner)` with `iss` / `aud` / `exp` and a one-time `jti`. memory-cloud holds the private signing key; the billing service verifies with the public key (`kid`-based rotation).
- Ensure all billing-mutating endpoints are owner-only.

## Acceptance
- Owner-only enforced (403 for admin/member/API-key).
- Token carries the `(user_id, workspace_id, role)` binding + `iss/aud/exp/jti`.
- Signing key configurable; rotation via `kid`; fail-closed when unset.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #389, opened 2026-04-19
fix(auth): restrict /workspace/resources to workspace owner


JFK:
## Summary

The `/workspace/resources` page is currently visible to any workspace member. Resources should be restricted to the workspace **owner** only — `admin` / `member` / `viewer` roles should not see the page or its navigation entry, and direct-URL access should redirect.

Rationale: Resources carry credential-issuing capability (resource tokens used for ingest) and schema decisions that shape the data pipeline. This places them in the same risk class as external API keys (#381 / #385), which are already owner-only. Defaulting to least privilege is the safer posture; `admin` read-only access can be opened later if a concrete need arises.

Filed as a **v0.12.2 hotfix** because the exposure is live in production (metadata reconnaissance surface) and waiting for v0.13 is not justified given the small, low-risk fix size.

## Context

Steps to reproduce:
1. Log in as a user whose workspace role is `member` (or `admin` / `viewer`).
2. Navigate to `/workspace/resources` on any Kagura Memory Cloud instance.
3. Observed: the page renders and lists workspace resources.
4. Expected: direct-URL access redirects to the dashboard, and the nav entry is hidden.

Related work:
- #381 / #385 — external API keys moved to owner-only Settings (same risk class, same pattern)
- #365 — hide admin-only context settings from non-admin members (prior-art for frontend role gating)
- #382 — gate empty-state context-create CTA by workspace admin role (role-gating UX pattern)
- #316 — feat: resource event data browser (had an unresolved "authorization surface" open question that this issue settles in favor of owner-only)
- **#390** — follow-up: close CWE-639 on resource_* satellite reads (resource_pk enforcement + Phase 2 writer migration). Separate v0.13.0 issue — not addressed here per gate1 scope-split decision.

Existing structure to build on:
- `currentWorkspace.current_user_role` is already exposed to the UI (values: `owner` / `admin` / `member` / `viewer`). The same check pattern used in #365 / #382 can be reused.
- `/workspace/resources` already has a plan gate (free / basic plans see a CTA). The role gate layers **on top of** this plan gate; plan-gate behavior is unchanged.
- Backend already has a `WorkspaceOwner` dependency type (`backend/src/auth/dependencies.py` — `require_workspace_owner`) that returns **403 Forbidden** on role deny. Already used by `resource_tokens.py`.
[...]

PR #1099, opened 2026-06-27
feat(workspace): owner-initiated ownership transfer + epoch (#1094)

### LINKED ISSUE #1094: feat(workspace): owner-initiated ownership transfer + ownership epoch
Ownership is a single `owner_user_id` per workspace, and there is no voluntary transfer path today (only account-erasure auto-transfer). Owner-gated billing/admin actions make a sole owner a single point of failure.

## Proposal
- Add `POST /api/v1/workspaces/{id}/transfer-ownership` (current-owner-only, session-auth): target must be an existing member; atomic single-owner invariant (set owner / promote target / demote previous owner); audit log; idempotent; notify the new owner.
- Add an **ownership epoch/version** bumped on transfer or demotion, so external sessions bound to the previous owner can be invalidated.
- Break-glass admin force-transfer with mandatory audit (and optional dual-control), handling the zero-member case.

## Acceptance
- Single-owner invariant preserved under concurrency (row lock).
- Audit row written; ownership epoch bumped on success.

JFK:
Closes #1094

## Summary
- Add `POST /api/v1/workspaces/{id}/transfer-ownership`: **current-owner-only + session-only** (a Bearer API key / OAuth token is rejected 403), transferring ownership to an existing member.
- `WorkspaceOwnershipService.transfer_ownership` locks the workspace row (`SELECT ... FOR UPDATE`) as the single-owner synchronization point, **re-verifies the caller is still owner under the lock** (closes the route-check→lock TOCTOU → 409), is **idempotent** (no-op when the target already owns it — no epoch bump, no audit), promotes target → OWNER, demotes previous owner → ADMIN (skipping a missing legacy member row), bumps `ownership_epoch`, and writes an audit row in the **same transaction**.
- New `workspaces.ownership_epoch` column (Integer, NOT NULL, server_default 0) + migration `e45` (down_revision `e44_1065_host_arbitration`, linear head). The epoch is the **producer** side; the invalidation consumer is a documented follow-up.
- Owner gate binds to the explicit **path** `workspace_id` (#389 cross-tenant guard); target-not-member → 400.

## Implementation notes
- New `backend/src/services/workspace_ownership_service.py` (dedicated transactional service); route added to the existing `workspaces` router.
- Reuses `WorkspaceRole` / `check_workspace_owner` / `AuditLog`; owner user_ids are stored in the audit `user_metadata` (JSON), not the `String(64)` hash columns.
- **18 tests**: 11 route RBAC/mapping (DB-free) + 7 service integration (real DB) including a **two-session concurrent-transfer test** proving the row lock keeps a single OWNER + one epoch bump (the headline acceptance criterion).

## Pre-PR review summary
- gate2 mode: advisor-only · audit: skipped · binary_gate: (none)
- cso: green · qa-lead: green · cto: green
- gate1: yellow via /claude-c-suite:ask (3 constraints pinned + implemented: under-lock TOCTOU re-check, missing-prev-member, accurate invariant docstring)
- review provider: none (post-PR review opt-in)
- An independent `/code-review max` pass ran during implementation and caught + fixed a real audit-column truncation bug (owner user_ids in `String(64)` → moved to `user_metadata`) and the missing concurrency test.

## Scoped as follow-ups (beyond the stated acceptance)
- Break-glass admin force-transfer + dual-control + zero-member case.
- New-owner notification.
- The ownership-epoch invalidation consumer (e.g. invalidating handoff tokens from #1093 bound to a previous owner).
- Unifying all owner-mutating paths (incl. the account-erasure auto-transfer) under one workspace row lock.

### DIFF
--- a/backend/alembic/versions/e45_1094_workspace_ownership_epoch.py
+++ b/backend/alembic/versions/e45_1094_workspace_ownership_epoch.py
@@ -1,0 +1,32 @@
+"""Add workspaces.ownership_epoch (#1094).
+
+Issue #1094: owner-initiated ownership transfer needs a monotonic per-workspace
+version that bumps on each transfer, so external sessions / tokens bound to a
+previous owner can be invalidated by a downstream consumer. NOT NULL with a
+``0`` server_default so every existing row backfills to epoch 0 without a data
+migration.
+
+Revision ID: e45_1094_ownership_epoch
+Revises: e44_1065_host_arbitration
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e45_1094_ownership_epoch"
+down_revision = "e44_1065_host_arbitration"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "workspaces",
+        sa.Column("ownership_epoch", sa.Integer(), nullable=False, server_default="0"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("workspaces", "ownership_epoch")
--- a/backend/alembic/versions/e45_1094_workspace_ownership_epoch.py
+++ b/backend/alembic/versions/e45_1094_workspace_ownership_epoch.py
@@ -8,6 +8,10 @@ migration.
 
 Revision ID: e45_1094_ownership_epoch
 Revises: e44_1065_host_arbitration
+
+Note: the revision id is kept <=32 chars (shorter than the filename stem) for
+the ``alembic_version.version_num`` VARCHAR(32) column — same convention as the
+sibling capped migrations.
 """
 
 import sqlalchemy as sa
--- a/backend/src/api/routes/workspaces.py
+++ b/backend/src/api/routes/workspaces.py
@@ -14,14 +14,16 @@ from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth.analysis_allowlist import check_workspace_in_allowlist
-from auth.dependencies import get_current_user
+from auth.dependencies import SessionUser, get_current_user
 from auth.workspace_roles import WorkspaceRole
 from db.base import get_db
 from models.api_base import TZAwareBaseModel
 from models.auth import Context, ExternalAPIKey, User
 from models.schemas import OpenAIKeyStatusResponse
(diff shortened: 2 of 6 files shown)


JFK (commented):
## `/code-review max` — independent re-review (post-fix)

**No correctness bugs.** The transactional core (workspace `FOR UPDATE` + under-lock owner re-check, atomic demote/promote/epoch/audit) verified clean. Findings are cleanup / coverage / already-tracked follow-ups.

### Already tracked as follow-up #1102 (unify owner-mutating paths under one lock)
- **`ownership_epoch` is producer-only and the erasure path doesn't bump it** — `account_erasure_service` transfers ownership without bumping `ownership_epoch` / taking the lock / writing the audit row. Latent: harmless until the epoch **consumer** (#1100) ships, at which point an erasure-driven transfer would leave a stale epoch. Acknowledged in the service docstring.
- **member-row SELECT isn't under `FOR UPDATE`** — a concurrent `update_member_role` / `remove_member` (neither takes the workspace lock) can race the target's role write. Transfer-vs-transfer is fully serialized by the workspace lock; cross-path member mutations are the gap.

### Worth a quick cleanup
- **`(workspace.ownership_epoch or 0) + 1`** — `ownership_epoch` is NOT NULL with `server_default "0"` and the row is always DB-loaded here, so `or 0` is dead. Simplify to `+ 1`.
- **`user_metadata["performed_by"]` duplicates `AuditLog.user_id`** (both = `current_owner_id`). Drop the metadata key; no test asserts it.

### Minor / contract nits
- The "legacy previous-owner without a member row" branch is **unreachable via the route**: `check_workspace_owner` resolves ownership via the member row, so such an owner gets 403 before the service runs. The branch + its integration test only exercise direct service calls — drop the branch or document it as direct-call-only.
- Unknown/soft-deleted workspace id returns **403, not the service's 404** (`check_workspace_owner`'s CWE-639 enumeration defense fires first) — intended, but the route test's 404 case only hits via a mocked service. Document the 403 contract.
- Redundant pre-lock query: `check_workspace_owner` loads workspace+member, then the service re-loads under the lock (kept for the 403-vs-409 distinction).
- Audit row omits `ip_address`/`user_agent` (route doesn't thread `Request`) — forensic gap vs the erasure audit trail; codebase is inconsistent here (member_credentials also omits).

_Posted via `/code-review max --comment`. CI is green; these are non-blocking._

PR #1148, opened 2026-06-30
fix(plan): admin Plan page — actor link + quota usage/effective order

JFK:
## Problem

On the admin **Quota/Limits → 監査ログ (audit log)** page, the **変更者 (changed by)** column showed an ambiguous display name (e.g. literally `admin`) instead of identifying who made the change, and it was not clickable.

## Root cause

`GET /api/v1/admin/plans/audit` joined `PlanChange.changed_by` (the actor's `user_id`) to **`User.name`** and overwrote `changed_by` with that label — so the UI had no id to link, and a name like "admin" was uninformative. (The codebase's other audit surface, `AuditLog`, already keys on `user_email`.)

## Fix

Resolve the actor the **same way the admin user pages do** — `User.name` + `User.email` ([admin.py](https://github.com/kagura-ai/memory-cloud/blob/main/backend/src/api/routes/admin.py#L479)) — and **keep the raw `user_id` in `changed_by`** so the UI can link the cell to `/admin/users/{user_id}`.

The 변更者 cell now renders **name over email** (matching the admin users list style), linked to the user-detail page. Actors that no longer resolve (erased → pseudonymized `changed_by`, null name/email) render the raw id as **plain text — no broken link**.

```
Fumikazu Kiyota        ← User.name
kiyota@noizz.co.jp     ← User.email      (whole block links to /admin/users/{user_id})
```

## Changes

- **backend** (`admin_plans.py`): `PlanChangeAuditEntry` gains `changed_by_name` / `changed_by_email`; `changed_by` stays the raw `user_id`. Audit query selects `User.name, User.email`.
- **frontend** (`admin.ts`, `admin/plans/page.tsx`): type fields + 2-line linked actor cell.
- **tests**: backend actor-resolution integration test (`test_admin_plan_audit_actor.py` — name+email + null fallback); frontend audit-tab link tests (name+email / email-only / unresolved).

## Verification

- frontend: `admin/plans/page.test.tsx` **16/16 pass**, `tsc --noEmit` clean
- backend: `py_compile` OK; new integration test runs in CI (`backend-integration`). Existing `test_admin_plans_soft_delete_filter.py` pin test asserts `workspace_id` only — unaffected.

## Note

For an existing audit row to render the linked name+email, its stored `changed_by` must resolve to a current `User` (it does for session-admin changes, which record the real OAuth sub). Rows whose `changed_by` does not resolve degrade gracefully to plain text.

### DIFF
--- a/backend/src/api/routes/admin_plans.py
+++ b/backend/src/api/routes/admin_plans.py
@@ -225,7 +225,14 @@ class PlanChangeAuditEntry(BaseModel):
     workspace_name: str
     old_plan: str | None
     new_plan: str
+    # `changed_by` is the actor's stable user_id (OAuth2 sub) as stored — the UI
+    # links it to /admin/users/{changed_by}. `changed_by_name` / `changed_by_email`
+    # are the resolved display labels (name over email); both are None when the
+    # actor no longer resolves to a User row (e.g. account erased → pseudonymized
+    # changed_by), in which case the UI shows the raw id without a link.
     changed_by: str
+    changed_by_name: str | None = None
+    changed_by_email: str | None = None
     changed_at: str
     reason: str | None
 
@@ -558,7 +565,7 @@ async def get_plan_change_audit(
         # that have since been soft-deleted must remain visible so admins can
         # reconcile past billing / plan transitions.
         audit_result = await db.execute(
-            select(PlanChange, Workspace.name, User.name)
+            select(PlanChange, Workspace.name, User.name, User.email)
             .join(Workspace, PlanChange.workspace_id == Workspace.id)
             .outerjoin(User, PlanChange.changed_by == User.user_id)
             .order_by(PlanChange.changed_at.desc())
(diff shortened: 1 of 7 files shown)


JFK:
Added a second admin Plan-page fix on this branch (commit c558499d):

**Quota numerator/denominator order** — the workspaces tab's expanded quota detail showed each row as `{effective} / {usage}` (e.g. memory `100,000 / 5,220`), the reverse of the summary row `{used} / {limit}` (`5,220 / 100,000`). Swapped the detail cell to `{usage} / {effective}` and renamed the column header `quota.effectiveUsage` → `quota.usageEffective` (`使用量 / 有効値` / `Usage / Effective`) so the fraction reads used-over-limit consistently. Rows with no tracked usage still show the effective value alone. Regression test added (memory row renders `50,000 / 110,000`, not the reverse).

Verification: `admin/plans/page.test.tsx` **17/17 pass**, `tsc --noEmit` clean.

Issue #1167, opened 2026-07-02
feat(config): ENABLE_BYOK flag — gate BYOK surface + reorder workspace menu


JFK:
## Summary

Add an `ENABLE_BYOK` env flag (default **true**) that lets a deployment turn off the BYOK (bring-your-own-key) surface. BYOK stays fully available for OSS/self-hosted by default; setting `ENABLE_BYOK=false` hides and blocks the external-API-keys console and the workspace cost dashboard. Also reorders the Workspace sidebar group so operational stores come first and reporting comes last.

## Motivation

- Managed / locked-down deployments may not want workspace owners registering their own provider keys; today there is no switch for that.
- When BYOK is off, the workspace cost dashboard (which tracks BYOK-driven LLM/embedding spend) is noise — hide it and block its API.
- Follows the established deployment-toggle pattern: `ENABLE_PLAN_PAGE` → `/system/info features.plan_page` → sidebar `requiredFeature` + page gate (#1145).

### Backend
- `settings.py`: `enable_byok: bool = Field(default=True)` (env `ENABLE_BYOK`).
- `GET /api/v1/system/info`: add `features.byok`.
- `external_keys.py`: router-level dependency returning **404** when disabled (feature-not-present semantics, consistent with plan_page; 403 would read as a permission problem).
- `cost_aggregation.py`: same 404 guard on `GET /workspaces/{id}/cost-aggregation` only. `GET /admin/cost-aggregation` stays ungated (platform env-key usage still accrues cost).

### Frontend
- `Sidebar.tsx`: `requiredFeature: "byok"` on the `externalKeys` and workspace `cost` nav items; skip the owner-only `listExternalAPIKeys` warning-icon fetch unless `features.byok`.
- `external-keys/page.tsx` + `workspace/cost/page.tsx`: page-level gate mirroring the plan page — spinner while features resolve, "not available in this environment" notice when off, data fetch only after the flag resolves enabled. i18n (en/ja).

### Menu reorder (same PR — touches the same `navigationGroups` array)
Workspace group order becomes: dashboard → contexts → resources → storage → secrets → members → sleepReports → cost.

## Semantics note

The LLM key-resolution layer is untouched: keys registered **before** turning the flag off are still resolved and used. Only create/read/update/delete of keys (and the cost dashboard) are blocked. A deployment that starts with `ENABLE_BYOK=false` never accumulates keys, so BYOK is effectively fully disabled there.

## Acceptance criteria

[...]

Issue #1165, opened 2026-07-02
feat(credentials): owner-key provisioning of member API keys (mint/list/revoke)


JFK:
## Summary

Member API-key endpoints (`member_credentials.py`) are `SessionUser`-gated **and strictly self-only** today — mint (`:276`) raises 403 `"Only owner can create API keys"` unless `path user_id == session user`, and the per-id DELETE (`:530`) is likewise self-only (`check_can_manage` deliberately excludes `create` for owner/admin). So an owner cannot provision CI/service-account keys headlessly, and cannot mint for members at all.

This issue adds a **new authorization capability**: workspace-owner-provisioned minting/listing/revocation of **other members'** keys, callable with the owner's API key. Session self-mint/self-delete semantics stay unchanged, and `api_keys.py` self-mint stays session-only — **decision #252 is NOT relaxed**.

Unblocks kagura-ai/kagura-memory-python-sdk#201 (`kagura auth create-key` family). Facts verified against `origin/main` @ `fe6d7376`.

### Auth & gating (same framework as the member-management issue)
- Endpoints move to `APIKeyOrSessionUser`; **OAuth Bearer → 403** (a `memory:write` device token must not mint long-lived credentials).
- Programmatic principal must pass `PermissionService.check_workspace_owner(user_id, path_workspace_id)` — in-handler against the **path** workspace (`require_workspace_owner` reads only `current_workspace_id` and is the wrong gate for a path-parameterized route).
- Workspace-scoped key whose bound workspace ≠ path workspace → uniform 404 (#963 pattern).
- Principal discrimination: shared helper, key-presence tests (`"api_key_workspace_id" in user` / `"oauth_scope" in user` / `"sub" in user`), fail closed.

### Anti key-self-replication (the #252 threat, made explicit)
- Programmatic mint **403 when target `user_id == caller user_id`** — a leaked owner key must not be able to self-replicate fresh keys and defeat revocation.
- Programmatic mint **403 unless the target member's workspace role is `member` or `viewer`** — owner-key minting is strictly privilege-**downgrade** provisioning for service identities (minted keys carry no role of their own: they authenticate as the target user with whatever workspace role that user holds at request time, `_build_api_key_user_dict` hardcodes platform `role: "user"`).

### Expiry (new field — required)
[...]

Issue #1164, opened 2026-07-02
feat(workspaces): owner-key programmatic access to member + invitation endpoints


JFK:
## Summary

Workspace member/invitation management is session-cookie-only today (`get_current_user` reads `request.state.user`, set only by the `kagura_session` cookie middleware), so the Python SDK / CLI cannot automate it. Open these endpoints to programmatic auth — **API keys only, gated at workspace-OWNER role** — keeping session (web UI) semantics identical.

Unblocks kagura-ai/kagura-memory-python-sdk#225 (`kagura workspace member` / `kagura workspace invite` CLI). All facts below verified against `origin/main` @ `fe6d7376`; a cross-repo review log lives in the SDK plan (`docs/superpowers/plans/2026-07-02-issue-201-225-owner-key-plan.md`).

## Endpoints in scope (8)

| Method | Path | Session gate today (unchanged) |
|---|---|---|
| GET | `/api/v1/workspaces/{workspace_id}/members` | member+ (`workspaces.py:574`) |
| POST | `/api/v1/workspaces/{workspace_id}/members` | admin+ (`workspaces.py:692`) |
| PUT | `/api/v1/workspaces/{workspace_id}/members/{user_id}` | admin+ with #254 guards (`workspaces.py:724`) |
| DELETE | `/api/v1/workspaces/{workspace_id}/members/{user_id}` | owner (#217, `workspaces.py:842`) |
| POST | `/api/v1/workspaces/{workspace_id}/invitations` | admin+ (`invitations.py:47`) |
| GET | `/api/v1/workspaces/{workspace_id}/invitations` | admin+ (`invitations.py:173`) |
| DELETE | `/api/v1/workspaces/{workspace_id}/invitations/{invitation_id}` | admin+ (`invitations.py:241`) |
| GET | `/api/v1/workspaces/{workspace_id}/member-quota` | member+ (`invitations.py:500`) |

**Explicitly NOT changed:** `POST /api/v1/invitations/accept`, `GET /api/v1/invitations/pending`, and the public token preview `GET /api/v1/invitations/{token}` (pre-auth, `main.py:465`).

### Auth modes
- **Session principal: behavior unchanged** (roles per table above; #254 self-change/owner-change guards intact).
[...]

Issue #1160, opened 2026-07-02
refactor: rename provider key ollama → self_hosted (vLLM-ready)


JFK:
## Summary

Rename the self-hosted inference provider key `ollama` → `self_hosted` across all four surfaces where it is persisted or compared — LLM provider registry (`_PROVIDERS`), reranker (`context_search_configs.reranker_provider` CHECK/enum), embeddings (`EMBEDDING_PROVIDER`), and sleep (`SLEEP_LLM_PROVIDER`) — and normalize the provider transport to pure OpenAI-compatible endpoints so vLLM (and any OpenAI-compatible backend) works without introducing an engine subfield.

**Motivation**: `ollama` is an engine name occupying what is actually a *self-hosted inference backend* slot. Deploying on a pgx/spark cluster requires vLLM (Ollama is not cluster-capable). Today three Ollama-native calls block any non-Ollama backend: the health check `GET /` (expects Ollama's root banner; vLLM roots 404), `GET /api/tags` (model listing), and the reranker's `POST /api/generate` prompt-scoring.

## Design decisions

| # | Decision |
|---|---|
| D1 | Unify the key as `self_hosted` on all four surfaces (LLM / reranker / embeddings / sleep). A partial rename would leave two conventions for the same slot. |
| D2 | `ollama_cloud` is **out of scope** — a distinct hosted SaaS (ollama.com). `WHERE provider='ollama'` exact match cannot touch it. `OLLAMA_API_KEY` keeps its name. |
| D3 | Normalize transport to OpenAI-compatible only (no engine subfield): health `GET /` → `GET /v1/models`; list_models `/api/tags` → `/v1/models`; reranker prompt-scoring `/api/generate` → `/v1/completions` (same prompt, `max_tokens=5`, stop on newline). Add optional `SELF_HOSTED_API_KEY` (vLLM `--api-key`; today `api_key="ollama"` is hardcoded). Ollama and vLLM both serve all of these endpoints. |
| D4 | Env clean cutover: `OLLAMA_BASE_URL` → `SELF_HOSTED_BASE_URL` (settings field `ollama_base_url` → `self_hosted_base_url`). No read alias; log a startup **warning** when the legacy var is set (misconfiguration detection — value not honored). |
[...]

PR #1176, opened 2026-07-03
fix: resolve all v0.42.0 max code-review findings

JFK:
## Summary

Resolves **all findings** from the v0.42.0 max code-review (range `v0.41.0..main`, 39 CONFIRMED). Includes a **release blocker**: every API-key-authenticated request 500s on `main`. 15 commits, ~40 files; correctness fixes, three design-level fixes, the reuse/simplify/efficiency cleanups, and the conventions doc fix — each with tests.

## Blocker
**`VerifiedKey.key_prefix`** — `dependencies.py` reads `result.key_prefix` (added for #1164 audit attribution), but `VerifiedKey` never carried it → `AttributeError` → **HTTP 500 on every REST + remote-MCP request authenticated with an API key**. Route tests missed it by stubbing the auth dependency with plain dicts. Must land before tagging v0.42.0.

## Correctness / security
- **ollama→self_hosted (#1160):** `SLEEP_LLM_PROVIDER=ollama` crashed *every* `NeuralMemoryConfig`; `EMBEDDING_PROVIDER=ollama` silently routed to OpenAI — both coerced to `self_hosted` with a boot warning.
- **Reranker (#1161):** self-hosted default model is an Ollama id that 404s on vLLM (added `SELF_HOSTED_RERANK_MODEL` + re-resolution on the `/v1/rerank` path); `VLLMReranker` sent no auth (added `RERANK_API_KEY`); telemetry now accounts for `RERANK_BASE_URL`.
- **Member credentials (#1165):** soft-revoke now drops the at-rest plaintext (sweeper skips revoked rows — was Fernet-decryptable forever); an already-revoked forensic row is 404 for any further delete/revoke (a member could hard-delete the evidence; a repeated revoke duplicated the audit row); non-member target → 404 not 500; `expires_at` is now surfaced; fail-closed principal guards.
- **Admin/auth:** `delete_user` file-purge race (private-file widening) closed; the `#963` confinement's doubled `"not found not found"` message fixed.
- **BYOK UI robustness (#1167):** `useSystemFeatures` retries transient `/system/info` blips (was fail-closing a default-on flag); `SearchSettingsSection` no longer blocks Save when BYOK-off hides key enumeration; contexts page double-fetch fixed.
- **Eval (placebo):** production `edge_gate` calibration snapshotted + restored (was clobbered during the run); rewire-swap metric counts accepted swaps, not edges.

## Design-level findings
- **#32 — BYOK depth:** `ENABLE_BYOK` now gates only the `/external-keys` **write** paths; list/toggle-off/delete stay reachable so a customer can disable/delete a key that is still being resolved and billed (the write gate runs before auth → no existence leak). Resolution stays untouched so live embeddings don't break on a flag flip.
- **#33 — owner-key scope:** added `enable_owner_key_member_management` (default on). When off, an owner's ordinary API key no longer carries member-management power — a deployment kill-switch for the stolen-owner-key hazard, pending per-key scopes.
- **#35 — `/system/info` auth:** the endpoint is test-enforced public and exposes only version + non-sensitive flags, so the code is correct-by-design; `security.md`'s auth allowlist now lists it (the omission was the actual conventions gap).

## Cleanup (reuse / simplify / efficiency, #17–#31)
All applied: `apply_zero_knowledge_hide` / `_minted_key_response` / `_reject_oauth` (member-credentials); `_bearer_headers` + `REMOTE_RERANKER_DEFAULT_MODELS` (reranker); module-level `_fetch_v1_models` (system, both consumers); `_reject_programmatic_owner_assignment` (workspaces); per-workspace usage upsert aggregation (file-storage); `verify_self_hosted_reachable` (provider + embedding); `cli/_env_probe` (both CLIs); shared `provision_eval_context` (3 eval runners); `FeatureDisabledNotice` (both BYOK pages); a drift-guard test for the rerank-default literals (#25); the `soft_revoke` alias dropped (#29). #31 (per-call reranker httpx client) is documented as deliberate — a shared client would bind to one event loop and flake across test loops for a negligible single-request gain.

## Testing
- Backend: **355 tests** across auth / member-credentials / reranker / config / neural / eval / file-storage / cli / provider suites + the `account_erasure` integration test; `ruff` + `pyright` clean on all 18 changed source files.
- Frontend: **33 tests** across the BYOK consumer pages; `tsc --noEmit` clean.
- Added regression tests for each correctness fix (key_prefix, soft-revoke plaintext, forensic-404, non-member-404, vLLM auth header, ollama coercion, rewire-swap metric, BYOK write-gate split, owner-key kill-switch, rerank-default drift-guard).

### DIFF
--- a/.claude/rules/security.md
+++ b/.claude/rules/security.md
@@ -6,7 +6,11 @@ paths:
 # Security Rules
 
 ## Authentication
-- All API routes MUST use auth dependency (except `/`, `/health`, `/.well-known/*`, `/docs`, `/openapi.json`)
+- All API routes MUST use auth dependency (except `/`, `/health`, `/api/v1/system/info`, `/.well-known/*`, `/docs`, `/openapi.json`)
+  - `/api/v1/system/info` is intentionally public: version + non-sensitive
+    deployment feature flags the web UI reads before workspace context exists.
+    It exposes no user/workspace data. Enforced public by
+    `tests/api/test_system_info.py` + `tests/smoke/test_health.py`.
 - Session-only routes: `Depends(get_current_user)`
 - Session + API key routes: `Depends(APIKeyOrSessionUser)`
 - Admin routes: verify `system_admin` or `workspace_admin` role after auth
--- a/backend/src/api/routes/admin.py
+++ b/backend/src/api/routes/admin.py
@@ -1050,7 +1050,19 @@ async def delete_user(
             .scalars()
             .all()
         )
-        await db.execute(delete(Context).where(Context.created_by == user_id))
+        # Delete EXACTLY the contexts we just purged (the FOR UPDATE snapshot),
+        # not a fresh ``created_by == user_id`` predicate. Under READ COMMITTED
+        # the latter would also delete a context the user's own session created
+        # AFTER the snapshot (e.g. in a workspace they only belong to, not own):
+        # its files were never in ``ctx_ids`` so they skipped the purge, and a
+        # ``created_by`` DELETE would SET-NULL their context_id — widening a
+        # private file into a workspace-scoped object downloadable by any viewer,
+        # with its bytes never released. Scoping the DELETE to the purged
+        # snapshot keeps purge-set == delete-set; a context raced in after the
+        # snapshot simply survives with its files intact (created_by is a plain
+        # string, not an FK, so it does not block the user delete) (v0.42 review).
+        if ctx_ids:
+            await db.execute(delete(Context).where(Context.id.in_(ctx_ids)))
 
         # Remove user from workspace memberships
         # #687: operation-on-deleted — full user erasure must purge memberships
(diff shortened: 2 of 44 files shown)


JFK:
Thanks @copilot-pull-request-reviewer — all four comments were valid; addressed in aab84c0f:

1. **`SearchSettingsSection.tsx` — `externalKeysKnown` tied to `byokEnabled`.** Correct: `GET /external-keys` is workspace-owner-only, so it 403s for a non-owner editing search config even when BYOK is on. Now gated on whether the list call actually **loaded** (`externalKeysLoaded`), so a 403/404/network error leaves key presence "unknown" and never blocks Save or mislabels a working reranker.

2. **`external-keys/page.tsx` — full block contradicts the #32 backend semantics.** Fixed the real inconsistency: when BYOK provisioning is off the page now renders a **management-only console** (list + disable + delete, with a `provisioningDisabled` banner), and only the create/update affordances are hidden — so an owner can remove a still-billed key, matching the route change. Added `provisioningDisabled` i18n (en+ja) and updated the page test.

3. **`self_hosted_provider.py` — non-200 all reported as "not responding".** Now distinguishes 401/403 ("backend rejected the request — set/check `SELF_HOSTED_API_KEY`") from genuine unreachability.

4. **`_env_probe.py` — inaccurate Bandit `S310` justification.** Reworded: `base_url` is an operator-supplied config value (env/Docker), not a fixed internal URL; the suppression stands as a local bootstrap probe of the operator's own backend.

Verified: frontend `tsc` clean + 33 tests green (incl. the reworked BYOK-off management-console test); backend `ruff` clean + provider/embedding/cli suites green.

PR #1186, opened 2026-07-04
fix(api): WARN when neural_config DB row overrides SLEEP_LLM_* env (#1182)

### LINKED ISSUE #1182: fix(neural): NeuralMemoryConfig.from_db silently overrides SLEEP_LLM_* env with stale DB rows
## Problem

`NeuralMemoryConfig.from_db` prefers rows in the `neural_config` table over the container's `SLEEP_LLM_*` env vars, with no warning when they disagree.

Observed in the week1-derisk Day-5 eval campaign (eval repo `experiments/week1-derisk.md`, Day 5 mechanism-attribution note; surfaced during the #1177-era whole-branch review): a stale row (`sleep_llm_provider=openai`, `sleep_llm_model=gpt-5-nano`) overrode the rig's `SLEEP_LLM_*` env pointing at the self-hosted qwen3.5-9b endpoint. Every Sleep judge call failed (`llm_call_failures=5/5`) and the failure was invisible (see companion issue on `sleep_summary.ok` semantics).

The pgx01 row was fixed by a manual `UPDATE` — **the code path is unchanged and the same trap exists in any deployment where DB config drifts from env.**

## Proposed fix

1. Document the precedence explicitly (DB > env today).
2. On config load, if a DB row exists AND the corresponding env var is set AND they differ → emit a WARN log naming both values and which one won.
3. (Optional) an explicit `SLEEP_LLM_FORCE_ENV=1` escape hatch so operators can pin env over DB without surgery on the table.

## Acceptance

- Mismatch between `neural_config` row and `SLEEP_LLM_*` env produces a visible WARN at startup/first Sleep run.
- Precedence is documented.

Evidence: kagura-memory-eval `experiments/week1-derisk.md` Day-5; related: #1177.

JFK:
Closes #1182

## Summary
`NeuralMemoryConfig.from_db` silently preferred `neural_config` DB rows over the container's `SLEEP_LLM_*` env vars. In the week1-derisk Day-5 eval a stale row (`openai`/`gpt-5-nano`) overrode the rig's self-hosted endpoint and every Sleep judge call failed with zero signal.

## Changes (`backend/src/neural/config.py`)
- **Precedence documented** on `from_db`: DB > env; env fills what the DB doesn't set.
- **Mismatch WARN**: new `_resolve_sleep_llm` — when a DB row exists AND the env var is explicitly set AND they differ, emits structured `sleep_llm_config_mismatch` naming both values and the winner.
- **`SLEEP_LLM_FORCE_ENV=1` escape hatch**: pins env over DB for exactly the `sleep_llm_provider`/`sleep_llm_model` pair (no-op when the env var isn't explicitly set — defaults never override a DB row).

WARN cadence follows the 5-min from_db cache (fires at startup / first Sleep run / each cache refresh), per the issue's acceptance.

## Testing
- 5 new tests (`TestFromDbSleepLLMPrecedence`) covering the DB×env×FORCE_ENV matrix; `tests/neural/test_config.py` 43 passed.
- Independent review: no Critical/Warning. Empirically verified all 5 new tests **fail on origin/main** and pass here; ollama→self_hosted coercion interaction checked both directions (no false-positive WARN); no import cycle (`utils.logger` is a leaf).

### DIFF
--- a/backend/src/neural/config.py
+++ b/backend/src/neural/config.py
@@ -633,7 +633,17 @@ class NeuralMemoryConfig:
                 continue
             env_is_set = os.getenv(env_var) is not None
             use_env = force_env and env_is_set
-            if env_is_set and str(db_value) != env_value:
+            # Compare CANONICAL values: ``env_value`` (from base_config) is
+            # already ollama→self_hosted coerced by __post_init__, so a legacy
+            # 'ollama' DB row must be canonicalized the same way or the
+            # comparison would WARN a mismatch even though the effective
+            # provider is identical after construction (Copilot, PR #1186).
+            # Assignment keeps the RAW value — __post_init__ still owns the
+            # legacy coercion + its dedicated deprecation warning.
+            db_cmp = str(db_value)
+            if cfg_key == "sleep_llm_provider" and db_cmp == "ollama":
+                db_cmp = "self_hosted"
+            if env_is_set and db_cmp != env_value:
                 logger.warning(
                     "sleep_llm_config_mismatch",
                     key=cfg_key,
--- a/backend/src/neural/config.py
+++ b/backend/src/neural/config.py
@@ -9,8 +9,12 @@ import time
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, ClassVar
 
+from utils.logger import get_logger
+
 if TYPE_CHECKING:
     from sqlalchemy.ext.asyncio import AsyncSession
+
+logger = get_logger(__name__)
 
 # Cache TTL in seconds (5 minutes)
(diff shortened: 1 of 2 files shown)


JFK:
Addressed the Copilot comment: the mismatch compare now canonicalizes a legacy `ollama` DB row to `self_hosted` before comparing (comparison only — assignment keeps the raw value so `__post_init__` still owns the legacy coercion + deprecation warning). Added regression test `test_no_warn_for_legacy_ollama_db_row_vs_coerced_env`; `tests/neural/test_config.py` 44 passed.

Issue #1183, opened 2026-07-04
fix(sleep): sleep_summary.ok=true masks total LLM judge failure


JFK:
## Problem

In the week1-derisk Day-5 eval campaign, the Sleep judge failed on **every** call (`llm_call_failures=5/5`, root cause: stale `neural_config` DB row — see #1182) yet the run still reported `sleep_summary.ok=true`. A fully non-functional judge is indistinguishable from a healthy run at the summary level.

`llm_call_failures_total` was since promoted to a top-level results field (#1177 era), which makes the count *visible* — but the `ok` **semantics are unchanged**, so anything gating/alerting on `ok` still sees green.

## Proposed fix

- If an LLM judge is configured and **all** judge calls in the run failed → `ok=false` (hard failure), or
- introduce a three-state `status: ok | degraded | failed`, with `degraded` for partial failures (some calls failed) and `failed` for 100% failure.
- Whichever shape: monitoring/alerts must be able to key off it without parsing counters.

## Acceptance

- A run with judge configured and 0 successful judge calls can no longer report plain `ok=true`.
- Partial-failure state is distinguishable from full success.

Evidence: kagura-memory-eval `experiments/week1-derisk.md` Day-5 mechanism-attribution note. Related: #1182, #1177.

PR #1188, opened 2026-07-04
fix(api): split oversize dedup clusters instead of wholesale deferral (#1184)

### LINKED ISSUE #1184: fix(sleep/dedup): union-find mega-cluster > MAX_CLUSTER_SIZE silently defers ALL dedup candidates unjudged
## Problem

On the Day-5 `update_slice` eval corpus, dedup was **structurally inert**: the templated register pushes cross-unit similarity over the 0.92 candidate threshold, so **782 candidate pairs union-find into a single cluster**, the cluster exceeds `MAX_CLUSTER_SIZE=5`, and the whole set is **deferred unjudged BEFORE any LLM call** — silently. `dedup: 0 merges` looks identical to "nothing to merge".

This is not an eval-only artifact: real workloads with templated notes (daily standups, status logs, form-like records) will produce the same high mutual similarity and hit the same mega-cluster deferral, permanently disabling dedup for exactly the corpora that need it most.

## Proposed fix

Two independent parts:

**1. Observability (small, do first)**
- Report deferred work in `sleep_summary`: e.g. `dedup_deferred_pairs`, `dedup_oversize_clusters` (count + max size). "0 merges because nothing matched" and "0 merges because everything was deferred" must be distinguishable.

**2. Oversize-cluster splitting strategy (design needed)**
- Options: (a) threshold escalation within an oversize cluster (re-cluster at a stricter similarity until components fit `MAX_CLUSTER_SIZE`), (b) re-split by descending pair similarity and judge only the top-k subclusters per run (budgeted, makes progress across runs), (c) sample representative pairs from the mega-cluster instead of skipping it entirely.
- Whatever the choice, an oversize cluster should degrade to *partial* dedup progress, not zero.

## Acceptance

- `sleep_summary` exposes deferred-unjudged dedup volume.
- A corpus whose candidate graph collapses into one oversize cluster still gets some judged merges per Sleep run.

Evidence: kagura-memory-eval `experiments/week1-derisk.md` Day-5 mechanism-attribution note (782 pairs → 1 cluster > MAX_CLUSTER_SIZE=5, judge-independent deferral). Related: #1182, #1183.

JFK:
Closes #1184

## Summary
On templated corpora (daily standups, status logs) every note is pairwise-similar above the 0.92 candidate threshold, union-find collapses ALL candidates into one mega-cluster (Day-5 eval: **782 pairs → 1 cluster** > `MAX_CLUSTER_SIZE=5`), and the whole set was deferred unjudged **before any LLM call** — dedup went structurally inert, with `0 merges` indistinguishable from "nothing to merge".

## Changes (`backend/src/services/sleep/dedup_merge.py`)
**Part 1 — Observability**: `dedup_result.details` now carries `oversize_clusters`, `oversize_max_size`, `split_subclusters`, and `deferred_pairs` (cap-blocked union attempts — a documented proxy for unjudged candidate volume). `deferred_clusters` is retained as the legacy key (original oversize count; consumed by the frontend narrative reader).

**Part 2 — Splitting**: new `_split_oversize_cluster` — capacity-capped greedy agglomeration. Walk the cluster's internal pairs in **descending similarity** (deterministic str-uuid tie-break) and union components only while the merged size stays ≤ `MAX_CLUSTER_SIZE`. The tightest duplicates coalesce first, every subcluster fits one LLM batch, skipped cross-component pairs are re-candidates next run (post-merge the landscape shrinks). Subclusters append after normal clusters so the pre-existing set keeps budget priority; `sleep_max_llm_calls_per_run` still bounds total cost.

Acceptance: a corpus whose candidate graph collapses into one oversize cluster now gets judged merges every run instead of zero, and the deferred volume is visible in `sleep_summary`.

## Testing
- `TestSplitOversizeCluster` (deterministic): weakest-link split point, cap invariants on a complete 23-node graph, external-pair filtering, degenerate no-pairs guard.
- Execute-level: 6-chain (split + partial merges + `deferred_pairs==1`), complete-12 Day-5 shape (deterministic 3 subclusters, merges > 0).
- Sleep suite in container harness: 233 passed; ruff clean.
- Independent code review: no Critical; 1 Warning (proxy-semantics overclaim) + 2 actionable Info — all addressed in the last commit. Verifier ran randomized simulations (2k–20k trials) confirming the split algorithm's invariants and the test math.

## Notes
- Expect a trivial additive conflict with #1187 (#1183) in the same `details` dict — whichever merges second rebases (I'll handle it).
- Follow-up (from review): surface `oversize_clusters`/`split_subclusters`/`deferred_pairs` in the dedup UI narrative — today the narrative i18n string doesn't render deferral info at all.

### DIFF
--- a/backend/src/services/sleep/dedup_merge.py
+++ b/backend/src/services/sleep/dedup_merge.py
@@ -129,8 +129,11 @@ def _split_oversize_cluster(
         max_size: per-subcluster member cap (defaults to MAX_CLUSTER_SIZE).
 
     Returns:
-        (subclusters with >= 2 members, count of internal pairs skipped by
-        the size cap — i.e. candidate pairs left unjudged this run).
+        (subclusters with >= 2 members, count of union attempts blocked by
+        the size cap — an order-of-magnitude PROXY for candidate pairs left
+        unjudged this run, not an exact count of final cross-subcluster
+        edges: an early cap-blocked pair may still end up same-component via
+        other edges, and same-component skips are never counted).
     """
     internal = sorted(
         (p for p in pairs if p[0] in cluster and p[1] in cluster),
@@ -260,6 +263,9 @@ class DedupMergePhase:
             subclusters, skipped = _split_oversize_cluster(big, pairs)
             split_subclusters += len(subclusters)
             deferred_pairs += skipped
+            # Appended AFTER the normal-size clusters deliberately: under
+            # budget exhaustion the pre-existing cluster set keeps priority
+            # and mega-cluster subclusters only consume the margin.
             processable.extend(subclusters)
 
         if oversize:
@@ -336,8 +342,9 @@ class DedupMergePhase:
             "oversize_clusters": len(oversize),
             "oversize_max_size": max((len(c) for c in oversize), default=0),
             "split_subclusters": split_subclusters,
-            # Candidate pairs left unjudged this run by the size cap —
-            # distinguishes "0 merges: nothing matched" from "0 merges:
+            # Cap-blocked union attempts — a PROXY for candidate volume left
+            # unjudged this run (see _split_oversize_cluster docstring).
+            # Distinguishes "0 merges: nothing matched" from "0 merges:
             # work was deferred".
             "deferred_pairs": deferred_pairs,
             "merged": merged_count,
--- a/backend/src/services/sleep/dedup_merge.py
+++ b/backend/src/services/sleep/dedup_merge.py
@@ -120,7 +120,11 @@ def _split_oversize_cluster(
     stays within ``max_size``. The tightest duplicates coalesce first, every
     component stays judgeable in one LLM batch, and skipped cross-component
     pairs are re-candidates next run (post-merge, the landscape shrinks).
-    Deterministic: ties break on the stringified pair ids.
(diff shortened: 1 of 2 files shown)


JFK:
Both Copilot comments addressed in de465dbe:

1. **Orientation-dependent tie-break — valid catch, fixed.** `_split_oversize_cluster` now canonicalizes pair endpoints (str-ascending) before sorting, so the split is independent of the `(memory.id, hit_id)` orientation `_find_similar_pairs` happens to emit. Added regression test `test_split_is_orientation_independent` (same logical pair set, flipped orientations → identical subclusters + skipped count).

2. **`split_subclusters == 3` assertion — kept, now provably deterministic.** With canonical orientation the equal-score tie-break walk is lexicographic on `(str(a), str(b))`: the str-smallest node's edges sort first, so its component fills to capacity (5), then the next-smallest unmerged node fills the second (5), leaving the final pair (2) → always (5,5,2) = 3 subclusters for a complete 12-node equal-score graph, regardless of concrete UUID values. (An independent reviewer's 20k randomized-UUID simulation of the algorithm also never produced anything but 3.) The test comment now states this reasoning.

`tests/services/sleep/test_dedup_merge*`: 59 passed.

Issue #439, opened 2026-04-25
feat(api+ui): UUID-addressed memory update endpoint + wire EditMemoryDialog


JFK:
Follow-up from #433. Edit was deferred because no UUID-addressed update endpoint exists.

## Problem

`MemoriesTabPanel` (introduced in #433) deliberately omits the Edit affordance — `MemoryDetailDialog.onEdit` is rendered conditionally and the panel passes nothing. Reason:

- Backend `backend/src/api/routes/memory.py` exposes only POST `/remember`, `/reference`, `/forget` (UUID-addressed) — no PUT/PATCH.
- The legacy `updateMemory()` in `frontend/src/lib/api/memory.ts` calls `PUT /memory/{key}` which **does not exist** on the router (404 in production).
- New `/memory/list` rows lack `key/scope/agent_name` so even if the legacy PUT were live, the dialog couldn't construct the call.

Without a UUID update path the Edit button is dead by design.

### Backend
1. Add `PATCH /api/v1/memory/{memory_id}` (or `PUT`) accepting partial updates: `summary`, `content`, `type`, `importance`, `tags`, `details`. Not `scope` (orthogonal lifecycle), not `context_id` (move-to-context is a separate operation).
2. Permission: reuse forget's pattern — `can_access_memory` check via `PermissionService` (see `memory_service.py:1104`). No new `can_update_memory` method needed; the access envelope (workspace member for shared, creator for private) is identical.
3. Updates `memory.updated_at` automatically (existing `onupdate=func.now()`).
4. Sync the embedding pipeline if `summary`/`content` changes — re-invoke `add_memory_to_qdrant` for upsert (synchronous, same as remember path; qdrant upserts by point id, so re-calling with the same memory id overwrites the vector).
5. **Invalidate neural edges when `summary`/`content` changes** — call `edge_repo.delete_node_edges(...)` (forget's pattern). Old Hebbian/activation weights were computed against the previous embedding and become stale once the vector changes; the next sleep run rebuilds them.
6. **Reject PATCH on soft-deleted memories** — return 410 Gone when `memory.deleted_at IS NOT NULL`. forget uses soft delete, so update must respect that lifecycle.
7. Returns `MemoryReference` (full detail), reusing the existing serializer + `Z`-suffix datetime fix from #433.
[...]

PR #1324, opened 2026-07-16
fix(api): enforce soft-delete tombstone in reference/explore/forget

### LINKED ISSUE #1316: fix(mcp): reference() and explore() return soft-deleted memories
## Summary

`reference()` returns the **full content of a soft-deleted memory** with `status=success`, and `explore()` returns the deleted memory as `seed_memory` (summary exposed). All other read surfaces correctly exclude soft-deleted rows, so the `is_deleted` filter is inconsistent across read paths.

Verified against v0.51.1 via a live MCP verification run (2026-07-16), in a disposable test context.

## Repro

1. `remember(...)` → memory_id X
2. `forget(memory_id=X)` → `deleted_count=1`; X disappears from `recall()` and from `get_context_info` stats ✓
3. `reference(memory_id=X)` → **`status=success` with full `content`** ✗ (expected `memory_not_found`)
4. `explore(memory_id=X)` → `status=success`, `seed_memory` carries the deleted memory's summary, `related_memories=[]` with `reason="seed_not_in_graph"` ✗

Cross-check of the same deleted id (correct behavior to mirror):
- `update_memory` → `Memory not found` ✓
- `feedback` → `not_found` ✓
- `recall` (semantic + type filter) → absent ✓

## Why it matters

- Soft delete has a 30-day retention window; during that window "deleted" data remains readable by direct-ID fetch. Users reasonably assume `forget` removes visibility everywhere.
- Erasure semantics: any erasure/retention audit that samples via reference() will contradict recall behavior (relevant to the erasure runbook's guarantees).

## Expected

`reference()` and `explore()` (seed lookup) apply the same soft-delete filter as `recall`/`update_memory`, returning `memory_not_found`.

## Acceptance criteria

- reference() on a forgotten memory returns `memory_not_found`
- explore() on a forgotten seed returns the not-found error (not an empty exploration)
- Regression tests covering both paths (delete → each read surface)

## Root cause (investigated)

`MemoryRepository.get()` (`backend/src/repositories/memory.py:31-41`) returns soft-deleted rows (no `deleted_at` filter); callers are individually responsible for the tombstone guard and these two omit it:
- `reference()`: `backend/src/services/memory_service.py:1095-1098` — no `deleted_at is not None` check
- `explore()` seed lookup: `backend/src/services/memory_service.py:3475-3477` — same; forget() already removed the edges, hence `seed_not_in_graph`

Pattern to mirror: `update_memory`'s guard at `memory_service.py:559-560` (`if memory.deleted_at is not None: raise NotFoundException(...)`). recall filters at query level (`:2570`); explore's neighbor query already filters (`:3616`) — only the seed leaks.

### LINKED ISSUE #1320: fix(mcp): forget() double-delete reports deleted_count=1 again (idempotency accounting)
## Summary

`forget(memory_id=X)` called **twice** returns `deleted_count=1` (with `memory_ids=[X]`) **both times**. The second call should report `deleted_count=0` — the row is already soft-deleted. The delete path apparently doesn't exclude already-deleted rows before counting/re-marking.

Contrast with `delete_file`, which correctly returns `not_found` on a second delete — the two delete surfaces disagree on idempotency reporting.

Verified against v0.51.1 (2026-07-16). Note the memory IS actually deleted on the first call (recall/stats exclude it) — this is an accounting/response bug, not a persistence bug.

## Why it matters

- Clients that sum `deleted_count` for audit/cleanup accounting get inflated numbers on retries.
- Retry-safe clients can't distinguish "deleted now" from "was already deleted".

## Expected

Second `forget` of the same id returns `deleted_count=0, memory_ids=[]` (soft-delete UPDATE filtered by `is_deleted = false`, count from rows actually transitioned).

## Acceptance criteria

- Double-forget regression test asserting 1 then 0
- Query-mode forget unaffected (still deletes top-k live matches only)

## Root cause (investigated)

`forget()` Case 1 (`backend/src/services/memory_service.py:3185-3260`): after `memory_repo.get()` (which returns soft-deleted rows), there is no tombstone check at `:3188` — the second call re-stamps `deleted_at`/`deleted_by` (`:3228-3230`) and re-appends the id (`:3256`). Fix: treat `memory.deleted_at is not None` as no-op (same guard pattern as `memory_service.py:559-560`).

JFK:
## Summary

Closes #1316, Closes #1320.

`MemoryRepository.get()` returns soft-deleted rows; three callers omitted the `deleted_at` tombstone guard (found by a live MCP verification run against v0.51.1):

- **reference()** returned the full content of a forgotten memory by direct-id fetch during the 30-day retention window (erasure-semantics violation)
- **explore()** surfaced a forgotten memory as the exploration seed (summary leaked with `reason="seed_not_in_graph"` — forget had already removed its edges)
- **forget()** re-stamped `deleted_at`/`deleted_by` on an already-deleted row and reported `deleted_count=1` again on repeat calls (audit-accounting bug)

All three now mirror `update_memory`'s existing guard (`memory_service.py` — `if memory.deleted_at is not None: raise NotFoundException(...)` / no-op for forget). recall/stats/feedback/list_edges were already correct; this closes the remaining inconsistent read/delete surfaces.

## Tests

- `test_reference_soft_deleted_raises_not_found` (#1316)
- `test_explore_soft_deleted_seed_raises_not_found` (#1316)
- `test_forget_by_id_already_deleted_counts_zero` — also asserts no re-stamp (#1320)
- Existing mocks gained explicit `deleted_at=None` (MagicMock auto-attrs / SimpleNamespace)

`pytest tests/services/test_memory_service.py tests/services/test_permission_agent_binding.py` — 122 passed. ruff clean.

### DIFF
--- a/backend/src/mcp_server/tools/explore.py
+++ b/backend/src/mcp_server/tools/explore.py
@@ -20,6 +20,7 @@ from mcp_server.tools._helpers import (
     _validate_memory_id,
     execute_with_timeout,
 )
+from utils.exceptions import NotFoundException
 
 logger = logging.getLogger(__name__)
 
@@ -94,6 +95,26 @@ async def handle_explore(
         except _ContextNotFoundError as e:
             await db.rollback()
             return e.to_response()
+        except NotFoundException:
+            # #1316: a missing, tombstoned, or inaccessible seed is a normal
+            # client outcome — return the structured envelope (mirroring
+            # handle_reference) instead of falling through to the generic
+            # dispatch handler as a 500-logged raw error.
+            await db.rollback()
+            await _log_tool_usage(
+                db,
+                user_id,
+                "explore",
+                start_time,
+                404,
+                args.get("context_id"),
+                workspace_id,
+            )
+            return _error_response(
+                "memory_not_found",
+                f"Memory not found or you don't have access: {memory_uuid}",
+                help="Use recall() to find memories you have access to.",
+            )
         except Exception:
             await db.rollback()
             await _log_tool_usage(
--- a/backend/src/mcp_server/tools/memory.py
+++ b/backend/src/mcp_server/tools/memory.py
@@ -768,6 +768,12 @@ async def handle_reference(
                     operation_name="reference",
                 )
             except NotFoundException:
+                # #1316 review: record the 404 in usage telemetry, keeping
+                # reference and explore symmetric for the identical outcome.
+                await db.rollback()
+                await _log_tool_usage(
+                    db, user_id, "reference", start_time, 404, current_context_id, workspace_id
+                )
                 return _error_response(
                     "memory_not_found",
                     f"Memory not found or you don't have access: {request.memory_id}",
--- a/backend/src/repositories/memory.py
+++ b/backend/src/repositories/memory.py
@@ -158,7 +158,12 @@ class MemoryRepository(BaseRepository[Memory]):
         Returns:
             True if deleted
         """
-        memory = await self.get(id)
+        # include_deleted=True: hard-delete is how the nightly retention purge
+        # (tasks/neural_tasks.cleanup_deleted_memories_task) erases tombstoned
+        # rows after the retention window — a default-filtered fetch would
+        # make the purge a silent no-op and retention/erasure would never
+        # physically happen (#1316 review sweep).
+        memory = await self.get(id, include_deleted=True)
         if not memory:
             return False
 
--- a/backend/src/repositories/memory.py
+++ b/backend/src/repositories/memory.py
@@ -28,16 +28,27 @@ class MemoryRepository(BaseRepository[Memory]):
         """
         self.db = db
 
-    async def get(self, id: UUID) -> Memory | None:
+    async def get(self, id: UUID, *, include_deleted: bool = False) -> Memory | None:
         """Get memory by ID.
+
+        #1316/#1320: soft-deleted (tombstoned) rows are EXCLUDED by default —
+        every read/mutation path that fetches by id must treat a forgotten
+        memory as absent during the retention window. The only caller that
+        legitimately needs tombstones is ``patch_memory`` (its #439 contract
+        returns 410 ``MemoryGoneError`` to authorized callers); it opts in
+        with ``include_deleted=True``.
 
         Args:
             id: Memory UUID
+            include_deleted: When True, return the row even if soft-deleted.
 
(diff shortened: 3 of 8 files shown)


JFK:
## xhigh code-review (10 finder angles + verify + sweep) — applied in 740a85d

**Confirmed findings (fixed):**
1. **Altitude**: the per-call-site guards were a bandaid — `MemoryRepository.get()` now default-filters tombstones (`include_deleted` opt-in for `patch_memory`'s #439 410 contract). Also hardens forget's query-mode loop (recall→get TOCTOU) which the inline approach missed.
2. **Existence oracle**: explore's tombstone raise used a different message shape than never-existed/denied ("Memory not found: <id>" vs "Memory <id> not found not found") — a caller could distinguish "once existed and was forgotten". All three raises unified to the two-arg form; test asserts the exact message.
3. **MCP surface**: `handle_explore` lacked an `except NotFoundException` arm — a forgotten seed became a 500-logged raw dispatch error (JSON-RPC -32603) instead of the graceful `memory_not_found` envelope `handle_reference` returns. Arm added with 404 usage telemetry.
4. **Self-caught during refactor**: `repo.update()`'s internal fetch had to opt in (`include_deleted=True`) — forget stamps `deleted_at` on the in-session row before calling `update()`, and the SELECT's autoflush would filter out its own row and break soft-delete entirely.

**Deliberate, documented (no change):** reads return uniform 404 pre-permission (privacy-preserving); 410 `MemoryGoneError` stays a patch-path affordance for authorized callers. Any future 410 adoption on reads must move the check after the permission gate (comment pinned in code). Tombstoned forget emits the same success/count=0 audit event as never-existed — consistent by design.

2345 backend tests green after refactor.

JFK:
**Sweep (Phase 3) findings — applied in the follow-up commit:**

1. **CRITICAL (caught by sweep):** `repo.delete()`'s internal fetch was still default-filtered after the refactor — the nightly retention purge (`cleanup_deleted_memories_task`) hard-deletes tombstoned rows via `repo.delete()`, so it would have silently no-oped on every row (telemetry still reporting `purged=N`). Tombstones would never be physically erased. Now opts in with `include_deleted=True` + regression test (soft-delete → delete → row gone).
2. `handle_reference` 404s now log usage telemetry, symmetric with the arm `handle_explore` gained.
3. `_update_in_place` pinned against a future `include_deleted=True` opt-in (would re-embed deleted content).

441 affected tests green.

Issue #1228, opened 2026-07-12
fix(api): attribute cross-context recall usage to every listed context (#1225 follow-up)


JFK:
## Summary

A cross-context `recall(context_ids=[A, B, ...])` logs usage under the FIRST listed context only (`current_context_id = cross_context_ids[0]` in the MCP recall handler; `_log_tool_usage` writes a single `UsageStats` row with that `context_id`). Read activity on the other listed contexts is therefore invisible to per-context consumers.

Concrete impact (documented as a known limitation in docs/ops/memory-health-report.md since #1225): the memory-health retrieval section grades a context `write_only_store` (WARN) when it has active memories but zero attributed recalls — a context read exclusively via cross-context recall false-positives, the exact false-WARN class the report's grading philosophy forbids.

## Proposal

Fix at the usage-logging layer: record read-attribution for EVERY listed context (e.g. one attribution row per listed context, or an attributed-context list alongside the primary row). Keep billing/quota semantics unchanged — one recall call remains one quota unit; attribution is diagnostic, not billable.

## Acceptance Criteria

- [ ] After `recall(context_ids=[A, B])`, per-context usage queries see read activity for BOTH A and B
- [ ] Quota/usage counting is unchanged (one call = one unit; no double-billing)
- [ ] Memory-health retrieval section no longer WARNs `write_only_store` for a context read only via cross-context recall (test pins this)
- [ ] The known-limitation caveat in docs/ops/memory-health-report.md is removed/updated
- [ ] Existing usage-stats consumers (quota service, workspace usage views) verified unaffected

PR #1325, opened 2026-07-16
fix(api): usage-log FK fallback + calibration vector-type crash

### LINKED ISSUE #1318: fix(api): usage_stats INSERT violates FK when MCP tool is called with a nonexistent context_id
## Summary

Calling any MCP tool with a **nonexistent context_id** (e.g. the all-zeros UUID) returns a clean `context_not_found` to the client, but server-side usage logging then fails with a ForeignKeyViolation:

```
event="usage_logging_failed"
IntegrityError … usage_stats_context_id_fkey
INSERT INTO usage_stats (…, context_id, …) VALUES (…, '00000000-0000-0000-0000-000000000000', …)
endpoint='mcp:get_context_info', status_code=404
```

The usage logger stamps the *requested* context_id without validating it exists, so the row insert violates the FK. Result: the call is never counted in usage_stats and each such request emits an error-level log line (noise + monitoring false positives + a cheap way for a client to spam error logs).

Verified against v0.51.1 (2026-07-16) — reproduced with `recall`/`get_context_info` on a fake UUID.

## Expected

Usage logging never fails on user input. Either:
- validate/null the context_id before insert (`context_id=NULL` when the context is unknown), or
- make the column nullable-on-unknown by catching the FK violation and retrying with NULL (first option preferred — cheaper and no error path).

## Acceptance criteria

- MCP call with fake context_id → 404-equivalent to client AND a usage_stats row recorded (context_id NULL)
- No `usage_logging_failed` in logs for this scenario
- Regression test at the usage-logging layer

## Root cause (investigated)

`get_context_info`'s `_ContextNotFoundError` handler (`backend/src/mcp_server/tools/context.py:183-191`) passes the RAW request value `args.get("context_id")` to `_log_tool_usage` (line 189) — the id was never validated to exist. `_helpers.py:613-622` forwards it verbatim, `usage_logger.py:64-77` INSERTs it, FK violation → caught at `usage_logger.py:90` and the WHOLE usage row is dropped (`return False`). Same raw-passing pattern on the 500 path (`context.py:201`) and `delete_context`'s 500 handler (`context.py:765`).

Fix (combine both):
- Call sites: pass `None` on not-found/error paths where the context was never resolved.
- Central hardening: in `log_usage`, catch `IntegrityError`, rollback, retry INSERT with `context_id=None` (column is nullable, FK is `ondelete="SET NULL"` per `models/auth.py:1235-1236`) so the usage/quota event is preserved for any future caller mistake.

### LINKED ISSUE #1319: fix(api): neural calibration task crashes with float(dict) — no calibration row ever persists
## Summary

During `remember()` calls, a background calibration task crashes:

```
event="calibration_task_failed"
model="text-embedding-3-small", dimensions=512, context_id=null
error="float() argument must be a string or a real number, not 'dict'"
exc_info=true
```

Observed 4× within a short verification window on v0.51.1 (2026-07-16), correlated with `remember()` writes (including plain-ASCII payloads — not a Unicode issue). The task passes a dict to `float()` — likely an embedding-API response object (or config/stats dict) where a raw scalar is expected. Whatever this task calibrates (similarity-score calibration?) is silently not happening.

## Expected

- Calibration task handles the actual payload shape (extract the scalar it needs) and completes.
- If the failure is benign-by-design, it should be handled without `exc_info` error logs on every write.

## Acceptance criteria

- No `calibration_task_failed` on normal remember() traffic
- Unit test feeding the real payload shape through the calibration path

## Root cause (investigated)

Type mismatch between the two consumers of fetched vectors in the neural calibration task:
- `fetch_vectors()` returns `dict[str, list[float]]` (`backend/scripts/measure_embedding_threshold.py:161-189`)
- `measure_top_k(sampled, vectors, ...)` consumes the dict correctly (`backend/src/tasks/neural_calibration.py:448`)
- **`neural_calibration.py:500`** passes the same dict to `measure_random_pair(vectors, SAMPLE_RANDOM_PAIRS)` whose signature expects `list[list[float]]`; `np.asarray(<dict>, dtype=np.float64)` at `measure_embedding_threshold.py:257` → `float(dict)` TypeError.

Fix: `measure_random_pair(list(vectors.values()), SAMPLE_RANDOM_PAIRS)` at `neural_calibration.py:500`.

Impact (why this matters more than log noise): the exception fires BEFORE `db.commit()` (`neural_calibration.py:532`), so the already-staged `knn_seed` row is rolled back too — **neither calibration kind persists**. Runtime effect: `resolve_knn_threshold` finds no row → kNN-seed auto-association edges stay disabled at remember() time, and the semantic edge gate falls back to the uncalibrated config default; every qualifying remember() re-triggers bootstrap (throttled) and fails again. Regression test should run the calibration path end-to-end with a small vector fixture.

JFK:
## Summary

Closes #1318, Closes #1319. Two server-side error-log findings from the v0.51.1 live MCP verification run, both reproducible in production logs.

### usage_stats FK violation (#1318)
Any MCP call with a nonexistent context_id returned a clean 404 to the client, but the usage INSERT stamped the raw, unvalidated request id → `usage_stats_context_id_fkey` violation → `usage_logging_failed` error log **and the whole usage/quota row was dropped**.

- `get_context_info`'s 404 handler no longer stamps the known-unresolved id
- `log_usage` now catches `IntegrityError`, rolls back, and retries with `context_id=NULL` (column is nullable / `ondelete="SET NULL"`), preserving the event for any caller that stamps an invalid id (logged as `usage_logging_context_fk_fallback` warning)

### calibration crash (#1319)
`fetch_vectors()` returns `dict[str, list[float]]`; `measure_random_pair()` expects the bare vector list. Passing the dict made `np.asarray(<dict>)` call `float(dict)` → every calibration run failed **pre-commit, so neither calibration kind (knn_seed / edge_gate) ever persisted** — kNN-seed auto-association stayed disabled and the semantic edge gate ran on its uncalibrated fallback, with `calibration_task_failed` on every qualifying remember(). Fixed by passing `list(vectors.values())`.

## Tests

- New `tests/utils/test_usage_logger.py`: success / FK-fallback (asserts the retry INSERT carries `context_id=None`) / no-context FK / retry-failure / generic-failure
- New `test_compute_calibration_passes_vector_list_to_measure_random_pair`: drives `compute_calibration` with a faked measurement module; pins the list-shaped edge_gate sampling, both upserts staged, single commit

`pytest tests/mcp_server tests/tasks tests/utils` — 811 passed. ruff clean.

### DIFF
--- a/backend/src/mcp_server/tools/context.py
+++ b/backend/src/mcp_server/tools/context.py
@@ -42,6 +42,7 @@ async def handle_get_context_info(
     from db.base import get_db
 
     start_time = time.time()
+    current_context_id = None
     async for db in get_db():
         try:
             from services.memory_service import MemoryService
@@ -201,7 +202,8 @@ async def handle_get_context_info(
                 "get_context_info",
                 start_time,
                 500,
-                args.get("context_id"),
+                # #1318: resolved-or-None — never the raw request value.
+                current_context_id,
                 workspace_id,
             )
             logger.error(f"get_context_info_failed: {e}", exc_info=True)
--- a/backend/src/mcp_server/tools/context.py
+++ b/backend/src/mcp_server/tools/context.py
@@ -186,7 +186,10 @@ async def handle_get_context_info(
                 "get_context_info",
                 start_time,
                 404,
-                args.get("context_id"),
+                # #1318: the requested context_id failed resolution — it must
+                # not be stamped as the usage_stats FK (nonexistent UUIDs
+                # violate the constraint and drop the whole usage row).
+                None,
                 workspace_id,
             )
             return e.to_response()
--- a/backend/src/mcp_server/tools/edge.py
+++ b/backend/src/mcp_server/tools/edge.py
@@ -166,6 +166,7 @@ async def handle_list_edges(
     limit = args.get("limit")
 
     start_time = time.time()
+    current_context_id = None
     async for db in get_db():
         try:
             current_context_id = _resolve_context_id(args["context_id"])
@@ -221,14 +222,14 @@ async def handle_list_edges(
             )
         except _ContextNotFoundError as e:
             await db.rollback()
-            await _log_tool_usage(
-                db, user_id, "list_edges", start_time, 404, args.get("context_id"), workspace_id
-            )
+            # #1318: the requested context_id failed resolution — do not
+            # stamp it as the usage_stats FK.
+            await _log_tool_usage(db, user_id, "list_edges", start_time, 404, None, workspace_id)
             return e.to_response()
         except Exception:
             await db.rollback()
             await _log_tool_usage(
-                db, user_id, "list_edges", start_time, 500, args.get("context_id"), workspace_id
+                db, user_id, "list_edges", start_time, 500, current_context_id, workspace_id
             )
             raise
 
@@ -268,6 +269,7 @@ async def handle_create_edge(
     from repositories.neural_edge import NeuralEdgeRepository
 
     start_time = time.time()
+    current_context_id = None
     async for db in get_db():
         try:
             current_context_id = _resolve_context_id(args["context_id"])
@@ -308,7 +310,7 @@ async def handle_create_edge(
         except Exception:
             await db.rollback()
             await _log_tool_usage(
-                db, user_id, "create_edge", start_time, 500, args.get("context_id"), workspace_id
+                db, user_id, "create_edge", start_time, 500, current_context_id, workspace_id
             )
             raise
 
@@ -355,6 +357,7 @@ async def handle_update_edge(
     from repositories.neural_edge import NeuralEdgeRepository
 
     start_time = time.time()
+    current_context_id = None
     async for db in get_db():
         try:
             current_context_id = _resolve_context_id(args["context_id"])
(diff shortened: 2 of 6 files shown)


JFK:
## xhigh code-review (3 consolidated finder angles) — applied in 5df7032

**Confirmed (fixed):**
1. **True-after-fallback defeated the #1228 attribution guard**: `_log_tool_usage` keys `billable_row_written` on the return to gate secondary `ContextReadAttribution` writes. A degraded row (context nulled by the FK fallback) returning True would attribute secondaries while the primary read is invisible — `memory_health_service` could then false-flag the primary as `write_only_store`. Fallback now returns **False** (row still preserved for quota); docstring redefines the signal as 'written WITH requested attribution' + shared-session rollback caveat; retry failure logs the original IntegrityError too.
2. **`list_edges` 404 arm was the exact unfixed twin** of the get_context_info fix (frequent path: a client polling a stale context id = FK violation + fallback round-trip + warning per request) — now passes None. Edge handlers' and get_context_info's 500 arms pass the resolved-or-None local.

**Deliberate (documented):** `delete_context`/`merge_contexts` 500 arms keep their format-validated ids (context existence nearly guaranteed on those paths; central fallback covers the residual case; merge's `source_raw` carries a documented audit rationale). A `workspace_id` FK violation correctly still fails (retry re-passes it → False; never a false True). The 404→None change also stops **denied-but-existing** contexts from appearing in context usage timelines — behavior change judged an improvement (denied probes shouldn't inflate a context's api_calls).

Verified clean: no double-commit path in the retry; `now` reuse intentional; closure captures only scalars; calibration list fix matches the script's own CLI call site; `measure_top_k` keeps the dict.

Issue #1047, opened 2026-06-18
Surface recency/staleness markers and a "not in memory" confidence signal in recall results


JFK:
## Summary

The consumer of `recall` is an AI agent, not a human reading a UI. Two agent-facing signals are missing from recall output today:

1. **Recency / staleness marker** -- results carry no per-result recency cue, so an agent cannot self-assess whether a fact may be stale; it must trust blindly. Surfacing `last_updated` / age (and, once available, a superseded/decayed flag) lets the agent weight trust without extra calls.
2. **"Not in memory" confidence** -- when nothing relevant exists, recall returns weak low-score hits indistinguishable from a genuine miss. The agent must run multiple probing searches to *infer* absence (observed in practice: 3 searches to conclude a fact was not stored). A low-confidence / empty signal lets the agent stop early, avoid hallucination, or decide to go external.

Both improve the experience for the actual consumer -- the agent -- without client changes beyond reading the new fields.

## Context

**Priority**: medium
**Suggested labels**: enhancement
**Suggested milestone**: v0.34.0

Agent-facing recall UX. Independent of the adoption-capture work (#1046) and the reinforce half. Grounded in a session where an agent had to run three recall queries to conclude a fact was absent.

## Acceptance Criteria

- recall results expose a per-memory recency cue (e.g. `last_updated` age) in the response payload.
- recall returns a top-level confidence/quality signal distinguishing "relevant results found" from "likely nothing relevant", without changing default ranking. The signal MUST be **per-context normalized / relative** (e.g. gap between the top hit and the context's background score distribution), **NOT a global absolute score threshold** -- hybrid scores are not calibrated across contexts/embedding models (e.g. text-embedding-3-small vs qwen3-embedding), so a fixed cutoff mis-calibrates.
- Additive and backward compatible: existing fields unchanged; new fields additive.
- Documented in the `recall` tool description so agents know the signals exist.

**Out of scope**: the staleness/supersede *detection* itself (separate track); UI surfacing.

Issue #979, opened 2026-06-10
feat(api): auto-requeue transiently failed embeddings in the sweep (bounded retries)


JFK:
## Context

Found during the #967 code review (PR #978). The eval harness now retries transient embedding/Qdrant-upsert failures by resetting `embedding_status` `failed` → `pending` (bounded at 2 attempts). Production has **no equivalent automatic recovery**:

- The periodic sweep (`src/tasks/embedding_tasks.py`) only requeues `pending` (>10s) and stale `processing` (>60s) rows — never `failed` (`embedding_tasks.py:43-48`).
- The only recovery for `failed` is the manual admin endpoint `retry_failed_embeddings` (`src/api/routes/admin.py:1155-1209`).

## Problem

A transient client-side stall (observed on WSL2 during #967: the Qdrant upsert times out at the client without ever reaching the server, while the server handles every arriving request in ~3ms) lands the memory in `failed` permanently. The memory is unsearchable until an admin notices and manually triggers the retry endpoint. Users have no signal and no self-service path.

## Proposal

Add a bounded auto-requeue path for transiently failed embeddings:

- Extend the sweep to also claim `failed` rows, gated by a retry counter (e.g. new `embedding_retry_count` column, max 2-3 attempts) so poison rows (bad API key, oversized input) don't loop forever.
- Distinguish transient vs permanent failure classes if cheap (timeout/connection vs 4xx-style provider errors).
- Keep the admin endpoint as the manual override for exhausted rows.

## References

- #967 / PR #978 — harness-side retry that surfaced the asymmetry
- `src/services/memory_service.py` `process_pending_embedding` claim logic (pending / stale-processing only)

PR #1326, opened 2026-07-16
fix(db): stop read paths from re-stamping memory.updated_at

### LINKED ISSUE #1317: fix(api): read paths mutate memory.updated_at, destroying the staleness cue
## Summary

Read operations mutate `memory.updated_at`: every `recall()` bumps `updated_at` of **all returned hits** to the current time; `reference()` does the same for the fetched memory. The MCP docs define `updated_at` as "the last time that fact was changed … a staleness cue" — read traffic destroys that signal permanently (verified on a high-traffic context: all frequently-recalled memories show `updated_at` = time of the latest recall, regardless of when the fact was last edited).

Verified against v0.51.1 via a live MCP verification run (2026-07-16).

## Evidence

- Create memory at T0 → first recall at T1 shows `updated_at=T1`; every subsequent recall bumps it again (T2, T3, …) with no writes in between.
- `reference()` alone also bumps it.
- Suspected cause: a read-path write-back (access counting / Hebbian co-activation / reinforce tracking) issuing an ORM UPDATE that triggers the `onupdate` of `updated_at`.

## Related observation (may be by-design — please confirm)

`scope` auto-promotes `working` → `persistent` after ~2 read accesses (observed within a minute of creation). If access-based promotion is intentional, it should still not touch `updated_at`.

## Expected

- `updated_at` changes only on explicit content/metadata edits (update_memory, supersede, sleep-maintenance mutations).
- Access tracking, if any, uses a separate column (e.g. `accessed_at` / `access_count`) that is excluded from `onupdate` semantics.

## Acceptance criteria

- recall/reference/explore/get_agent_bootstrap leave `updated_at` unchanged
- Regression test: recall twice, assert `updated_at` stable
- Docs stay truthful ("staleness cue") without edits

## Root cause (investigated)

`Memory.updated_at` declares `onupdate=func.now()` (`backend/src/models/memory.py:220-222`). Read paths call `MemoryRepository.update_access_stats()` (`backend/src/repositories/memory.py:202-248`) which dirties the row (`access_count`/`last_used_at`/`accessed_by_clients`) and flushes → `onupdate` stamps `updated_at`. #1047 patched only the RESPONSE snapshot (recall `memory_service.py:2741/2761`, reference `:1126/1181`); the DB column still mutates. Call sites: recall `:2744`, reference `:1132`, explore `:3540/:3596/:3683-3685` (explore lacks even the snapshot).

Fix direction (either):
1. Drop `onupdate=func.now()` — every genuine edit path already sets `updated_at` explicitly (`memory_service.py:644/854`, `repositories/memory.py:127`, resource_indexer `:761`, sleep paths), so `onupdate`'s only live effect is the spurious read bump; or
2. Keep `onupdate` and pin `updated_at=Memory.updated_at` in access-stat updates via Core update — exact precedent at `backend/src/mcp_server/tools/_helpers.py:383` (Context last-used touch).

Scope promotion is intentional (`config/retention.py:42-85`: access_count>=3 OR age>=7d OR importance>=0.7 OR >=2 clients) — the "~2 reads" observation is the OR-conditions. Whether `promote_to_persistent()` (`repositories/memory.py:250-267`) should bump `updated_at` is a value judgment to settle in this issue (scope genuinely changes).

JFK:
## Summary

Closes #1317.

`Memory.updated_at` declared `onupdate=func.now()`, so the access-stats flush performed by **every** recall/reference/explore (`update_access_stats`: access_count / last_used_at / accessed_by_clients) re-stamped it. Verified live: all frequently-recalled memories carry `updated_at` = time of the latest recall, which permanently destroys the documented staleness cue ("last time the fact was changed"). #1047 patched only the *response* snapshot — the DB column still mutated on every read.

## Fix

Drop `onupdate` from the column (client-side default — no migration). Every genuine edit path already sets `updated_at` explicitly (`update_memory`/`patch_memory`, repo `update()`, resource re-index, sleep merge/importance/reindex), so the `onupdate`'s only live effect was the spurious read-path bump.

Semantics settled per the issue: scope promotion (`promote_to_persistent`) also no longer counts as an "edit" — `scope`/`promoted_at` carry that history; `updated_at` now means content/metadata edits only. The #1047 response-snapshot mitigations in recall/reference remain (harmless, and still guard any other in-session dirtying).

## Tests

- Behavioral: `update_access_stats` (adoption path) leaves `updated_at` unchanged while `access_count` increments
- Declaration pin: `Memory.__table__.columns["updated_at"].onupdate is None`
- Sweep: `tests/services + repositories + api + mcp_server` — **3394 passed** (no hidden onupdate dependencies)

### DIFF
--- a/backend/src/api/routes/memory.py
+++ b/backend/src/api/routes/memory.py
@@ -782,10 +782,11 @@ async def list_memories(
         # naive UTC datetimes (DateTime without timezone=True). Tag the
         # serialized form with "Z" so JS clients (which parse naive ISO as
         # local time) don't render JST-shifted relative timestamps.
-        # ``updated_at`` is nullable (only set onupdate), so fall back to
-        # ``created_at`` for fresh rows that haven't been touched since
-        # insert. The frontend renders both as the row "updated" timestamp,
-        # so created_at is the correct fallback rather than null/empty.
+        # ``updated_at`` is nullable (set explicitly by edit paths — #1317
+        # removed the column's onupdate), so fall back to ``created_at`` for
+        # fresh rows that haven't been touched since insert. The frontend
+        # renders both as the row "updated" timestamp, so created_at is the
+        # correct fallback rather than null/empty.
         memory_items = [
             MemoryListItem(
                 id=str(m.id),
--- a/backend/src/mcp_server/tools/_helpers.py
+++ b/backend/src/mcp_server/tools/_helpers.py
@@ -380,6 +380,11 @@ async def _touch_context_last_used(db: "AsyncSession", context: Any) -> bool:
         update(Context)
         .where(Context.id == context.id)
         .where(or_(Context.last_used_at.is_(None), Context.last_used_at <= now - throttle))
+        # Self-assignment pins updated_at so Context's onupdate does not fire
+        # for this pure access touch. NOTE the sibling model solved the same
+        # problem the other way: Memory.updated_at has NO onupdate at all
+        # (#1317) and edit paths stamp it explicitly — don't "harmonize"
+        # either model without reading #1317 first.
         .values(last_used_at=now, updated_at=Context.updated_at)
     )
     return True
--- a/backend/src/models/memory.py
+++ b/backend/src/models/memory.py
@@ -217,9 +217,12 @@ class Memory(Base):
     created_at: Mapped[datetime] = mapped_column(
         DateTime, nullable=False, server_default=func.now()
     )
-    updated_at: Mapped[datetime | None] = mapped_column(
-        DateTime, nullable=True, onupdate=func.now()
-    )
+    # #1317: NO ``onupdate`` — ``updated_at`` is the documented staleness cue
+    # ("last time the fact was changed"), so it must move only when an edit
+    # path sets it explicitly (update/patch/upsert, resource re-index, sleep
+    # merge/re-eval all do). With ``onupdate=func.now()`` every read-path
+    # access-stats flush (recall/reference/explore) silently re-stamped it.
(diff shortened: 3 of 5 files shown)


JFK:
## xhigh code-review (2 consolidated finder angles) — applied in ce322ed

**Confirmed regression (fixed):** the #979 embedding-failure UPDATE relied on the removed `onupdate` to stamp `updated_at` with the **failure time** — `embedding_retry_eligible_clause`'s own docstring documented that mechanism. Without it, retry backoff anchored to the earlier processing-claim time, making slow failures retry-eligible early. The failure UPDATE now stamps `updated_at` explicitly; docstring corrected.

**Stale comments (fixed):** reference/recall snapshot comments still described the onupdate-expiry MissingGreenlet trap this PR removed (snapshots retained as cheap guards); `api/routes/memory.py` '(only set onupdate)' corrected; `Context._touch` now cross-references #1317 so the two models' opposite mechanisms (Context: self-assignment pin, Memory: no onupdate) aren't blindly 'harmonized'.

**Deliberate (no change):** `promote_to_persistent` no longer bumps updated_at (scope change = lifecycle, not an edit; `scope`/`promoted_at` carry the history). Embedding SUCCESS transition doesn't bump it. Pre-existing nuance flagged for awareness, not changed here: the processing-claim stamps `updated_at≈created_at` on every new memory, so 'null if never edited' in the tool description is rarely literally null — that stamp is itself the #979 anchor and predates this PR.

Verified clean: no alembic migration needed (onupdate is client-side); no test relies on the bump; sleep undo/restore paths unaffected.

PR #1327, opened 2026-07-16
fix(mcp): decode stringified state values

### LINKED ISSUE #1322: fix(mcp): get_state returns stringified JSON — coerce typeless 'value' arg (round-trip broken)
## Summary

`get_state` returns the stored value **double-encoded as a JSON string** instead of round-tripping the original JSON value:

```
set_state(key="step", value={"phase": "running", "n": 1})
get_state(key="step") → value = "{\"phase\": \"running\", \"n\": 1}"   # string, not object
set_state(key="n", value=42) → get_state → value = "42"                  # string, not number
```

The same double-encoded strings appear in `get_agent_bootstrap`'s `state` component. The docs promise "Arbitrary JSON value" round-trip (set_state stores "object, array, string, number, or boolean").

Verified against v0.51.1 (2026-07-16).

## Impact

Every agent consuming the session-state lane (incl. bootstrap) must speculatively `json.loads` values — and can't distinguish the string `"42"` stored deliberately from the number `42` double-encoded.

## Expected

`get_state` (single-key and list-all forms) and the bootstrap state component return the value with its original JSON type.

## Acceptance criteria

- Round-trip tests for object / array / string / number / boolean
- get_agent_bootstrap state component returns decoded values
- Note in the fix whether existing stored values need a read-time compatibility shim (values written before the fix)

## Root cause (investigated)

The column is JSONB (`backend/src/models/agent_state.py:49`) and the store/read paths are pass-through (no dumps/loads). The double-encoding happens **client-side and is never coerced back**: the `set_state` schema declares `value` with NO `"type"` key (`backend/src/mcp_server/tools/_definitions.py:2091-2093`), and the arg-coercion pass (`backend/src/mcp_server/tools/_arg_coercion.py:126-135`) only coerces args whose declared type is array/object/boolean — typeless fields are skipped. MCP clients that JSON-stringify complex args (the exact case the coercion layer exists for) therefore get their stringified `value` stored verbatim into JSONB and returned verbatim. A client sending a genuine JSON object is unaffected, which makes this look intermittent.

`get_agent_bootstrap`'s state component flows through the same `AgentStateService.list_state` (`agent_state_service.py:98-120`) and inherits the encoded values.

Fix direction: extend the coercion layer with a best-effort `json.loads` for typeless/any-typed string values (or coerce `value` in the set_state handler at `backend/src/mcp_server/tools/state.py:46`). Consider a read-time compatibility shim for values already stored as strings.

Sanjays2402:
## Related Issue

Closes #1322

## Summary

Decode JSON-stringified values for schema fields that accept arbitrary JSON, so `set_state` preserves object, array, string, number, and boolean types.

## Changes

- Decode valid JSON strings for untyped schema fields while preserving native and invalid string values
- Add round-trip regression coverage for every supported JSON value kind

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests passing
- [x] Manual testing completed
- [ ] All tests passing locally (`make test-local`)

Focused result: 51 tests passed. The new regression failed against the original source and passed with this fix.

## Quality Checklist

- [x] Code follows DRY/SOLID principles
- [x] Type hints added (target: 85%+ coverage)
- [x] No lint errors (`make lint`)
- [x] No type errors (`make type-check`)
- [x] No code duplication
- [x] Functions < 50 lines (Single Responsibility)
- [x] Files < 500 lines
- [x] Documentation updated (if needed)

## Screenshots (if applicable)

N/A — backend-only change.

---

To request AI review, comment `/review` on this PR.

### DIFF
--- a/backend/src/mcp_server/tools/_arg_coercion.py
+++ b/backend/src/mcp_server/tools/_arg_coercion.py
@@ -1,7 +1,8 @@
 """JSON-string argument coercion for MCP tool calls.
 
 Issue #197 / #196: some MCP clients serialize complex arguments (arrays,
-objects, booleans) as JSON strings before sending them over the wire:
+objects, booleans, or arbitrary JSON values) as JSON strings before sending
+them over the wire:
 
     {"tags": "[\"a\", \"b\"]"}      instead of  {"tags": ["a", "b"]}
     {"is_private": "true"}           instead of  {"is_private": true}
@@ -82,6 +83,16 @@ def _coerce_to_boolean(value: Any) -> Any:
     return value
 
 
+def _coerce_to_any(value: Any) -> Any:
+    """Decode JSON strings for schema fields that accept any JSON value."""
+    if not isinstance(value, str):
+        return value
+    try:
+        return json.loads(value)
+    except (ValueError, TypeError):
+        return value
+
+
 _COERCERS = {
     "array": _coerce_to_array,
     "object": _coerce_to_object,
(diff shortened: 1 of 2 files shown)


JFK:
Thank you for this contribution — the diagnosis is spot-on and the implementation is clean, with good regression coverage. Unfortunately this collided with an in-house fix for the same issue that was reviewed and merged earlier today (#1328), which is why this branch now conflicts — that's bad timing on our side (the issue was filed and fixed the same day), and I'm sorry your effort raced against it.

Two design notes from #1328's review, for context on where the merged fix differs: (1) decoding was deliberately scoped to objects/arrays only — retyping `"42"`→`42` / `"true"`→`true` corrupts legitimately-string values and diverges from the REST path, which stores exactly what the client sent; (2) `json.loads("null")` → `None` needs a guard, since the state column is `JSONB NOT NULL` and a decoded null would fail at insert.

We'd be glad to see more contributions from you. To avoid another race like this one, feel free to drop a comment on an issue you'd like to take before starting — we'll assign it to you so in-house work steers clear.

Issue #196, opened 2026-04-06
fix(mcp): create_context rejects is_private boolean parameter as string


JFK:
## Summary

Calling `create_context` via MCP with `is_private: true` fails because the backend treats the parameter as string `'true'` and PostgreSQL rejects it with `TypeError: Not a boolean value: 'true'`.

## Reproduction

MCP tool call (Claude Code, via `mcp__kagura-memory__create_context`):

```json
{
  "name": "kagura-dev",
  "display_name": "Kagura Memory Cloud Development",
  "is_private": true
}
```

Result:

```
(builtins.TypeError) Not a boolean value: 'true'
[SQL: INSERT INTO contexts (... is_private, ...) VALUES (..., $11::BOOLEAN, ...)]
[parameters: [{..., 'is_private': 'true', ...}]]
```

The parameter arrives at the SQL layer as the string `'true'` instead of the Python boolean `True`.

## Root cause (hypothesis)

MCP tool definition for `create_context` declares `is_private` as a boolean in the JSON schema, but the handler or argument binding is not coercing the incoming value from a string. Possibly:
- JSON schema type is `string` instead of `boolean`
- Handler unconditionally forwards the raw argument without type coercion
- Pydantic model at the REST side expects `bool` but receives `str` and doesn't auto-coerce (SQLAlchemy then rejects)

## Workaround

Omit `is_private` entirely — the DB default is `true` (private), which matches the most common intent.

## Impact

- `is_public` likely has the same bug (same parameter type)
- Blocks #18 context sync dogfood — any migration that carries `is_private`/`is_public` state hits this
- User-facing: confusing error message mentioning raw SQL internals

## Proposed fix

1. Verify the MCP tool JSON schema declares `is_private` and `is_public` as `{"type": "boolean"}`
2. Ensure the handler coerces string `"true"` / `"false"` to Python `True` / `False` if strict typing isn't enforced upstream
3. Add a unit test that calls `create_context` via the MCP handler with each combination

## Acceptance criteria

- [ ] `create_context(name=..., is_private=True)` succeeds via MCP
- [ ] `create_context(name=..., is_private=False, is_public=True)` succeeds via MCP
- [ ] `update_context` has the same guarantee
- [ ] MCP schema JSON explicitly types booleans

## Related

Discovered during v0.9.0 Soft Launch dogfood (#18) when migrating `kagura-dev` context from local → production via MCP.

PR #1328, opened 2026-07-16
fix(mcp): state value round-trip + error-envelope consistency

### LINKED ISSUE #1322: fix(mcp): get_state returns stringified JSON — coerce typeless 'value' arg (round-trip broken)

### LINKED ISSUE #1323: fix(mcp): normalize validation errors and not-found messages (pydantic leak, duplicated suffix, missing slug)
## Summary

Error-response consistency batch for MCP tools (three related paper cuts, one fix area):

### 1. Validation errors leak raw pydantic internals

Out-of-range inputs (remember importance, recall k, update_search_config weights, summary length) return:

```
{"status":"error","error":"1 validation error for RememberRequest\nimportance\n  Input should be less than or equal to 1 … https://errors.pydantic.dev/2.13/v/less_than_equal"}
```

while `list_tags(limit=501)` already does it right:

```
{"status":"error","error":"invalid_argument","message":"limit must be an integer between 1 and 500.","received":501}
```

Each pydantic leak is also logged server-side as a traceback (`mcp_tool_remember_failed` + `pydantic_core…ValidationError`), so ordinary client typos produce error-level log noise.

### 2. Duplicated "not found not found" in messages

- `explore` (bad memory_id): `"Memory <uuid> not found not found"`
- `delete_file` / `get_file_download_url` (deleted file): `"file <uuid> not found in workspace <uuid> not found"`

A formatter appends " not found" to an exception message that already ends with it.

### 3. `update_memory` error lacks the structured shape

Returns `{"status":"error","error":"Memory not found: <uuid>"}` — no error slug / message / help fields, unlike sibling tools (`reference` → `memory_not_found` + help).

Verified against v0.51.1 (2026-07-16).

## Expected

- One normalization layer converts pydantic ValidationError into the structured `{error: "invalid_argument", message: <human field summary>}` shape (no pydantic URLs), logged at warning without traceback.
- Not-found formatting deduplicated; update_memory returns `error: "memory_not_found"` + message + help like reference().

## Acceptance criteria

- Validation-error responses contain no "pydantic" tokens; field + constraint stated plainly
- Grep tests for the duplicated suffix
- update_memory not-found matches the documented error envelope

## Root cause (investigated)

1. Pydantic leak: the generic dispatch handler (`backend/src/mcp_server/tools/__init__.py:410-419`) does `except Exception as e: logger.error(..., exc_info=True); return {"status":"error","error": str(e)}` — remember (`memory.py:51-66`) and recall (`memory.py:423-434`) construct their `*Request` models with no local catch, so `str(ValidationError)` (with pydantic.dev URLs) goes to the client and a traceback goes to the logs. `update_search_config` (`search_config.py:91-94`) and `update_memory` (`memory.py:158-173`) catch locally but still stringify the raw dump into `message`. Clean pattern to mirror: `list_tags` (`context.py:919-933`). Suggested: a `ValidationError → {error:"invalid_argument", message:<loc+msg, no URLs>}` helper special-cased in the dispatch handler before the raw fallback, logged at warning without exc_info.

2. "not found not found": `NotFoundException` (`backend/src/utils/exceptions.py:152-159`) appends " not found" to the `resource` arg; explore passes a full sentence (`memory_service.py:3477/:3496` — `NotFoundException(f"Memory {id} not found")`), file storage passes `f"file {id} not found in workspace {ws}"` (`file_storage_service.py:171-175/:626-630/:138-139`). Switch to the two-arg form `NotFoundException("Memory", str(id))` used everywhere else.

3. update_memory envelope: `handle_update_memory` (`memory.py:145-252`) catches `_ContextNotFoundError` and `ValueError` but NOT `NotFoundException`, so it falls to the generic handler (raw string, no slug). Mirror `handle_reference`'s arm (`memory.py:770-775` → `memory_not_found` + help).

JFK:
## Summary

Closes #1322, Closes #1323. Two MCP response-quality fixes from the v0.51.1 live verification run.

### get_state round-trip (#1322)
`set_state`'s `value` field is typeless ("any") in the tool schema, so the #196/#197 arg-coercion pass skipped it — clients that JSON-stringify complex args got their stringified `value` stored verbatim into JSONB and returned verbatim (`"{\"phase\":…}"`, `"42"`), breaking the documented round-trip (also visible in `get_agent_bootstrap`'s state component). Typeless fields now get a best-effort JSON decode: strings that parse as JSON are decoded; plain strings and the literal `"null"` pass through unchanged (JSON null would violate NOT NULL storage). Values already stored double-encoded are ephemeral (TTL-bounded lane) and age out naturally.

### Error-envelope consistency (#1323)
- **Dispatch-level `ValidationError` arm**: out-of-range inputs (remember importance, recall k, …) returned raw pydantic dumps (`"1 validation error for RememberRequest\n… https://errors.pydantic.dev/…"`) and logged error-level tracebacks per client typo. Now: structured `invalid_argument` with a plain `field: constraint` message via the new `_format_validation_error`, logged at warning without traceback.
- `update_search_config` / `update_memory` construction errors use the same formatter (slug preserved).
- `handle_update_memory` gains the `memory_not_found` envelope arm (slug + help), mirroring `handle_reference` — previously fell through to the generic handler as a slug-less raw string.
- `file_storage_service`'s `NotFoundException` call sites switch to the two-arg form — the constructor appends " not found" itself, so full-sentence resource args produced doubled `"file <id> not found in workspace <ws> not found"` messages.

## Tests

Any-coercion matrix (object/number/boolean/plain-string/null/native), formatter unit tests, dispatch-arm envelope tests (remember + recall), update_memory envelope test. `tests/mcp_server + tests/services`: 2291 passed.

### DIFF
--- a/backend/src/mcp_server/tools/__init__.py
+++ b/backend/src/mcp_server/tools/__init__.py
@@ -418,9 +418,26 @@ async def execute_tool_call(
         # #1323: expected client-input errors — return the structured envelope
         # (no pydantic model names / errors.pydantic.dev URLs) and log at
         # warning without a traceback instead of an error-level exc_info dump.
-        message = _format_validation_error(e)
-        logger.warning(f"mcp_tool_{tool_name}_invalid_argument: {message}")
-        return _error_response("invalid_argument", message)
+        #
+        # Gate on the failing MODEL being a *Request* model: handlers also
+        # build pydantic Response models from DB rows deep in the service
+        # layer, and one of those failing is a SERVER data-integrity bug
+        # (e.g. a row that no longer fits the response schema — see the
+        # source_type Literal incident in models/schemas.py). Those must keep
+        # the loud error-level traceback + generic envelope below, not be
+        # misattributed to the caller as invalid_argument.
+        title = getattr(e, "title", "") or ""
+        if title.endswith("Request"):
+            message = _format_validation_error(e)
+            logger.warning(f"mcp_tool_{tool_name}_invalid_argument: {message}")
+            return _error_response("invalid_argument", message)
+        logger.error(f"mcp_tool_{tool_name}_failed: {e}", exc_info=True)
+        return [
+            TextContent(
+                type="text",
+                text=json.dumps({"status": "error", "error": str(e)}),
+            )
+        ]
     except Exception as e:
         logger.error(f"mcp_tool_{tool_name}_failed: {e}", exc_info=True)
         return [
--- a/backend/src/mcp_server/tools/__init__.py
+++ b/backend/src/mcp_server/tools/__init__.py
@@ -17,10 +17,15 @@ from typing import Any
 from uuid import UUID
 
 from mcp.types import TextContent
+from pydantic import ValidationError
 
 from mcp_server.tools._arg_coercion import coerce_mcp_arguments
 from mcp_server.tools._definitions import get_tool_definitions  # noqa: F401
-from mcp_server.tools._helpers import _error_response, _resolve_context_id
+from mcp_server.tools._helpers import (
+    _error_response,
+    _format_validation_error,
+    _resolve_context_id,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -409,6 +414,13 @@ async def execute_tool_call(
 
     try:
         return await handler(args, user_id, workspace_id)
+    except ValidationError as e:
+        # #1323: expected client-input errors — return the structured envelope
+        # (no pydantic model names / errors.pydantic.dev URLs) and log at
+        # warning without a traceback instead of an error-level exc_info dump.
+        message = _format_validation_error(e)
+        logger.warning(f"mcp_tool_{tool_name}_invalid_argument: {message}")
+        return _error_response("invalid_argument", message)
     except Exception as e:
         logger.error(f"mcp_tool_{tool_name}_failed: {e}", exc_info=True)
         return [
--- a/backend/src/mcp_server/tools/_arg_coercion.py
+++ b/backend/src/mcp_server/tools/_arg_coercion.py
@@ -82,6 +82,26 @@ def _coerce_to_boolean(value: Any) -> Any:
     return value
 
 
+def _coerce_any(value: Any) -> Any:
+    """Best-effort decode for schema fields declared WITHOUT a ``type`` ("any").
+
+    #1322: quirky clients JSON-stringify complex values for typeless fields
+    too — ``set_state``'s ``value`` arrived as ``"{\\"phase\\": ...}"`` /
+    ``"42"`` and was stored verbatim into JSONB, breaking the documented
+    round-trip. Decode strings that parse as JSON; keep everything else,
+    including strings that decode to ``None`` (a JSON null would violate
+    NOT NULL storage, and a literal ``"null"`` string is more plausibly
+    intentional than a stringified null).
+    """
+    if not isinstance(value, str):
+        return value
+    try:
+        decoded = json.loads(value)
+    except (ValueError, TypeError):
+        return value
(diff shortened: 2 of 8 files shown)


JFK:
## Focused code-review (correctness angles) — applied in the follow-up commit

**Confirmed (fixed):**
1. **Dispatch arm over-catch**: it wrapped the whole handler, so ValidationError from RESPONSE-model construction (server data-integrity bugs like the documented source_type Literal incident) would be misattributed as client `invalid_argument` and logged quietly. Now gated on the pydantic error title ending in `Request`; response-model failures keep the loud error+traceback path (regression test included).
2. **update_memory NotFound arm mislabel**: the upsert (external_id) path can raise the service's `NotFoundException("Context", …)` from the isolation gate — was returned as `memory_not_found: None`. Now re-cast to the uniform `context_not_found` deny (mirroring handle_recall's CWE-639 precedent); memory variant keys on memory_id or external_id; arm logs 404 usage telemetry like its siblings.
3. **_coerce_any scalar retyping**: decoding `"42"`→42 / `"true"`→true would corrupt legitimately-string values and diverge from REST (stores what was sent). Scoped to objects/arrays only — the actual reported round-trip bug. Bonus: `["array","null"]`-typed fields (reserved binding filters) now decode stringified lists (tested).

**Noted, deferred:** repo-wide error-slug unification (`validation_error` dominates the codebase; `invalid_argument` is the new dispatch-arm slug per #1323's clean example) — a full sweep would churn ~10 tools and their tests; out of scope here.

**Verified clean:** pydantic 2.12 `errors(include_url=False)` + v1 TypeError fallback; no test/frontend asserts the old file-storage message text.

PR #1330, opened 2026-07-16
fix(mcp): deterministic create_edge duplicate contract (#1321)

### LINKED ISSUE #1321: fix(mcp): create_edge silently upserts existing edges and can clobber declared links
## Summary

`create_edge` on an **existing (source, target) pair silently upserts** the existing edge instead of creating/failing — overwriting weight, confidence, and even `edge_type`, with no warning and no origin protection:

Observed on v0.51.1 (2026-07-16):
- A declared link created via `remember(linked_memory_ids=[…])` (origin="declared", weight 1.0 / confidence 1.0) was overwritten to weight 0.9 by a later `create_edge(…, edge_type="neural_association", weight=0.9)` — the declared provenance's fixed 1.0/1.0 signal is gone.
- A write-time auto-similarity edge was converted to `edge_type="supersedes"` by `create_edge(edge_type="supersedes")` on the same pair — `created_at` stayed at the original edge's timestamp, proving type-mutating upsert. (The supersede shadowing then works, but the original edge's history is rewritten.)

## Questions to settle (design)

1. Is upsert-on-duplicate intended? If yes, it should be documented in the tool description and the response should say `operation: "updated"` vs `"created"`.
2. Should `origin="declared"` edges be protected from silent weight/type overwrite by a generic create_edge (declared links are provenance, per #741)?
3. `update_edge` exists for mutation — create_edge doubling as update makes the two tools' contracts overlap.

## Suggested direction

Either (a) reject duplicates with `edge_exists` + pointer to update_edge, or (b) keep upsert but return `operation` and refuse to downgrade declared-origin edges without an explicit flag.

## Acceptance criteria

- Documented, deterministic duplicate behavior + test
- Declared-link provenance can't be silently clobbered

JFK:
## Summary

Closes #1321.

`create_edge` on an existing (source, target) pair silently upserted, overwriting a declared link's `edge_type`/`weight`/`confidence` with no warning — observed clobbering `remember(linked_memory_ids=…)` provenance (weight 1.0/1.0 → 0.9) and retyping an auto edge to `supersedes` in place.

## New MCP-level duplicate contract

| Existing edge | Behavior |
|---|---|
| none | insert → `operation: "created"` |
| auto edge (`origin` hebbian/semantic) | values applied → `operation: "updated"` + `previous` pre-image (hebbian is promoted to declared; semantic keeps its origin — visible via the new `edge.origin` response field) |
| declared, identical values | no write → `operation: "unchanged"` (timeout-retries stay idempotent) |
| declared, different values | error `edge_exists` echoing the existing edge; points to `update_edge` / `overwrite=true` |
| declared + `overwrite=true` | re-assert → `operation: "updated"` + `previous` |

The repository upsert is intentionally untouched — SDK/worker replay paths depend on its unconditional 3-col semantics. The non-overwrite arms pass `protect_declared_link=True` so a declared edge racing in between the pre-check and the upsert keeps its type/origin.

## Review (multi-agent adversarial, 5 angles × 3 refuters)

19 findings raised, 5 confirmed and fixed in the second commit:
- **critical**: `edge_exists` arm serialized the ORM instance after `db.rollback()` → MissingGreenlet on the async session; the 409 envelope was unreachable against a real DB. Snapshot now taken before rollback.
- **warning**: `overwrite` failed open on unrecognized strings (`bool("null") is True`); now fails closed with `validation_error`.
- **warning**: description over-promised origin promotion for semantic edges; wording fixed and `origin` added to edge responses (additive).
- 2 test-coverage gaps (semantic-origin update, `overwrite=true` fresh pair) closed.

## Testing

- 17 handler/schema tests in `tests/mcp_server/test_edge_tools.py` (5-scenario contract matrix + fail-closed overwrite + schema pins)
- mcp_server suite: 344 passed locally; edge-related suites (repositories/services/neural): 544 passed


JFK:
Addressed the Copilot review comment: all four created-edge mocks now carry `origin="declared"` (matching the handler's user-assertion contract), and both `operation: "created"` cases additionally assert `edge.origin == "declared"` so origin plumbing is pinned end-to-end.

PR #1345, opened 2026-07-17
fix(db): accept exponent coordinates in e69 location guard (#1344)

### LINKED ISSUE #1344: fix(db): e69 location generated columns NULL out exponent-notation coordinates on json details column
Final pre-deploy inspection of the v0.53.0 delta found a confirmed silent-invisibility bug in the #1331 WHERE axis, reproduced on PG 18.4.

## Root cause
- \`memories.details\` is PostgreSQL **json** (baseline migration \`sa.JSON()\`), not JSONB. json stores the inserted text **verbatim**.
- The app write path serializes via \`json.dumps\`, which renders any \`0 < |value| < 1e-4\` coordinate in exponent notation (e.g. \`5e-05\`) after \`normalize_location\`'s \`round(x, 7)\`.
- e69's generated-column regex guard \`^-?[0-9]+(\.[0-9]+)?$\` rejects exponent notation, so \`location_lat/lon\` become **NULL**: the write returns success, \`recall_nearby\` silently never returns the row. No 422, no log.
- The e69/model docstrings justify the regex with "JSONB renders numerics via numeric (never exponent notation)" — a false premise for this schema. The migration pin test masked the gap by inserting through \`CAST(:details AS JSONB)\`, which canonicalizes \`1e-7\` → \`0.0000001\` before it lands in the json column — a path the app never takes.

Affected band: any lat within ~11 m of the equator, any lon within ~11 m of the prime meridian (real coordinates, e.g. Greenwich).

## Fix
- Widen both regexes (e69 migration + \`models/memory.py\` Computed parity) to accept exponent notation: \`^-?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]{1,2})?$\` (\`'1e-07'::double precision\` casts cleanly — verified). The 2-digit exponent cap keeps any castable value inside double-precision range, preserving the "malformed → NULL, never error" guard contract.
- e69 is merged but **not applied in production** (prod deploys v0.53.0 today), so the migration is edited in place — no e70 needed. Local/CI DBs are rebuilt from scratch.
- Correct the false-premise docstrings (json verbatim storage, not JSONB canonicalization).
- Pin tests: add a \`CAST(:details AS JSON)\` app-fidelity insert path with exponent-notation pins alongside the existing jsonb-cast raw-SQL-defense pins; add a service-path e2e (remember → recall_nearby round-trip at lon=5e-05).

Found by the pre-deploy 13-angle review workflow + prod schema inspection (details column type + 0 legacy location rows, 5,725 rows / 27 MB).

Ref: #1331, #1341.

JFK:
Closes #1344

## Problem
`memories.details` is PostgreSQL **json** (not jsonb) — inserted text is stored verbatim. The app write path (`json.dumps`) renders any `0 < |value| < 1e-4` coordinate in exponent notation (`5e-05`), which the e69 generated-column regex `^-?[0-9]+(\.[0-9]+)?$` rejected → `location_lat/lon` silently NULL → the write succeeds but `recall_nearby` never returns the row (a ~11 m band along the equator / prime meridian, e.g. Greenwich).

The existing pin test masked this by inserting through `CAST(:details AS JSONB)`, whose jsonb round-trip canonicalizes `1e-7` → `0.0000001` before the json-column assignment — a path the app never takes. The e69/model docstrings' "JSONB renders numerics via numeric" premise is false for this schema.

## Fix
- Widen the guard in **both** e69 and the model Computed (parity pinned by `test_create_all_vs_alembic_drift.py`, passing) to `^-?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]{1,2})?$`. The 2-digit exponent cap keeps every accepted lexeme castable without double-precision overflow, preserving the "malformed → NULL, never error" contract.
- e69 edited **in place**: production has not applied it (v0.53.0 ships it today); local/CI DBs are rebuilt from scratch.
- Docstrings + design doc corrected to the json-verbatim reality.

## Tests (red→green verified)
- New `CAST(:details AS JSON)` app-fidelity insert path: `5e-05` / `1e-07` / `-1.23e-05` survive; `1e123` (3-digit exponent) NULLs instead of erroring.
- ORM-path e2e: memory 5 m east of the prime meridian written via the real serializer is found by `query_nearby_memories` at ~3.5 m.
- Both new pins **fail under the old regex** (verified against an old-regex DB) and pass with the fix.
- Full geo suite: 23 integration + 62 unit passed; lint & pyright clean (0 errors, identical to main).

## Prod evidence
- `information_schema`: prod `details` column type = `json`; rows with `details->'location'`: **0**; `memories`: 5,725 rows / 27 MB (e69 rewrite is trivial).

https://claude.ai/code/session_01CDnEwGpK7semEzr42hushN

### DIFF
--- a/backend/alembic/versions/e69_1331_location_cols.py
+++ b/backend/alembic/versions/e69_1331_location_cols.py
@@ -5,9 +5,14 @@ columns' TEXT trick (::timestamp is only STABLE), ``::double precision`` is
 IMMUTABLE, so these are real numeric STORED generated columns. The regex
 guard NULLs malformed details.location values instead of failing the INSERT
 (raw-SQL defense); MemoryService._apply_location's normalize_location is the
-writer-side contract (validation + 7-decimal fixed-point write-back) and
-JSONB renders numerics via ``numeric`` (never exponent notation), so
-well-formed writes always match the guard.
+writer-side contract (validation + 7-decimal fixed-point write-back).
+
+``memories.details`` is PostgreSQL json (NOT jsonb) — inserted text is
+stored VERBATIM, so numerics reach ``->>`` exactly as the writer serialized
+them. The app path (json.dumps) renders 0 < |value| < 1e-4 in exponent
+notation (``5e-05``), so the guard accepts exponent forms (#1344); the
+2-digit exponent cap keeps every accepted lexeme castable without
+double-precision overflow ("malformed -> NULL, never error").
 
 NOTE: adding STORED generated columns rewrites the table — check the prod
 ``memories`` row count before applying (e30_877 was the same shape and
@@ -30,7 +35,7 @@ def upgrade() -> None:
         r"""
         ALTER TABLE memories
         ADD COLUMN location_lat DOUBLE PRECISION
-        GENERATED ALWAYS AS (CASE WHEN details->'location'->>'lat' ~ '^-?[0-9]+(\.[0-9]+)?$'
+        GENERATED ALWAYS AS (CASE WHEN details->'location'->>'lat' ~ '^-?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]{1,2})?$'
         THEN (details->'location'->>'lat')::double precision ELSE NULL END) STORED
         """
     )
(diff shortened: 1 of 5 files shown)


JFK:
All 4 Copilot findings addressed in the follow-up commit: corrected the overflow rationale (1e123 is in double range — the cap prevents the 1e309 cast-error class and keeps 1e123 off the range CHECK), reworded the model/migration comments to stop overstating the guarantee, fixed the `:d` → `:details` docstring mismatch, and extended the cap pin to test both `1e123` and `1e309`. Behavior unchanged; 9/9 migration pins pass.

Issue #1403, opened 2026-07-20
Make supersedes first-class UX: auto-suggest supersession at remember time (it is the mechanism that carries update-correctness)


JFK:
## Evidence-backed product lever (prereg-v2 F3 factorial, v0.53.0)

The 2×2×2 mechanism factorial on the hardened update slice (N=100, one pre-registered run) attributes update-correctness on v0.53 as:

| mechanism | stale-suppression main effect (Holm-corrected) |
|---|---|
| **supersedes shadowing (#1208)** | **+0.7825** [0.7325, 0.830] — dominant |
| Sleep judged dedup/consolidation | +0.2175 [0.170, 0.2675] |
| reinforce recency | +0.0025 (ranking-only; conditional 0.76→0.99 but no suppression) |

With pre-authored supersedes edges, stale-fact suppression hits 100/100 and update_success@10 goes 0.76 → 1.00, with `stale_only` = 0 (the #1198 veto holds). Full numbers: "Two Wins, Two Nulls" §14.3 (kagura-memory-eval `paper/draft-full.md`; run artifact `results/v053/f3-run0-2026-07-19.json`).

## Proposal

The mechanism that actually wins is the one users must explicitly invoke (`remember(supersedes=…)` / `create_edge`). Make it first-class:

1. **Auto-suggest at remember time**: ingest-time k-NN neighbor seeding already finds near-duplicate candidates (in the F3 corpus it had pre-linked all 100 update pairs before any declaration). When a new memory's top neighbor exceeds a similarity threshold, surface "does this supersede X?" in the MCP response / clients (opt-in confirm, never auto-create — the sleep judge is deliberately barred from inventing supersession, and that judgment call should stay with the user).
2. SDK/docs: promote `supersedes` in remember examples and the tool description; today it reads as an afterthought.
3. Telemetry: count declared-supersedes usage so adoption of the mechanism is visible.

## Note on dose equivalence

The eval's smoke check confirmed declared-origin and semantic-origin supersedes edges shadow identically, so suggestion-then-confirm creates edges exactly as effective as dedup-authored ones.

---

## Addendum — pre-B de-risk: the auto-suggest detection signal is precise (2026-07-20)

[...]

Issue #1299, opened 2026-07-15
feat(auth): per-memory type/source binding filter enforcement (RFC-0002 P0-2 → P1)


JFK:
Split out of #1286 at the 2026-07-15 design gate. This is item 1 of the RFC-0002 P0 hardening residue (originally #1281 item 1).

**Decision recorded at the split: per-memory type/source filtering is a later phase (P1), not P0-2.** It is materially larger than the context-level gate #1275 shipped, and touches every agent read/write path on both REST and MCP. #1286 keeps only the tractable P0-4/P0-5 items; this issue carries the deferred enforcement.

## Background

`agent_context_bindings.allowed_memory_types` / `allowed_source_types` are forward-provisioned in the F1 DDL (`docs/design/agent-registry-and-bindings.md`, the `agent_context_bindings` table: `NULL` = all types, `[]` = none) and are listed in the P0-2 backward-compat matrix. But the **normative** F1 read/write-set is defined via `can_read` / `write_policy` only — the type/source arrays were never part of the normative read/write definition. #1275 shipped the context-level gate and **rejects non-`NULL` type/source filters at CRUD** (fail-closed), so no operator can configure a filter that silently does nothing.

## Design decision (records the F1 re-sign-off outcome for item 1)

The F1 doc was ambiguous between its normative section (context-level only) and the DDL / backward-compat matrix (type/source arrays reserved). **Resolved: type/source filtering is P1.** F1 stays authoritative that P0-2 is context-level only; the array columns remain schema-reserved and CRUD-rejected until this issue lands.

## Scope

- [ ] Annotate `docs/design/agent-registry-and-bindings.md`: mark `allowed_memory_types` / `allowed_source_types` explicitly as **P1** in the normative semantics (remove the normative-vs-matrix ambiguity), and re-point the backward-compat matrix line "*Type/source filters are schema-reserved but non-`NULL` values are rejected until #1286 lands*" → "…until this issue lands".
- [ ] Implement per-memory type/source filtering in `recall` / `reference` / `forget` / `explore` for **enforce-mode** agent credentials — **on both the REST and MCP chokepoints**. (The #1291 / #1292 lesson: a filter present only at the MCP handler layer silently leaks on the REST route for the same operation. Any per-memory filter must land on both faces or one leaks.)
[...]

Issue #1208, opened 2026-07-08
feat(api): supersedes/contradicts relations + recall shadowing


JFK:
## Background

The eval program confirmed update-correctness as this system's strongest measured win (+0.36 over vanilla RAG; `update_success@10` 0.96 vs 0.54 after the v0.44.0 winner veto). The code trace behind that experiment also established what the system does **not** have:

> There is no supersede/contradicts machinery for independently-stored documents.

Today only three levers can prefer a current fact at recall: (1) the reinforce recency re-rank (carried the whole +0.36), (2) Sleep judged dedup — **update-by-removal**: winner survives, loser is soft-deleted out of recall, (3) client `external_id` upsert. Update-by-removal became the *dominant success mode* in the v0.43.1 replication **and** produced the worst measured failure (stale-deletion at 10–18%, #1195, closed by the #1198 veto). Even fixed, the mechanism represents succession by deleting one version rather than by modeling it.

## Proposal

Make fact succession first-class and non-destructive:

1. **Typed relations**: extend `_ALL_EDGE_TYPES` (`backend/src/models/memory.py`) with `supersedes` and `contradicts`, reusing the #722 `origin` discriminator (`declared` when client-asserted, `semantic` when Sleep infers it). No new tables.
2. **Write path**: `remember(..., supersedes=<memory_id>)` / `update_memory` create the edge and demote (not delete) the predecessor.
3. **Sleep path**: for dedup clusters that look like update pairs (high overlap + clear recency ordering), record a `supersedes` edge and *shadow* the loser instead of soft-deleting it — configurable, removal stays available.
4. **Read path (shadowing)**: recall demotes/filters memories with an outgoing `supersedes` edge, annotates results (`superseded_by`), and supports `include_superseded=true` for audit. `contradicts` never auto-hides — both sides surface with a flag (unresolved contradiction is an arbitration problem, not a deletion problem).

## Acceptance criteria

- Frozen update-slice protocol (`backend/tests/eval/update_runner.py`): `update_success@10` ≥ the removal-based baseline (0.96), `stale_only` stays 0.
- **Over-supersede placebo**: a control set where naive linking would wrongly shadow *non*-superseded near-duplicates — shadowing must not reduce recall of distinct facts. (Without this honesty check, aggressive supersede inflates the metric.)
- History preserved: superseded memories reachable via `include_superseded` and `explore()`; deleting the edge restores full visibility.
- No regression on held-out P@5 (240-query corpus).

[...]

PR #1415, opened 2026-07-20
feat(mcp): supersede-candidate read exposure + acceptance (#1403)

JFK:
Partially addresses #1403

## Summary
- Completes the backend + read-time exposure for the #1403 supersede auto-suggest mechanism (the #1413 foundation only logged detection telemetry).
- Stores the top-1 near-duplicate detected at k-NN ingest in a dedicated server-only column `memories.supersede_candidate` (migration `e75`); lowers the detection threshold 0.92 → 0.85 (de-risked operating point: precision==recall==1.0 across τ 0.60–0.88).
- Surfaces it, liveness-guarded, on `recall()` / `reference()` / `patch_memory()` via a new `SupersedeCandidate` schema field so a client can offer a confirm → `create_edge(supersedes)` action — never auto-created (#1208 over-supersede prevention intact).
- `create_edge(supersedes)` matching a stored suggestion emits `supersede_suggestion_accepted` and clears the column (self-heal), in the same transaction as the edge.

## Implementation notes
- **Dedicated column, not the client-writable `details` blob**: a security review found that storing it in `details` enabled an IDOR (a client could forge the target `memory_id` and leak another memory's `summary` via the read-path enrichment, bypassing `can_access_memory`), a fail-open break (poison `detected_at` → `ValidationError` blanks recall), and data-loss on a `details` replace. The column has exactly two server-only writers.
- `_resolve_supersede_candidates`: one batched query per recall page (after the top-k slice), with three guards before exposing a target's `summary` — liveness / not-deleted (self-healing, mirrors `#1208`), same `(workspace, context)` as the source, and the `#1299` agent-binding row filter (the same boundary every other recall lane enforces). Fail-open per item and overall.
- `expire_on_commit=False` verified for the mid-task commit in `_create_knn_seed_edges`.
- Migration is additive + nullable (no backfill); single alembic head (`e75`).

## Pre-PR review summary
- gate2 mode: advisor-only
- audit: skipped
- binary_gate: (none)
- cso: green
- qa-lead: green
- cto: green
- review provider: code-review (post-PR loop via /gh-issue-driven:review)

Also reviewed twice by the code-reviewer agent (2 Critical + 2 Warning fixed: client-write leak, fail-open ValidationError, patch drop, agent-binding bypass). Full gate2 review saved in the plugin cache.

## Verification
- 205 targeted unit tests pass (k-NN storage/threshold, resolver liveness/cross-context/binding/malformed/fail-open, acceptance/self-heal). ruff + format clean; zero new pyright errors.
- api/integration tests require Docker (Redis/Postgres) — CI is the authoritative gate.

## Remaining (not in this PR)
- Client-side confirm → `create_edge` loop (agent/UX consumer).
- Follow-up tests: agent-binding drop (with an agent scope set), exactly-0.85 boundary, live-DB recall wiring.

### DIFF
--- a/backend/alembic/versions/e75_1403_supersede_candidate.py
+++ b/backend/alembic/versions/e75_1403_supersede_candidate.py
@@ -1,0 +1,44 @@
+"""Add memories.supersede_candidate (server-only supersede suggestion).
+
+Issue #1403 (option B): the k-NN cold-start seeding detects, at ingest, when a
+new memory's top-1 neighbor is a near-duplicate (cosine >= the supersede-suggest
+threshold) — i.e. the new memory likely supersedes an existing fact. That
+suggestion is stored here and surfaced, liveness-guarded, on recall()/reference()
+so a client can offer a confirm -> create_edge(edge_type="supersedes") action
+(never auto-created — the #1208 over-supersede prevention stays intact).
+
+A DEDICATED column rather than a key inside the client-writable ``details`` JSON:
+storing it in ``details`` would let a client forge the ``memory_id`` (leaking
+another memory's summary through the read-path enrichment, bypassing the
+per-memory access check) or silently drop it on a ``details`` replace. This column
+is written only by the server (the async k-NN seeding) and cleared only when the
+suggested supersedes edge is created.
+
+Additive + nullable: existing rows read as NULL (no suggestion), no backfill.
+
+Revision ID: e75_1403_supersede_candidate
+Revises: e74_1401_mae_explore
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e75_1403_supersede_candidate"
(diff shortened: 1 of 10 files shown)


JFK:
Addressed both Copilot review points in `2e8a70f8`:

1. **Detection coupled to seed threshold** — `_create_knn_seed_edges` now detects the supersede candidate from the **raw top-1 neighbor (excluding self)**, before/independent of the calibrated/operator seed-edge threshold. An operator `knn_seed_min_similarity` > 0.85 no longer filters a valid suggestion (e.g. 0.87) out before the check. Seed-edge creation still applies the calibrated threshold. Added a regression test (`test_supersede_candidate_independent_of_seed_threshold`).
2. **patch_memory operation** — now threads `operation="update"` (not `"reference"`) into the agent-binding row filter, so a shadow-mode `would_deny` is attributed to the correct MAE operation for the PATCH surface.

49 targeted unit tests pass; ruff + format clean.

PR #1438, opened 2026-07-25
refactor(api): decompose MemoryService.recall (828 lines / cx 80) (#1437)

JFK:
Closes #1437.

`MemoryService.recall()` was **828 lines / cyclomatic complexity 80** — the most
complex function in `backend/src` by a wide margin (next was 59) and the hottest,
most security-sensitive read path in the product. Every authorization gate
(#1291 agent binding, #708 BYOK, #887 trust tier, #1299 row filter) was inlined
into one body, so any change to it meant re-reading all 828 lines.

## What changed

Extracted the phase boundaries that were **already marked by comment banners**
into named private helpers, leaving `recall()` as an orchestrator that calls each
gate in exactly the same order:

| Helper | Concern |
|---|---|
| `_recall_prepare` | #708 paying workspace, #1220 config prefetch, #1212 search-mode resolution |
| `_recall_check_agent_bindings` | #1291 subtractive binding gate |
| `_recall_resolve_cluster_filter` | #496 `analysis_cluster` parse + resolve |
| `_recall_enforce_shared_byok` | #708 Option A H1 deferred BYOK gate |
| `_recall_hydrate_memories` | PG hydration + #214/#887/#496 filters, #972 dual keying |
| `_recall_filter_rows_by_binding` | #1299 per-row binding filter |
| `_recall_apply_selection_plan` | #1306 evidence + pool reorder |
| `_recall_build_responses` | #1403 supersede hints + `MemoryResponse` rows |
| `_recall_run_hebbian_learning` | #982/#983 neural graph update |
| `_recall_finalize` | hints + tags + confidence + audit row |
| `_empty_recall_response` | the deduplicated zero-candidate exit |

Plus a `_RecallPlan` NamedTuple carrying the resolved query parameters.

## Defects fixed in passing

1. **Duplicated empty-response construction.** The 18-line `RecallResponse(results=[], …)`
   block was written out **twice, character-identical** — the empty-`analysis_cluster`
   short-circuit and the no-search-hits exit. Any field added to the empty-recall
   contract had to be added in both places or the two exits would silently diverge.
   Now one helper.
2. **Guard/value drift in the confidence filter.** The comprehension guarded with
   `r.get("hybrid_score", r.get("score"))` but read `r.get("hybrid_score", r["score"])`.
   Safe *only* because the `if` short-circuits the `KeyError` case — the two had to be
   kept in lockstep by hand. Replaced with a single walrus binding.
[...]

PR #1441, opened 2026-07-26
fix(mcp): sleep rollback logger TypeError + setup_resource None return (#1440)

### LINKED ISSUE #1440: fix(mcp): sleep rollback TypeError + setup_resource None return
Two functional bugs in `backend/src/mcp_server/tools/`, both surfaced by running
`make type-check` (`pyright src/`) — which is configured in this repo but **not run by CI**.
Neither is caught by ruff or by the 6 433-test suite.

Found while auditing for #1437; see the correction comment there.

---

## Bug 1 — `sleep.py:400` raises `TypeError` during sleep-run rollback

`mcp_server/tools/sleep.py:24` binds the **stdlib** logger:

```python
logger = logging.getLogger(__name__)
```

but line 400 calls it with structlog-style kwargs:

```python
logger.warning(
    "shadow_merge_rollback_edge_mismatch",
    src_id=str(action.memory_id),
    dst_id=str(action.target_id),
)
```

`logging.Logger._log()` accepts only `exc_info` / `stack_info` / `stacklevel` / `extra`, so this
raises `TypeError: Logger._log() got an unexpected keyword argument 'src_id'`.

**Impact.** It sits on the `else` branch of `if reverted:` inside `handle_rollback_sleep_run`. When a
shadow-merge edge does not match the recorded `prior_edge`, the handler is supposed to log a warning
and carry on tallying the rollback. Instead the rollback itself dies with a `TypeError` — the
degraded-but-recoverable path becomes a hard failure, and the operator loses the rest of the
rollback summary.

**This is a recurrence.** `api/routes/auth.py:50` carries a comment documenting the identical bug
class being fixed in PR #522:

> auth.py historically bound logger via stdlib `logging.getLogger` while every other module in the
> project uses `utils.logger.get_logger` (structlog). The mismatch meant `logger.info("event",
> key=value)` calls anywhere in this file would TypeError at runtime — only the f-string style
> worked.

`.claude/rules/backend.md` also states the rule: *"Use `structlog` logger, never `print()`"*.

**Fix.** Bind via `utils.logger.get_logger` like the rest of the codebase. The other `logger.*` calls
in the module are f-string + `exc_info=True` style, which structlog accepts unchanged.

**Scope note.** 31 other modules under `backend/src` still bind `logging.getLogger`, but an AST sweep
shows **`sleep.py:400` is the only site passing non-stdlib kwargs** — i.e. the only one that actually
breaks. Converting the other 31 is not in scope here; preventing recurrence is tracked separately.

---

## Bug 2 — `resource.py:handle_setup_resource` can return `None`

Declared `-> list[TextContent]`, but the body is:

```python
async for db in get_db():
    try:
        ...
    except ...:
        return _error_response("setup_resource_error", error_str)
# <- nothing here
```

If the `get_db()` generator yields nothing, control falls off the end and the function returns
`None` against its declared `list[TextContent]`.

**All five sibling handlers in the same file get this right** — `handle_ingest_events` and four
others each close with:

```python
    return _error_response("internal_error", "Database session unavailable")
```

at lines 291, 401, 533, 687 and 1129. `handle_setup_resource` (690–967) is the one that was missed.
A `None` return propagates into the MCP transport where a `list[TextContent]` is expected, so the
failure surfaces far from its cause rather than as the intended structured error envelope.

**Fix.** Add the same trailing `return`, matching its siblings exactly.

`pyright`'s `reportReturnType` is whole-program and flags exactly one such handler, so this is the
only occurrence of the pattern in the file — no wider sweep needed.

---

## Acceptance criteria

- [ ] A test asserting `logger.warning(...)` with structlog kwargs on the sleep rollback
      edge-mismatch branch does not raise (fails before the fix).
- [ ] A test asserting `handle_setup_resource` returns an error envelope, not `None`, when
      `get_db()` yields nothing (fails before the fix).
- [ ] `pyright src/` no longer reports `mcp_server/tools/sleep.py:402`, `:403` or
      `mcp_server/tools/resource.py:692`.
- [ ] Existing suite green, `make lint` clean.

JFK:
Closes #1440.

Two functional bugs, both surfaced by running `make type-check` (`pyright src/`) — configured in
this repo but **not run by CI**. Neither is caught by ruff or by the 6 433-test suite.

## Bug 1 — `sleep.py` misreports a benign edge mismatch as a failed rollback

`mcp_server/tools/sleep.py:24` bound the **stdlib** logger, then called it with structlog-style
kwargs on the shadow-merge rollback edge-mismatch branch:

```python
logger = logging.getLogger(__name__)          # stdlib
...
logger.warning("shadow_merge_rollback_edge_mismatch", src_id=..., dst_id=...)
```

`logging.Logger._log()` rejects unknown kwargs → `TypeError`. A per-action `except` catches it into
`rollback_summary["errors"]`, so the observed behaviour was:

```json
{
  "error": "partial_rollback",
  "message": "Rollback completed with 1 error(s). Report marked as 'failed' — inspect errors and retry if needed.",
  "rollback_summary": {
    "merges_reversed": 0,
    "errors": ["Action <...> (merge): Logger._log() got an unexpected keyword argument 'src_id'"]
  }
}
```

An edge mismatch is a *"log it and carry on"* condition. Instead the operator is told the rollback
**failed**, the report is marked `failed`, and the recorded reason is Python internals rather than
the mismatch — so "inspect errors and retry" points at a message that says nothing about the cause.
The rest of the rollback does complete; it is not aborted. (My original issue text said the rollback
"dies" — corrected [in a comment](https://github.com/kagura-ai/memory-cloud/issues/1440#issuecomment-5082366591).)

**This is a recurrence.** `api/routes/auth.py:50` carries a comment documenting the identical bug
class being fixed in PR #522, and `.claude/rules/backend.md` states the structlog rule.

Fix: bind via `utils.logger.get_logger`. **Blast radius is exactly two call sites** — `sleep.py` has
only the broken one and one positional f-string `logger.warning(...)`, which structlog accepts
unchanged. `import logging` is removed; no other `logging.*` usage remained.

## Bug 2 — `handle_setup_resource` returns `None`

Declared `-> list[TextContent]`, but nothing followed its `async for db in get_db():`, so an
exhausted generator fell off the end and returned `None` into the MCP transport — surfacing far from
the cause instead of as the intended error envelope.

The **five sibling handlers in the same module** all close with
`return _error_response("internal_error", "Database session unavailable")` (lines 291, 401, 533, 687,
1129). This one was missed. pyright's `reportReturnType` is whole-program and flags exactly one
handler, so this is the only occurrence — no wider sweep needed.

## Tests — written first, confirmed red

```
tests/mcp_server/test_sleep_tools.py::TestShadowMergeRollbackEdgeMismatch
tests/mcp_server/test_resource_tools.py::TestDatabaseSessionUnavailable
```

Before the fix:

```
AssertionError: the logger.warning(...) call raised instead of logging:
  ["Action <...> (merge): Logger._log() got an unexpected keyword argument 'src_id'"]
AssertionError: handler returned None instead of an error envelope
```

The sleep test drives the real `handle_rollback_sleep_run` down to the mismatch branch (patching
`revert_shadow_merge_edge` to return `False`) and asserts no `Logger._log()` string reaches
`rollback_summary["errors"]`, that the run is **not** reported as `partial_rollback`, that `errors`
is empty, and that the mismatch is still **not** counted as a reversal — so the fix cannot pass by
silently treating a mismatch as success.

`TestDatabaseSessionUnavailable` also pins `handle_ingest_events` (already correct) alongside the
fixed handler, so the pair cannot drift apart again.

## Verification

- New tests red before / green after; `2767 passed, 374 skipped` across the unit suites
  (`tests/services tests/mcp_server tests/repositories tests/neural tests/models tests/db` + MCP
  tool tests) — no pre-existing test modified.
- `ruff check src/ tests/` clean, `ruff format` clean.
- `pyright` no longer reports `sleep.py:402`, `:403`, or `resource.py:692`. The remaining
  `sleep.py:428 rowcount` error is a separate Tier-A/B residual, handled in the follow-up that adds
  pyright to CI.
**On review:** the codex-in-WSL second opinion used on #1438 **did not complete here** — WSL hung and
I killed it rather than wait. So the questions I would have asked it, I verified directly instead:

- *Does swapping the logger change behaviour at the other call sites?* No — `sleep.py` has exactly
  **two** `logger.*` calls: the fixed one and a positional f-string `logger.warning(...)`, which
  structlog's `BoundLogger` accepts unchanged. `import logging` is removed and no `logging.*` usage
  remains (ruff F401 confirms).
- *Could the new `return` mask a real error?* No. An AST check of `handle_setup_resource` shows every
  terminal path inside the `async for` body returns or raises, so the fall-through is reachable
  **only** when `get_db()` yields nothing — exactly the path that previously returned `None`.
- *Do the tests pass for the wrong reason?* Both were confirmed red before the fix with the specific
  assertions above. An instrumented run proves the sleep test reaches the branch
  (`revert_shadow_merge_edge` called once, the `TypeError` observed in `rollback_summary["errors"]`),
  and the resource test's pre-fix failure was `handler returned None`, which is only reachable at the
  fall-through.
- *Do the patch targets work for function-body imports?* Yes — both handlers import inside the
  function, so the attribute is resolved from the patched module at call time; the red→green
  transition demonstrates it.

A second opinion on this diff is still welcome in review.

## Follow-ups

Adding `pyright src/` as a CI gate (the reason these went unnoticed) and an automated guard against
this stdlib-logger bug class recurring a third time are tracked separately under v0.61.0.

### DIFF
--- a/backend/src/mcp_server/tools/resource.py
+++ b/backend/src/mcp_server/tools/resource.py
@@ -966,6 +966,12 @@ async def handle_setup_resource(
             )
             return _error_response("setup_resource_error", error_str)
 
+    # #1440: an exhausted ``get_db()`` generator must still produce the declared
+    # ``list[TextContent]``. Without this the function fell off the end and
+    # returned None into the MCP transport. Mirrors the fall-through every other
+    # handler in this module already has.
+    return _error_response("internal_error", "Database session unavailable")
+
 
 async def handle_setup_connector(
     args: dict[str, Any], user_id: str, workspace_id: UUID | None
--- a/backend/src/mcp_server/tools/sleep.py
+++ b/backend/src/mcp_server/tools/sleep.py
@@ -3,7 +3,6 @@
 Issue #164: get_sleep_history, get_sleep_report, rollback_sleep_run.
 """
 
-import logging
 import time
 from typing import Any
 from uuid import UUID
@@ -20,8 +19,16 @@ from mcp_server.tools._helpers import (
     _resolve_context_id,
     _success_response,
 )
+from utils.logger import get_logger
 
-logger = logging.getLogger(__name__)
+# #1440: bind through ``utils.logger.get_logger`` (structlog) like the rest of
+# the project. This module previously used stdlib ``logging.getLogger``, whose
+# ``Logger._log()`` rejects unknown kwargs — so the structlog-style
+# ``logger.warning("event", key=value)`` call in the shadow-merge rollback path
+# raised TypeError and was swallowed into ``rollback_summary["errors"]``,
+# turning a benign edge mismatch into a reported ``partial_rollback``. Identical
+# bug class to the one fixed in ``api/routes/auth.py`` (PR #522).
+logger = get_logger(__name__)
 
 
 def _report_to_summary(report: Any) -> dict[str, Any]:
--- a/backend/tests/mcp_server/test_resource_tools.py
+++ b/backend/tests/mcp_server/test_resource_tools.py
@@ -922,3 +922,49 @@ class TestSetupResourceHappyPath:
             )
         data = _json_of(result)
         assert data["error"] == "resource_id_conflict"
+
+
+class TestDatabaseSessionUnavailable:
+    """#1440: every handler must return an error envelope when get_db() yields nothing.
+
+    Each handler declares ``-> list[TextContent]`` and drives its work inside
+    ``async for db in get_db():``. If that generator yields nothing, control
+    falls off the end and the function returns ``None`` — which then propagates
+    into the MCP transport where a list is expected, surfacing far from the
+    cause. Five of the six handlers in this module close with the
+    ``internal_error`` fall-through; ``handle_setup_resource`` was missed.
+    """
+
+    @pytest.mark.asyncio
+    async def test_setup_resource_returns_envelope_not_none(self):
+        async def empty_get_db():
+            return
+            yield  # pragma: no cover - makes this an async generator
+
+        with patch("db.base.get_db", new=empty_get_db):
+            result = await handle_setup_resource(
+                {"name": "ctx-1440", "resource_id": "res1440"}, "user", uuid4()
+            )
+
+        assert result is not None, "handler returned None instead of an error envelope"
+        data = _json_of(result)
+        assert data["status"] == "error"
+        assert data["error"] == "internal_error"
+
+    @pytest.mark.asyncio
+    async def test_ingest_events_already_has_the_guard(self):
+        """Pins the sibling that was already correct, so the pair cannot drift."""
+
+        async def empty_get_db():
+            return
+            yield  # pragma: no cover - makes this an async generator
+
+        with patch("db.base.get_db", new=empty_get_db):
+            result = await handle_ingest_events(
+                {"resource_id": "res1440", "events": [{"op": "upsert", "doc_id": "d1"}]},
+                "user",
+                uuid4(),
+            )
+
+        assert result is not None
+        assert _json_of(result)["error"] == "internal_error"
--- a/backend/tests/mcp_server/test_sleep_tools.py
+++ b/backend/tests/mcp_server/test_sleep_tools.py
@@ -494,3 +494,97 @@ class TestRollbackSleepRun:
         data = json.loads(result[0].text)
         assert data["status"] == "error"
         assert data["error"] == "no_actions"
+
+
+class TestShadowMergeRollbackEdgeMismatch:
+    """#1440: the edge-mismatch branch must log a warning, not fail the rollback.
+
+    ``mcp_server.tools.sleep`` bound the STDLIB logger (``logging.getLogger``)
+    while calling it with structlog-style kwargs, so this branch raised
+    ``TypeError: Logger._log() got an unexpected keyword argument 'src_id'``.
+    The per-action ``except`` swallowed it into ``rollback_summary["errors"]``,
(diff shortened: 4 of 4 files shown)


JFK:
Correction to the **On review** paragraph in this PR's description (posted after merge, for the record).

That paragraph says the codex second-opinion review "did not complete here — WSL hung and I killed it
rather than wait." **The cause was wrong.** The review never started at all: I invoked it as
`wsl bash -lc 'codex exec ... "$(cat prompt.txt)"'`, and the prompt file contained markdown
backticks around code samples. Inside double quotes, bash executed those as command substitution and
died immediately:

```
bash: command substitution: line 1: syntax error near unexpected token '('
bash: command substitution: line 1: 'logger = logging.getLogger(__name__)'
bash: command substitution: line 1: 'async for db in get_db():'
```

The process had already exited; what looked like a hang was just an empty output file. (The
background task only surfaced this days later, which is how it came to light.)

**What does not change:** the review genuinely did not run, so the verification recorded in the
description stands on its own — and it was done directly, not inferred:

- logger swap blast radius enumerated (exactly two `logger.*` call sites in `sleep.py`);
- the new `return` proven reachable only on an exhausted `get_db()` via an AST check of every
  terminal path in the loop body;
- both tests confirmed red before the fix, with the specific failure strings quoted.

@Copilot also reviewed this PR and generated no comments.

I cannot re-run codex now — WSL itself no longer starts on this machine
(`Wsl/Service/0x80370107`). Flagging it here rather than leaving an inaccurate tooling claim in the
record; happy to run a second opinion over `821a82ee` if that path is restored.

Issue #612, opened 2026-05-12
chore(types): tier-B pyright re-enable — reportCallIssue + reportAttributeAccessIssue (~58 fixes)


JFK:
## Context

Follow-up to #599 Tier A (closed via PR #610). Tier A re-enabled `reportAssignmentType` (4 fixes) + `reportReturnType` (5 fixes) = 9 caller-side narrowings.

Tier B is the next intermediate slice of the staged pyright re-enable plan documented in `backend/pyproject.toml`:

```
Tier  Rule                         Residuals (pre-fix)  Status
----  --------                     -------------------  ------
A     reportAssignmentType                4             re-enabled (#599 / PR #610)
A     reportReturnType                    5             re-enabled (#599 / PR #610)
B     reportCallIssue                    22             deferred → THIS ISSUE
B     reportAttributeAccessIssue         36             deferred → THIS ISSUE
C     reportArgumentType                163             deferred (separate issue)
D     reportGeneralTypeIssues             ?             deferred (separate charter)
D     reportOptionalMemberAccess          ?             deferred
D     reportIndexIssue                    ?             deferred
D     reportOptionalOperand               ?             deferred
```

## Scope

Re-enable two pyright rules (~58 caller-side narrowings):

- **`reportCallIssue`** — 22 residual errors. Pyright complaints about call-site signatures (positional/keyword arg counts, missing required args, unexpected kwargs).
- **`reportAttributeAccessIssue`** — 36 residual errors. Pyright complaints about attribute access on possibly-None / unknown-type / wrong-type values.

## Acceptance criteria

- [ ] Audit: run `pyright src/` with each rule individually flipped to `true`, capture per-file error list
- [ ] Fix 22 + 36 = 58 caller-side narrowings (cast / explicit None checks / signature corrections)
- [ ] Remove `reportCallIssue = false` and `reportAttributeAccessIssue = false` from `backend/pyproject.toml`
- [ ] Add explicit `reportCallIssue = true` and `reportAttributeAccessIssue = true` (defense-in-depth, see Kagura memory `348846d6` — implicit-default avoidance)
- [ ] Update audit comment table to mark Tier B as re-enabled
- [ ] Verify `pyright src/` shows 0 new errors after rule flip (1 pre-existing `reportOperatorIssue` in `neural_tasks.py:263` is out-of-scope)
- [ ] `pytest tests/services tests/utils tests/auth tests/mcp_server` continues to pass

## Out of scope

[...]

PR #1444, opened 2026-07-26
build(ci): clear pyright residuals and gate type-check in CI (#1442)

### LINKED ISSUE #1442: build(ci): clear pyright residuals and gate type-check in CI
`pyright` is configured in this repo — `[tool.pyright]` in `backend/pyproject.toml` with a documented
tier-by-tier re-enablement plan, `pyright>=1.1.0` in the `dev` extra, and a `make type-check` target —
but **nothing runs it.** The CI `lint` job is `ruff check` + `ruff format --check` +
`lint-models-no-column` only.

The frontend does not have this gap: `ci.yml` runs `npx tsc --noEmit`. The backend is the asymmetry.

## Why this matters now

`make type-check` is currently **red**, and two of its findings were real functional bugs that ruff
and the 6 433-test suite both missed — shipped as #1440 / PR #1441:

- a stdlib-vs-structlog logger mismatch that turned a benign sleep-rollback edge mismatch into a
  reported `partial_rollback` with a Python-internals error message;
- an MCP handler returning `None` against a declared `list[TextContent]`.

A configured gate that nothing enforces drifts without bound. These two are what that drift cost.

## Current state

With #1440's fixes applied, `pyright src/` reports **18 errors** (excluding 2 that are purely local
environment — see below). Triage:

**Stub/inference limitations — safe, want a narrow documented suppression or cast:**

| Location | Rule | Why it is not a bug |
|---|---|---|
| `db/redis.py:69` | `reportReturnType` | client is built with `decode_responses=True`, so `.get()` really does return `str \| None`; the redis-py stub is not parameterized on that flag |
| `services/sleep/dedup_merge.py:161` | `reportOperatorIssue` | the `(*sorted(...), p[2])` splat widens every tuple slot to `UUID \| float`; `p[2]` is the score at runtime |
| `neural/hebbian.py:138` | `reportOperatorIssue` | `repetition_active` is defined as `... and floor_threshold is not None`, so the guard holds — pyright cannot narrow through the intermediate bool |
| `mcp_server/tools/sleep.py:428`, `services/sleep/undo.py:90` | `reportAttributeAccessIssue` | `.rowcount` needs `CursorResult`, not `Result[Any]` |
| `api/routes/contexts.py:1510`, `services/account_erasure_service.py:946`, `services/workspace_service.py:717` | `reportAttributeAccessIssue` | assignment to a `Mapped[...]` attribute |
| `services/llm_pricing_service.py:66` | `reportAssignmentType` | `TTLCache` generic parameters |
| `services/memory_service.py:1968` (×2) | `reportAssignmentType`, `reportCallIssue` | `dict(rows.all())` inference from a SQLAlchemy `Row` sequence |
| `services/secret_store_service.py:909/916/924` (×3) | `reportCallIssue` | overload resolution |

**Needs investigation — not yet explained:**

- `api/routes/auth.py:1043/1089/1130/1136` (×4) — `"github_auth_url" is not a known attribute of
  module "auth"`. The functions are plainly defined at `auth/oauth_endpoints.py:113–126`, the import
  is `from auth import oauth_endpoints`, and there is no shadowing or dynamic export. All four
  `oauth_endpoints.*` call sites in the file are flagged and there are no others, so this looks like
  a submodule-resolution issue rather than four separate ones. Resolve the cause before suppressing.

**Local-environment only (not counted in the 18):** `services/llm_providers/gemini_provider.py:76`
and `ollama_cloud_provider.py:44` (`google.genai`, `ollama`).

## The gate needs an install-extras decision

`google-genai` and `ollama` are declared in the **`llm_providers`** optional extra, not in `dev`. A
CI job that installs only `.[dev]` will therefore reproduce those two `reportMissingImports` and the
gate will fail for a reason unrelated to code quality. Options:

1. install `.[dev,llm_providers,neural]` in the lint job (slower, but pyright then sees what
   production sees);
2. keep the lean install and add the optional-provider modules to pyright's `ignore`;
3. keep the lean install and set `reportMissingImports = false` — **not recommended**, it is the rule
   that catches genuinely missing first-party imports.

Option 1 is the honest one if lint-job runtime allows; otherwise 2, scoped to exactly those modules.

## Scope

- [ ] Explain and fix the `api/routes/auth.py` cluster (4).
- [ ] Clear the remaining 14 with narrow, **individually commented** suppressions or real type fixes.
      No blanket `reportX = false` additions to `[tool.pyright]` — that would re-open the gap.
- [ ] Decide the extras question above and wire `pyright src/` into the CI `lint` job as a **hard
      gate** (not `continue-on-error`).
- [ ] `make type-check` green from a clean checkout.

Deliberately **out of scope**: Tier C (`reportArgumentType`, ~163 residuals) and Tier D remain
disabled in `[tool.pyright]`. This issue gates the tiers that are already meant to be on; raising the
tier is a separate decision.

## Ordering

Depends on #1440 / PR #1441 — three of the errors are fixed there, and the gate cannot go green
without it.

### LINKED ISSUE #1443: fix(api): Gemini list_models always returns the hardcoded fallback
`GeminiProvider.list_models()` never returns live models. It always falls back to the hardcoded pair.

## Cause

`backend/src/services/llm_providers/gemini_provider.py:128`:

```python
response = await self._ensure_client().aio.models.list()
return [
    {...}
    for m in response.models          # <- AttributeError
    if getattr(m, "name", None)
]
```

`aio.models.list()` returns a **`google.genai.pagers.AsyncPager`**, which has no `.models`
attribute — it *is* the async-iterable of models:

```
AsyncPager attrs: ['config', 'name', 'next_page', 'page', 'page_size', 'sdk_http_response']
has .models?   False
async-iterable? True
```

(verified against `google-genai` 2.14.0; `pyproject.toml` constrains only `>=1.0.0`, so current
installs get 2.x)

The `AttributeError` is caught by the surrounding `except Exception`, which logs
`gemini_list_models_failed` and returns:

```python
[{"id": "gemini-3.1-pro", ...}, {"id": "gemini-3.1-flash-lite", ...}]
```

So the failure is **silent from the caller's perspective** — it receives a plausible, well-formed
list that is simply never the live one.

## Impact

BYOK users on Gemini always see the two hardcoded models regardless of what their key can actually
reach — no newly released model, no restricted-access model, no model their project has been granted.

## Why tests did not catch it

`tests/services/test_llm_providers.py::TestGeminiProvider::test_list_models_returns_models` mocks the
return value as `MagicMock()` with `.models` assigned. The test asserts against a **fabricated API
shape the SDK does not have**, so it passes while production always takes the fallback path. The test
needs to be rewritten against an async-iterable pager stub.

## How it surfaced

`pyright` reports `Cannot access attribute "models" for class "AsyncPager[Model]"` — but **only when
`google-genai` is installed**. With the lean `.[dev]` install it instead reports an unresolved
import, masking the real finding. This is direct evidence for installing the `llm_providers` extra
in the type-check gate (#1442).

## Fix

Iterate the pager (`async for m in response`) — the documented idiom — and rewrite the test to use an
async-iterable stub so the real contract is pinned.

JFK:
Closes #1442. Closes #1443.

> **Stacked on #1441** (`1440-fix/...`). Three of the errors are fixed there and the gate cannot go
> green without it — **merge #1441 first**.

`pyright` was configured in this repo — `[tool.pyright]`, `pyright` in the `dev` extra, a
`make type-check` target, and a documented tier-by-tier plan (#599, #612) — but **nothing ran it**.
The frontend has run `tsc --noEmit` in CI since day one; the backend had no equivalent, so the gate
drifted red. Two of the findings it had accumulated were real functional bugs (#1440). Clearing the
rest surfaced a third (#1443).

## The gate

A new **`type-check` job**, as a **hard gate**. Not `continue-on-error` — a permanently-red check
nobody reads would have surfaced these bugs exactly as well as no check at all. Kept separate from
`lint` so that job stays ruff-only and fast.

**The install line is load-bearing:**

```yaml
run: cd backend && pip install -e ".[dev,llm_providers,neural]"
```

`google-genai` and `ollama` live in the `llm_providers` extra. With a lean `.[dev]` install, pyright
reports them as unresolved imports and **stops analysing those modules** — which is exactly how
#1443 stayed hidden. Installing what production imports is what makes the gate mean something.

## #1443 — folded in because it blocks the gate

`GeminiProvider.list_models()` read `response.models`, but `aio.models.list()` returns an
`AsyncPager`, which **is** the async-iterable and has no `.models`:

```
AsyncPager attrs: ['config', 'name', 'next_page', 'page', 'page_size', 'sdk_http_response']
has .models?    False
async-iterable? True          # verified against google-genai 2.14.0
```

The `AttributeError` was swallowed by the surrounding `except Exception`, so **every** call silently
returned the hardcoded `["gemini-3.1-pro", "gemini-3.1-flash-lite"]` fallback. BYOK users on Gemini
never saw their actual models — and the response was well-formed, so nothing looked wrong.

**Why the test missed it:** it stubbed the result as `MagicMock(models=[...])` — an API shape the SDK
does not have. It asserted against a fabricated contract. Rewritten against an async-iterable stub;
confirmed red first:

```
gemini_list_models_failed  error="'_AsyncPagerStub' object has no attribute 'models'"
AssertionError: expected the live model list; the hardcoded fallback means the pager was not iterated
```

## Residuals cleared (17)

Each carries an inline reason. **No blanket `reportX = false` was added to `[tool.pyright]`** — that
would re-open the gap this PR closes.

| Site | Was it a bug? | Resolution |
|---|---|---|
| `api/routes/auth.py` (×4) | No, but a real ambiguity | This file is itself `auth.py`, so `from auth import oauth_endpoints` is ambiguous from inside `api/routes/` and pyright resolved it to the **sibling**. Runtime always resolved correctly (absolute import). Spelled as an explicit submodule import — unambiguous for readers too. |
| `neural/hebbian.py` | No | `repetition_active` already implies `floor_threshold is not None`; restate it so narrowing survives |
| `sleep.py`, `sleep/undo.py` | No | `.rowcount` needs `CursorResult` — same cast `account_erasure_service.py` already uses |
| `contexts.py`, `workspace_service.py`, `account_erasure_service.py` | No | role columns are `Mapped[StrEnum]`; values already constrained by the request schemas' regex. **Cast only** — converting via the enum constructor would add runtime validation, which does not belong in a gate PR. Unifying the schemas onto the enums is a follow-up. |
| `db/redis.py` | No | client is built with `decode_responses=True`; the redis-py stub is not parameterized on that flag |
| `sleep/dedup_merge.py` | No | `(*sorted(...), score)` widened every tuple slot to `UUID \| float`; build the 3-tuple explicitly |
| `llm_pricing_service.py` | No | parameterize the `TTLCache` constructor |
| `memory_service.py`, `secret_store_service.py` (×3) | No | `dict()` over a `Row` sequence |

One note on that last row: I first used `.tuples()`, which is the idiomatic SQLAlchemy 2.0 method —
**it broke three tests** that mock `Result`. Reverted to a `cast` so the *executed call is unchanged*.
A typing-only PR should not alter runtime call patterns.

## Verification

- `pyright src/` — **0 errors** (from 20).
- `ruff check src/ tests/` and `ruff format --check` — clean.
- **2 770 passed / 374 skipped**, no pre-existing test weakened. The one test I changed
  (`test_list_models_returns_models`) was asserting a false contract; it is now stricter, not looser.

## Out of scope

Tier C (`reportArgumentType`, ~163 residuals) and Tier D stay disabled in `[tool.pyright]`. This PR
gates the tiers already meant to be on; raising the tier is a separate decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,14 +55,16 @@ jobs:
           cache: pip
           cache-dependency-path: backend/pyproject.toml
 
-      # The optional extras are NOT decoration here. google-genai and ollama
-      # live in `llm_providers`; with a lean `.[dev]` install pyright reports
-      # them as unresolved imports and stops analysing those modules — which is
-      # exactly how #1443 (Gemini list_models always returning its hardcoded
-      # fallback) stayed hidden. Install what production imports so the gate
-      # sees what production runs.
+      # EVERY optional extra, not just dev. This is load-bearing, not
+      # thoroughness for its own sake: pyright silently stops analysing a module
+      # whose imports it cannot resolve, so a missing extra turns real findings
+      # into a single "unresolved import" line. google-genai/ollama
+      # (`llm_providers`) hid #1443 — Gemini list_models always returning its
+      # hardcoded fallback — behind exactly that. lancedb/pyarrow (`lite`) were
+      # missed on the first attempt at this job and CI caught it.
+      # If a new extra is added to pyproject.toml, add it here too.
       - name: Install dependencies
-        run: cd backend && pip install -e ".[dev,llm_providers,neural]"
+        run: cd backend && pip install -e ".[dev,llm_providers,neural,lite]"
 
       - name: Type check
         run: cd backend && pyright src/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,42 @@ jobs:
       - name: Models no-Column guard
         run: make lint-models-no-column
 
+  # Static type gate for the backend (#1442). The frontend has run
+  # `tsc --noEmit` since day one; pyright was configured in
+  # backend/pyproject.toml with a tier-by-tier plan (#599, #612) and a
+  # `make type-check` target, but nothing enforced it — so it drifted red.
+  # Two of the findings it had accumulated were real functional bugs that
+  # ruff and the 6k-test suite both missed (#1440), plus a third (#1443).
+  #
+  # This is a HARD gate on purpose. `continue-on-error` would have surfaced
+  # those bugs exactly as well as having no check at all.
+  #
+  # Kept as its own job rather than folded into `lint`: pyright needs the
+  # project's real dependency tree installed, while `lint` only needs ruff
+  # and stays fast.
+  type-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: pip
+          cache-dependency-path: backend/pyproject.toml
+
+      # The optional extras are NOT decoration here. google-genai and ollama
+      # live in `llm_providers`; with a lean `.[dev]` install pyright reports
+      # them as unresolved imports and stops analysing those modules — which is
+      # exactly how #1443 (Gemini list_models always returning its hardcoded
+      # fallback) stayed hidden. Install what production imports so the gate
+      # sees what production runs.
+      - name: Install dependencies
+        run: cd backend && pip install -e ".[dev,llm_providers,neural]"
+
+      - name: Type check
+        run: cd backend && pyright src/
+
   # Shell-script quality gate for the blue-green deploy script. shellcheck
   # catches the `set -e` / quoting bug class behind #986 (a bare
   # `var=$(curl ...)` dying silently); the bats suite pins the three
--- a/backend/src/api/routes/auth.py
+++ b/backend/src/api/routes/auth.py
@@ -30,7 +30,14 @@ from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from auth import oauth_endpoints
+# #1442: spelled as an explicit submodule import rather than
+# ``from auth import oauth_endpoints``. This file is itself named ``auth.py``,
+# so within ``api/routes/`` the bare form is ambiguous between the top-level
+# ``auth`` package and this sibling module — pyright resolves it to the sibling
+# and reports every ``oauth_endpoints.*`` call as an unknown attribute. Runtime
+# always resolved it correctly (absolute import); this makes the intent explicit
+# for readers and analyzers alike.
+import auth.oauth_endpoints as oauth_endpoints
 from auth.dependencies import SessionUser
 from auth.oauth2 import OAuth2Manager
 from auth.password import verify_password
--- a/backend/src/api/routes/contexts.py
+++ b/backend/src/api/routes/contexts.py
@@ -1511,7 +1511,7 @@ async def update_context_member_role(
     # regex; the column is Mapped[ContextRole] (a StrEnum). Cast rather than
     # convert so this stays a pure typing fix with no new runtime validation
     # — unifying the request schemas onto the enums is a separate change (#1442).
-    member.role = cast("ContextRole", body.role)
+    member.role = cast(ContextRole, body.role)
     member.updated_at = func.now()
 
     await db.commit()
--- a/backend/src/api/routes/contexts.py
+++ b/backend/src/api/routes/contexts.py
@@ -11,7 +11,7 @@ Each context maps to a separate Qdrant collection for memory isolation.
 from __future__ import annotations
 
 from datetime import datetime
-from typing import Any
+from typing import Any, cast
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
@@ -1507,7 +1507,11 @@ async def update_context_member_role(
             )
 
     # Update role
-    member.role = body.role
+    # body.role is constrained to owner|editor|viewer by the request schema's
+    # regex; the column is Mapped[ContextRole] (a StrEnum). Cast rather than
+    # convert so this stays a pure typing fix with no new runtime validation
+    # — unifying the request schemas onto the enums is a separate change (#1442).
+    member.role = cast("ContextRole", body.role)
     member.updated_at = func.now()
 
     await db.commit()
--- a/backend/src/db/redis.py
+++ b/backend/src/db/redis.py
@@ -71,7 +71,7 @@ async def get_cache(key: str) -> str | None:
         # get_redis_client() builds the client with decode_responses=True, so
         # values really are str; the redis-py stub is not parameterized on that
         # flag and widens to bytes | str (#1442).
-        return cast("str | None", await client.get(key))
+        return cast(str | None, await client.get(key))
     except Exception as e:
         logger.error("redis_get_failed", key=key, error=str(e))
         return None
--- a/backend/src/db/redis.py
+++ b/backend/src/db/redis.py
@@ -3,6 +3,8 @@
 Singleton pattern for connection pooling.
 """
 
+from typing import cast
+
 import redis.asyncio as aioredis
 
 from config.database import REDIS_URL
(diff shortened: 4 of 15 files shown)


JFK:
CI settled — **12/12 pass**, including the new `type-check` gate.

Worth recording that the gate failed on its own first run, and for exactly the reason this job
exists. The initial install was `.[dev,llm_providers,neural]`; CI reported:

```
db/lance_store.py:62,336 - error: Import "lancedb" could not be resolved (reportMissingImports)
db/lance_store.py:343    - error: Import "pyarrow" could not be resolved (reportMissingImports)
```

Both live in the `lite` extra. My machine had them installed from unrelated work, so the local run
was clean while CI was not — the same masking that hid #1443 behind an unresolved import. Fixed by
installing **every** extra (`dev,llm_providers,neural,lite`), with the reasoning written into the
job comment so a future extra gets added here instead of silently narrowing coverage again.

That is the gate catching a gap in itself on day one, which is a reasonable argument for keeping it
a hard failure.

JFK:
Both review findings addressed in `5dd512a1`.

## 1. `cast()` with a string-literal type — **valid, fixed**

@Copilot and github-code-quality both flagged that `cast("CursorResult[Any]", ...)` leaves the
import unreferenced at runtime. Ruff did not flag it, but the substance is right, and the codebase
already has a settled convention I diverged from — **every** pre-existing cast site uses the concrete
form:

```
api/routes/admin.py:1499        cast(CursorResult[Any], await db.execute(stmt))
repositories/neural_edge.py     cast(CursorResult[Any], ...)   x5
tasks/file_tasks.py:304         cast(CursorResult[Any], ...)
```

The two new sites were the only string-form ones in the repo. Converted **all** casts this branch
added, not just the two flagged, so the branch is internally consistent: `sleep.py`,
`sleep/undo.py`, `contexts.py`, `workspace_service.py`, `redis.py`, `secret_store_service.py` (×2),
`memory_service.py`.

## 2. `_AsyncPagerStub.__aiter__` — **stated mechanism is incorrect, but changed anyway**

The claim is that `async def __aiter__` returns a coroutine and `async for` will raise `TypeError`.
That is not what happens: an `async def` containing `yield` is an async **generator** function, so
calling it returns an `async_generator`, which is already an async iterator.

```
__aiter__ is async-generator function: True
__aiter__ is coroutine function     : False
calling it returns                  : async_generator
result has __anext__                : True
async-for result                    : [1, 2, 3]
```

The test also already passed on Python 3.11 in this PR's `backend-unit` run, which is the version the
comment names.

**That said, the finding points at something real.** The actual
`google.genai.pagers.AsyncPager.__aiter__` is a plain `def` that resets the cursor and returns
`self`, with the awaiting in `__anext__`:

```python
def __aiter__(self) -> AsyncIterator[T]:
    self._idx = 0
    return self
```

This stub exists *specifically* to pin the SDK's contract — that is the entire argument behind the
#1443 fix — so paraphrasing that contract as an async generator was sloppy. Rewritten to model the
real protocol (`__aiter__` returns `self`, `__anext__` walks the items, no `.models`).

Re-verified it is still load-bearing — restoring `response.models` fails it for the right reason:

```
gemini_list_models_failed  error="'_AsyncPagerStub' object has no attribute 'models'"
AssertionError: expected the live model list; the hardcoded fallback means the pager was not iterated correctly
```

`pyright` 0 errors · `ruff` clean · 441 passed / 14 skipped.

JFK:
Rebased onto `main` now that #1441, #1438 and #1446 have landed.

Dropped the duplicated #1441 commit (`git rebase --onto origin/main 0a7e7211`) so this PR contains
only its own three commits.

**One conflict, resolved by keeping both sides.** #1438 and this branch each edited the same import
line in `services/memory_service.py`:

```
#1438  -from typing import Any   +from typing import Any, NamedTuple
#1444  -from typing import Any   +from typing import Any, cast
```

Resolved to `from typing import Any, NamedTuple, cast`. That was the only conflict — #1438's changes
are confined to `recall()` and its new helpers, while this branch's `memory_service.py` change is the
`shadow_map` cast inside `_apply_supersede_shadowing`, which #1438 did not touch.

**Re-verified locally, because the combined #1438 + #1444 state had never been exercised together:**

- `pyright src/` — **0 errors** (this is the state the new gate will enforce on `main`)
- `ruff check` + `ruff format --check` — clean
- **2 779 passed / 374 skipped** across the unit suites, including the new
  `test_stdlib_logger_kwargs_guard.py` from #1446

Issue #1456, opened 2026-07-28
refactor(api): remaining complexity outliers after #1439


JFK:
Continues the #1437 → #1439 complexity audit. #1439 addressed its own suggested item 1
(`patch_memory` + `_update_in_place`) and recorded the "confirmed, not refactored" verdict on
`NeuralMemoryConfig.__post_init__`. This issue carries the rest so the measurement is not lost.

Metric unchanged: branch-node count (`If` / `For` / `While` / `Try` / `BoolOp` / `IfExp` /
`ExceptHandler`) per function, AST pass over `backend/src`. Numbers below re-measured on `main`
after #1439.

| Function | Lines | Complexity |
|---|---:|---:|
| `api/routes/memory.py:list_memories` | 354 | 59 |
| `mcp_server/tools/sleep.py:handle_rollback_sleep_run` | 347 | 58 |
| `mcp_server/transport.py:mcp_asgi_app` | 498 | 51 |
| `services/sleep/dedup_merge.py:DedupMergePhase.execute` | 395 | 36 |
| `services/sleep/consolidation.py:ConsolidationPhase.execute` | 230 | 35 |
| `mcp_server/tools/context.py:handle_update_context` | 190 | 32 |
| `mcp_server/tools/resource.py:handle_setup_resource` | 284 | 31 |
| `api/routes/admin.py:list_users` | 271 | 30 |

For scale: `MemoryService.recall` went 828/80 → 271/17 in #1437, `patch_memory` 351/55 → 120/8
and `_update_in_place` 183/38 → 77/9 in #1439.

## Suggested order

1. `list_memories` — the REST sibling of `recall`; check first whether filter construction can
   reuse `recall`'s filter block rather than duplicating it.
2. `handle_rollback_sleep_run` — one long per-action dispatch loop; each action type is a
   natural helper. Note #1450 just added a branch here, so re-read before splitting.
3. `mcp_asgi_app`, then the sleep phases and the MCP tool handlers.

## Settled, do not re-open

- `NeuralMemoryConfig.__post_init__` (49) — a flat sequence of independent `if not X: raise`
  validations, near-zero nesting. The decision is recorded in its docstring as of #1439.
- `mcp_server/tools/_definitions.py:get_tool_definitions` — 2 911 lines, complexity 3. A literal
  schema table; excluded from the audit by construction.

## Also noted, needs its own design discussion first

`services/memory_service.py` is still ~5 500 lines and would benefit from a module split
[...]

Issue #1437, opened 2026-07-25
refactor(api): decompose MemoryService.recall (828 lines / complexity 80)


JFK:
## Problem

`MemoryService.recall()` (`backend/src/services/memory_service.py:2499`) is **828 lines** with a
**cyclomatic complexity of 80** — an extreme outlier in the backend. Repo-wide AST measurement over
`backend/src` (320 files, 123 854 LOC, 2 101 functions):

| Rank | Function | Lines | Complexity |
|---|---|---:|---:|
| 1 | `MemoryService.recall` | **828** | **80** |
| 2 | `api/routes/memory.py:list_memories` | 354 | 59 |
| 3 | `mcp_server/tools/sleep.py:handle_rollback_sleep_run` | 312 | 57 |
| 4 | `MemoryService.patch_memory` | 351 | 55 |
| 5 | `mcp_server/transport.py:mcp_asgi_app` | 498 | 51 |

(`mcp_server/tools/_definitions.py:get_tool_definitions` is longer at 2 911 lines but is a flat
literal schema table — complexity 3 — and is *not* in scope.)

`recall()` is also the single hottest and most security-sensitive read path in the product: it
carries the #1291 agent-binding gate, the #708 shared-context BYOK gate, the #887 trust-tier filter,
the #1299 per-row binding filter, #1208 supersede shadowing and the #1306 selection-evidence seam.
Every one of those gates is inlined into one function body, which makes each subsequent change to
it a whole-function re-read.

## Concrete defects in the current body

1. **Duplicated empty-response construction.** The 18-line `RecallResponse(results=[], …,
   selection_evidence=(await self._build_selection_evidence(...))[1] if … else None)` block is
   written out **twice, character-identical** — at `:2705` (empty analysis cluster) and `:2809`
   (no search hits). Any future field added to the empty-recall contract has to be added in two
   places or the two exits silently diverge.
2. **Inconsistent dict access in the confidence filter** (`:3281`). The guard uses the safe
   `r.get("hybrid_score", r.get("score"))` but the value expression uses
   `r.get("hybrid_score", r["score"])`. It is *currently* safe only because the comprehension's
   `if` short-circuits the `KeyError` case; the two expressions must be kept in lockstep by hand.
3. **Function-body imports.** `import os` at `:2548` plus ~10 further deferred
   `from … import …` statements scattered through the body. Some are deliberate (circular-import
   breaks); several are not, and there is no way to tell them apart at a glance.
[...]

PR #1457, opened 2026-07-28
refactor(api): decompose patch_memory and _update_in_place (#1439)

### LINKED ISSUE #1439: refactor(api): remaining complexity outliers after #1437 (audit record)
Follow-up to #1437, which decomposed `MemoryService.recall` (828 lines / complexity 80).
This issue records the rest of the 2026-07-25 complexity audit so the measurement is not
lost, and tracks the remaining outliers.

## Method

AST pass over `backend/src` (320 files, 123 854 LOC, 2 101 functions). Complexity is a
branch-node count (`If` / `For` / `While` / `Try` / `BoolOp` / `IfExp` / `ExceptHandler`)
per function — same metric used in #1437, so the numbers are comparable.

## Remaining outliers, after #1437

| Function | Lines | Complexity |
|---|---:|---:|
| `api/routes/memory.py:list_memories` | 354 | 59 |
| `mcp_server/tools/sleep.py:handle_rollback_sleep_run` | 312 | 57 |
| `services/memory_service.py:MemoryService.patch_memory` | 351 | 55 |
| `mcp_server/transport.py:mcp_asgi_app` | 498 | 51 |
| `neural/config.py:NeuralMemoryConfig.__post_init__` | 221 | 49 |
| `services/memory_service.py:MemoryService._update_in_place` | 183 | 38 |
| `services/sleep/dedup_merge.py:DedupMergePhase.execute` | 395 | 36 |
| `services/sleep/consolidation.py:ConsolidationPhase.execute` | 230 | 35 |

For scale: `MemoryService.recall` is now **271 lines / complexity 17** and no longer
appears in this list.

Two entries are arguably fine as they are and should be confirmed-then-closed rather than
refactored on principle:

- `NeuralMemoryConfig.__post_init__` (49) is a flat sequence of independent
  `if not X: raise ValueError(...)` validations. High branch count, near-zero nesting —
  splitting it would spread one contract across several methods for no reading benefit.
- `mcp_server/tools/_definitions.py:get_tool_definitions` is 2 911 lines but complexity
  **3** — a literal schema table. Deliberately excluded from the audit.

## Suggested order

1. `patch_memory` + `_update_in_place` — same file and same class as #1437, so the phase-helper
   pattern established there transfers directly, and the two share the update path.
2. `list_memories` — the REST sibling of `recall`; check first whether filter construction
   can reuse the `_recall_hydrate_memories` filter block rather than duplicating it.
3. `handle_rollback_sleep_run`, `mcp_asgi_app` — independent, can be picked up any time.

Also noted but out of scope for a function-level sweep: `services/memory_service.py` is
5 355 lines and would benefit from a module split (`recall` / `write` / `graph-seed` /
`embedding` faces). Larger blast radius — worth its own design discussion before anyone
starts.

## Not a bug list

The static defect hunt run alongside #1437 came up clean: ruff (E/W/F/I/B/C4/UP/DTZ) passes;
no f-string SQL reaching user input; no naive-datetime writes; no unreferenced
`asyncio.create_task`; every division-by-zero candidate guarded. These are maintainability
items, not correctness ones — they should be scheduled as capacity allows, not treated as
a defect backlog.

JFK:
Closes #1439

## What

The #1437 audit's own suggested item 1: the two top outliers, both in
`MemoryService`, both on the update path.

| Function | Before | After |
|---|---|---|
| `patch_memory` | 351 lines / cx 55 | **120 / 8** |
| `_update_in_place` | 183 lines / cx 38 | **77 / 9** |

Extracted per path: `_load_authorized`, `_guard_size`, `_apply_fields`,
`_sync_qdrant_payload`, plus (PATCH only) `_patch_invalidate_edges` and
`_patch_build_response`. Largest helper is cx 14.

## Behaviour-preserving

Blocks moved whole; nothing reordered, no logic edited. Three guards were
rewritten from `if cond: body` to `if not cond: return` + body — those are the
only control-flow edits, and they were the review's first target.

**The ordering that matters** is in `_patch_load_authorized`: permission is
checked *before* the tombstone check, so a non-member cannot tell a soft-deleted
memory (would-be 410) from a UUID that never existed (404) — otherwise anyone
who guesses or harvests a UUID can confirm "this memory was once real". That
reasoning now sits in the helper's own docstring instead of a comment two
hundred lines away from the code it constrains.

## The two paths look alike and are not

Splitting them puts three deliberate differences where a reader will find them,
rather than leaving them to be rediscovered by diffing two long functions:

| | `patch_memory` (REST #439) | `_update_in_place` (MCP) |
|---|---|---|
| soft-deleted target | **410** `MemoryGoneError` | uniform **404** |
| omitted vs explicit null | `model_fields_set` + explicit `is None` | `None` means "leave alone"; truthy `or` is equivalent |
| Qdrant failure | **swallowed** — PG already committed | **raises** — commit comes after |

That last row is why the two `_sync_qdrant_payload` helpers are not merged: they
sit on opposite sides of their commit.

## Also

Records the audit's own "confirmed, not refactored" verdict on
`NeuralMemoryConfig.__post_init__` (cx 49) in its docstring — it is a flat
sequence of independent `if not X: raise` validations, and the metric counts
branches, so a validation table is nothing but branches. Splitting it would
scatter one contract across several methods. Re-measuring will surface it again;
the note is there so the answer does not have to be re-derived.

## Scope

#1439 lists eight outliers and an explicit suggested order. This PR does item 1
and settles the two "confirm-then-close" entries. The remaining six are carried
forward with re-measured numbers in **#1456** (v0.62.0) — including the note that
`memory_service.py` (~5 500 lines) wants a module split, which needs its own
design discussion before anyone starts.

## Verification

- `tests/services/` + `tests/repositories/`: **1831 passed**, 343 skipped
- `tests/neural/` + `tests/services/test_memory_patch.py`: 223 passed
- `ruff check` clean, `ruff format --check` clean, `pyright src/services/memory_service.py` clean

## Reviewed

One adversarial round (codex, read-only) → **green, no findings**, with all seven
attack categories confirmed explicitly: the three guard inversions, the
commit/Qdrant ordering on both paths (including the exception path), the
permission-before-tombstone order and `include_deleted=True` staying only on the
PATCH loader, the `effective_details` return value in every branch, static-method
scope and imports, and the `__post_init__` claim checked against the actual code.

### DIFF
--- a/backend/src/neural/config.py
+++ b/backend/src/neural/config.py
@@ -277,7 +277,18 @@ class NeuralMemoryConfig:
     sleep_importance_reeval_enabled: bool = True  # Phase 3 on/off
 
     def __post_init__(self) -> None:
-        """Validate configuration parameters."""
+        """Validate configuration parameters.
+
+        #1439 audit, confirmed not-a-refactor: this scores 49 on the branch-node
+        complexity metric, high enough to appear alongside functions that were
+        decomposed. It is deliberately left whole. The count comes from a flat
+        sequence of independent ``if not X: raise ValueError(...)`` guards with
+        near-zero nesting — the metric counts branches, and a validation table
+        is nothing but branches. Splitting it would scatter one contract across
+        several methods and make "what does a valid config look like" harder to
+        read, not easier. Re-measuring will surface it again; this note is here
+        so the answer does not have to be re-derived each time.
+        """
         # #1160 back-compat: 'ollama' was renamed to 'self_hosted' in a clean
         # cutover. A stale SLEEP_LLM_PROVIDER=ollama env (documented as valid in
         # v0.41) or an un-migrated neural_config row would otherwise fail EVERY
--- a/backend/src/services/memory_service.py
+++ b/backend/src/services/memory_service.py
@@ -663,36 +663,9 @@ class MemoryService:
         current_workspace_id: UUID | None = None,
     ) -> UpdateMemoryResponse:
         """In-place update by memory_id."""
-        from config.constants import MAX_CONTENT_SIZE
-        from services.permission_service import PermissionService
         from utils.text import normalize_for_search
 
-        # #1316: repo.get() excludes soft-deleted rows by default, so a
-        # tombstoned memory is uniformly "not found" here.
-        memory = await self.memory_repo.get(request.memory_id)
-        if not memory:
-            raise NotFoundException("Memory", str(request.memory_id))
-
-        # Permission check
-        from services.permission_service import CallerId, MemoryAuthorId
-
-        perm_service = PermissionService(self.db)
-        can_access = await perm_service.can_access_memory(
-            user_id=CallerId(user_id),
-            memory_user_id=MemoryAuthorId(memory.user_id),
-            workspace_id=memory.workspace_id,
-            context_id=memory.context_id,
-            access="write",  # #1275: WRITE path — can_read-only binding must not permit mutation
-            operation="update",  # #1286 (P0-5): deny-capture audit identity
-            memory_id=request.memory_id,
-            # #1301: id-addressed update of a read-denied row must be a
-            # uniform 404 (reference/forget doctrine) — REST PATCH and this
-            # MCP-reachable path stay in parity (#1291/#1292).
-            memory_type=memory.type,
-            memory_source_type=memory.source_type,
-        )
-        if not can_access:
-            raise NotFoundException("Memory", str(request.memory_id))
+        memory = await self._update_load_authorized(request.memory_id, user_id)
 
         # Pre-compute normalized values (avoid double normalization)
         normalized_summary = (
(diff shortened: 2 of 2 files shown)


JFK:
Copilot's suppressed low-confidence comment on `_update_guard_size` was correct — thank you for surfacing it anyway.

Verified: `UpdateMemoryRequest.context_summary` has no `min_length` and `details` is a free dict, so `""` and `{}` are valid values that `_update_apply_fields` **does** write to the row. Being falsy, the `or` chain falls back to the stored value and sizes the row as if the field were untouched — so clearing a large `context_summary` can be rejected for being too big while actually shrinking the row. (`summary`/`content` are safe: their schema `min_length` rejects `""`.)

What changed here: **only my docstring**, which claimed the `or` form was equivalent to the PATCH sibling's `is None` checks. It is not, and that claim is now corrected in 8f6cdd21.

What did **not** change: the behaviour. This PR's whole claim is that it is a behaviour-preserving decomposition, and editing the predicate would invalidate that. The quirk is pre-existing, fails in the safe direction (spurious rejection, never an oversized row), and is tracked in #1458 with the clearing-case regression test as an acceptance criterion. The docstring now carries a "do not tidy this into `is None` without that test" note.

PR #1484, opened 2026-08-03
docs: remove the remaining private-repo references and links (#1469 follow-up)

JFK:
#1469 scrubbed one spelling of the consumer's repository name. **A second
spelling was used elsewhere in the tree and survived** — that scrub grepped the
first one literally, so it never matched.

21 occurrences across 13 files: backend docstrings, two public contract
documents, an RFC, and infra comments. **Four are full GitHub URLs into the
private repository**, one of them a direct issue link.

## Approach

Option B of three considered:

| | |
|---|---|
| A | backend source only — leaves the URLs, which are the actual trail |
| **B (this)** | sever names + URLs everywhere; keep every document's substance |
| C | rewrite the documents' content too, including the RFC |

The two contract documents exist **to specify the integration**, so
genericizing their content would defeat their purpose. Only the identifying
references change — the specifications, tables and rules are untouched.

The RFC keeps its historical record; only the consumer's name becomes its role.

## Deliberately not changed

`BRIDGE_WORKER_CONTAINER`'s default still carries the container name. That is a
deployment fact rather than a repository reference, and removing it would break
the color flip on any host whose env has not been updated first — the same
ordering hazard #1483 called out. Doing it safely means updating the host env in
the same change; worth doing, not worth risking a silent flip regression for.

## Verification

`git ls-files | grep` for either spelling and for `github.com/<org>/<private>`
now returns nothing. No executable line changed; backend tests for the touched
modules pass (34); `ruff check` clean.

## Note on recurrence

This is the second scrub. There is **no CI gate**, and the references keep
coming back for a structural reason: a comment explaining *why* a guard exists
naturally names the system the guard protects against. A `grep` gate in CI would
end the class — happy to add one if wanted.

https://claude.ai/code/session_01HZpvFKGhBZrEQ4jyFRpTwJ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -109,7 +109,7 @@ AUDIT_PSEUDO_SALT=kagura-erasure-v1
 AUDIT_HMAC_KEY=change-me-audit-hmac-key
 
 # -----------------------------------------------------------------------------
-# ai-worker connectors (kagura-chat-bridge, Spec 2026-06-02)
+# ai-worker connectors (Spec 2026-06-02)
 # -----------------------------------------------------------------------------
 
 # Shared bearer token the ai-worker presents to GET /api/v1/workers/config.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,28 @@ jobs:
       - name: Models no-Column guard
         run: make lint-models-no-column
 
+      # This repository is PUBLIC; the connector worker that consumes its
+      # workers lane is not. Scrubbing its name has now been needed twice
+      # (#1469, #1484) and it keeps coming back for a structural reason: a
+      # comment explaining WHY a guard exists naturally names the system the
+      # guard protects against, and reviewers read that as good context.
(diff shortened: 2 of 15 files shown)


JFK:
Added a CI gate to this PR, because a second scrub in two days means memory is
not the right mechanism.

**It found a miss immediately.** `deploy_bridge_restart.bats` still carried the
repository name, the incident call count and the outage dates — in a file added
the same day the previous scrub landed. Fixed here, so the gate paid for itself
before running in CI once.

```
- name: No private-repo references
  run: |
    if git grep -nIE 'kagura-(chat-)?bridge' -- ':!*BRIDGE_WORKER_CONTAINER*' \
       | grep -vE 'kagura-bridge-worker[-0-9]*' ; then
      exit 1
    fi
```

Container names are exempt by design — `BRIDGE_WORKER_CONTAINER` has to address
a real container for the flip to restart it, and a deployment fact is not a
repository reference.

Issue #1487, opened 2026-08-04
fix(frontend): context creation is blocked by a UI-only key gate, with no reason shown


JFK:
## Symptom

A **Pro** workspace with 3 existing contexts cannot create another one: the
"New Context" button is disabled and **nothing on screen says why**. The natural
reading is "my plan only allows one context", which is what was reported.

Neither the plan nor the quota is involved.

## Root cause: a UI-only gate with no backend counterpart

`frontend/src/app/(authenticated)/workspace/contexts/page.tsx:479`

```tsx
disabled={hasOpenAIKey === false || isQuotaReached}
```

`isQuotaReached` is structurally `false` for a Pro workspace (see the second bug
below), so the disable comes entirely from `hasOpenAIKey === false`.

**Context creation has no server-side key precondition.** `contexts.py:474-553`
and `context_service.py:196-250` create the Context row, the
`ContextSearchConfig` and the Qdrant collection with zero embedding calls and
zero key lookups. Creation succeeds without a key on every tier.

The probe was never meant to be an entitlement gate. Both the API client
(`frontend/src/lib/api/workspaces.ts:551-561`) and the route
(`backend/src/api/routes/workspaces.py:1165,1185`) attribute it to **#181,
"OpenAI API key guidance in context creation" / "Prevent context creation
errors"** — guidance, so users would not later hit a confusing embedding
failure. The state variable is mislabelled `// Issue #165: API key check`.

`enable_byok` defaults to `True` (`settings.py:510-511`), so this is the default
posture, not a niche deployment.

### Why it is silent

The explanatory amber alert and the "Configure API Key" CTA live at
`page.tsx:861-919`, **inside the `contexts.length === 0` branch only**.
`hasOpenAIKey` appears nowhere in the non-empty branch. So:

| workspace state | what the user sees |
|---|---|
| 0 contexts, no key | alert + CTA explaining what to do |
| ≥1 context, no key | **a dead button and no explanation** |

The reported workspace has 3 contexts, so it lands in the second row.

### The gate is also empirically wrong

The affected workspace holds **676 memories across its 3 contexts** and
`recall`/`remember` work normally against it with no key row. Everything the key
would be needed for already works; only the button is dead.

## Second bug: the client re-implements the quota, wrongly

`page.tsx:129-130`

```tsx
const isQuotaReached =
  currentWorkspace?.plan_name === "free" && contexts.length >= 1;
```

The literal `1` and the free-only branch are wrong in four ways:

[...]

PR #1493, opened 2026-08-04
feat(auth): sign-out semantics for multi-account, and 3 review fixes (#1488 Phase 4)

### LINKED ISSUE #1488: feat(auth): multi-account switching from the sidebar
Switch between **accounts** from a menu at the bottom-left of the sidebar, the
way Google/Slack do it.

## Short answer: yes, but it is not a UI addition

The sidebar menu is the easy part. Multi-account fights **two invariants that
were chosen on purpose**, and a third thing that is not session state at all.

### 1. One session cookie per browser

All four login flows funnel through three `set_cookie` calls, and every one of
them writes the same key at the same path:

- `auth.py:874-882` (Google), `:1325-1333` (GitHub), `:1441-1449`
  (`_set_session_cookie`, used by password login `:1605` and MFA `:1654`)
- all `key="kagura_session"`, `path="/"`, no `domain`, no suffix or index

A browser stores exactly one value per (host, path, name), so signing in as B
destroys the browser's handle on A. The read side matches: one
`request.cookies.get("kagura_session")` (`middleware/session.py:107`) setting one
`request.state.user` (`:115`).

### 2. One session per user, enforced server-side

Every login calls `delete_user_sessions` before minting a new one
(`auth.py:809`, `:1281`, `:1402`), filtered by user_id
(`auth/session.py:340-342`). This is deliberate — **#114**. So even if the cookie
problem were solved, the server actively destroys the other session.

### 3. The active workspace is a column on `users`, not session state

This is the one that is easy to miss. `get_current_user` returns
`request.state.user` verbatim (`auth/dependencies.py:126`) and that dict never
contains `current_workspace_id`. It is grafted on later by `require_session_auth`
(`:482-484`) and `get_user_from_api_key_or_session` (`:606-608`), each doing a
fresh `select(User.current_workspace_id)` (`:84`).

So "which workspace am I in" is **global per user across every device, tab and
session**. Two tabs of the *same* account already cannot sit in different
workspaces. Any multi-account fix has to make account *and* workspace
request-scoped, or switching accounts in one tab will yank the other.

## Product tension worth naming

Account **linking** already exists (#517) and does the architectural opposite:
it collapses several identities into ONE user. Before building multi-account, it
is worth being explicit that we want both — "one human, several identities,
merged" *and* "one browser, several accounts, kept apart" are different answers
to overlapping problems.

Also note the top-left switcher already covers *workspaces*, and one user can
already belong to many. If the real need is "get to my other workspace faster",
that exists today and this issue is not the answer.

## Proposed design: container session

Two shapes were considered.

**Rejected — one cookie per account** (`kagura_session_<slot>` plus a pointer
cookie). It multiplies the cookie surface, needs a CSRF story for the pointer,
and makes `delete_user_sessions` semantics much harder to reason about.

**Chosen — one cookie, many accounts inside the server-side record.** The cookie
stays exactly as it is; the Redis session record becomes:

```
{ accounts: { <user_id>: {identity...} }, active: <user_id> }
```

- Adding an account = a login that ADDS to the record instead of replacing it
- Switching = mutating `active` server-side; no cookie semantics change at all
- #114's "one session per user" is preserved *per account*
- No new CSRF surface, because nothing new is client-writable

`current_workspace_id` moves into the per-account entry in that record, with the
`users` column kept as the persisted default for a fresh login.

## Phases

1. **Session container** — reshape the record, keep exactly one account in it.
   Pure refactor, no visible change. `require_session_auth` and
   `get_user_from_api_key_or_session` read the active account's workspace from
   the session, falling back to the column.
2. **Add / switch** — `POST /auth/accounts/switch`, and a login path that adds
   rather than replaces. Server-side only; still no UI.
3. **Sidebar menu** — bottom-left account switcher. Model it on the existing
   bottom user menu, not the top-left `WorkspaceSwitcher` (which is a hand-rolled
   non-standard control).
4. **Sign-out semantics** — sign out *this account* vs *all accounts*, and clear
   the client caches that sign-out currently leaves behind.

## Out of scope

- Account linking (#517) — the opposite direction, already shipped
- API-key auth, which is per-workspace and unaffected

## Acceptance criteria

- [ ] Two accounts can be signed in at once in one browser; neither evicts the other
- [ ] Switching accounts does not change what any other tab of the other account sees
- [ ] Two tabs of the *same* account can sit in different workspaces (this is
      currently impossible and is a prerequisite, not a bonus)
- [ ] Signing out of one account leaves the other signed in
- [ ] `delete_user_sessions`' one-session-per-user guarantee still holds per account

JFK:
Closes #1488.

Phase 4 completes the multi-account work, and repairs three defects an
adversarial review of the whole v0.64.0 surface turned up — two of them in
code already merged and running in production.

## Phase 4 — what "sign out" means now

`POST /auth/logout?scope=current|all`, wiring up `remove_account`: a Phase 2
primitive that no route had ever called.

- **The default stays `all`.** An older frontend served during a blue/green
  rollout posts to this route bare; if the default ever became `current` those
  callers would silently start leaving accounts signed in.
- **The cookie is cleared only when the session actually ended.** Clearing it
  while other accounts remain would log the user out anyway and orphan their
  records in Redis — which, from the browser, looks like the feature simply not
  working.
- **The route stays unauthenticated.** Adding an auth dependency here reads as
  hardening and would instead 401 exactly the users who most need to log out:
  those whose session already expired.
- **A refused removal degrades to a full sign-out.** Reporting failure would
  leave the user signed in as the account they just asked to leave; the wide
  action always satisfies the request.

On the client, `clearIdentityScopedClientState()` drops what the cookie cannot:
keys written with no user id in them. `kagura_last_workspace_id` is the one that
shows — sign out as A, back in as B, and B's workspace picker opens on A's
workspace. `theme`, `kagura_locale` and the sidebar layout are deliberately
**kept**: they are device preferences, and the common multi-account case is one
person with a work and a personal account.

## Review fixes

**Critical — adding an account deleted the session it was appending to.**
`delete_user_sessions` ran before the add-account intent was read. It matches by
MEMBERSHIP and deletes whole containers, so re-authenticating an identity the
container already held destroyed the session the flow exists to append to,
taking every other signed-in account with it, then failed. `prompt=consent`
means a user with one account at the IdP hit this *every* time they clicked
"Add another account". Fixed by reading the intent first and passing
`exclude_session_id`; #114 still invalidates every OTHER session for that
identity.

**Critical — the user menu closed itself.** `SidebarContent` was a component
defined inside `Sidebar`, so every render produced a new component type and
React remounted the whole subtree, discarding the Radix dropdown's uncontrolled
open state. Latent before this milestone; the switcher made it fire on every
open, ~200ms in, before the account rows were ever visible. **The Phase 3
switcher was effectively unreachable in production.**

**Major — "Add another account" assumed Google.** On a deployment without Google
configured (the OSS default is password login with no OAuth at all) the item
navigated the whole tab onto a raw 500 JSON page. Now reads `/auth/config` and
fails closed.

**Major — the quota banner told paid workspaces the free-plan rule.** #1487
widened the gate to "any plan at the server-sent cap" but left the banner
asserting "Free plan allows 1 context. Upgrade to Basic or Pro". A Pro
workspace at 20/20 was told all three of those, falsely.

Plus: the empty-state Create bypassed the quota gate its siblings honour, and
`contextLimitReached` read "per workspaceanization".

## Verification

Every fix is pinned by a test that was **verified by mutation** — the guard is
reintroduced, the test is confirmed to fail, the fix is restored. That includes
the two criticals: reverting `SidebarContent` to an inline component fails the
remount test, and dropping `exclude_session_id` fails four container tests.

The add-account ordering pin is measured on the **AST**, not the source text. A
first version searched the text and failed against *correct* code, because the
earliest textual occurrence of `delete_user_sessions` is now inside the comment
explaining the fix. An assertion a prose edit can break is not measuring code.

The create-gate suite's next-intl mock dropped interpolation vars, so it could
only see *that* a banner appeared, never that it said something false. It now
echoes them.

| Gate | Result |
|---|---|
| backend pytest (auth surface) | 426 passed |
| backend ruff | clean |
| backend pyright | 0 errors |
| frontend tsc | 0 errors |
| frontend vitest | no new failures |

Two pre-existing local failures (`useSystemFeatures`, `CostDashboard`) were
confirmed against a stashed baseline; the Windows-only `next build` segfault was
confirmed against a clean `origin/main` worktree. None are caused by this branch.

## Not covered

The two-account flow has **not** been exercised in a real browser — signing a
second account in requires entering credentials, which I cannot do. Everything
reachable without that is under test. Given that one of the criticals above made
the switcher unusable while its unit tests were green, this is worth a manual
pass before relying on it.


JFK:
Thanks — took both wording notes in a87726e2.

Pushing back on the CodeQL one (`Superfluous trailing arguments` on
`new StorageEvent("storage", {...})` in Sidebar.test.tsx), because I think the
alert is wrong rather than the code.

`StorageEvent` has taken a `StorageEventInit` dictionary as its second
constructor argument since the DOM standard replaced the legacy
`initStorageEvent()` form. The argument is not superfluous — it is the only way
to set `key` and `newValue`, which is exactly what the listener under test
reads. CodeQL is matching against a stale extern in its bundled
`webstorage.js`, which models the interface without the constructor init dict.

The suggested workaround (`new Event("storage")` plus two `Object.defineProperty`
calls and a cast) would make the test three lines longer, less faithful to what
a browser actually dispatches, and would need a comment explaining why it is
written the odd way — all to satisfy a stale type model. The `Analyze
(javascript-typescript)` job passes, so nothing is gated on it.

Happy to change it if the alert turns out to gate something I have missed, or if
the repo would rather keep the security dashboard empty regardless — just say
which and I will apply it.

Issue #1497, opened 2026-08-04
fix(api): merge_contexts silently drops memories whose embedding failed, then deletes the source holding them


JFK:
Found while designing #1496. Filed separately because that issue is about
*visibility* and this one is about *losing the data*.

## What

`ContextService.merge_contexts` copies only rows whose embedding succeeded
(`context_service.py:896-903`):

```python
select(Memory).where(
    Memory.context_id == source_context_id,
    Memory.deleted_at.is_(None),
    Memory.embedding_status == "success",     # <-- everything else is skipped
)
```

and then, with `delete_source=True`, soft-deletes the whole source context
(`:975-976` → `delete_context`, which sets `deleted_at`, `:1002`).

So every memory in the source at `pending`, `processing` or `failed` is **not
copied and not kept**. It stays parented to a context the user just removed.

## Why this is worse than #1496

#1496 is "467 memories are not searchable and nobody said so" — the rows are
intact and recoverable. This one removes them from the user's reach in a single
click, and the same deployment that produced #1496 is precisely the situation
where it bites: a workspace where 61% of one context's memories sit at `failed`
would merge that context and silently keep 39% of it.

The `merged_count` returned to the caller is the number copied, so the UI
reports a smaller number than the `memory_count` it displayed a moment earlier,
with no explanation. A user reading "652 memories" then seeing "merged 234" has
no way to know the other 418 were left behind rather than deduplicated.

Soft-delete means the rows are technically still in the table — but nothing in
the product surfaces or restores them, so from the user's side this is loss.

## Suspected shape of the fix

Decide deliberately; do not just widen the filter without thinking:

- **Copy everything and carry the status.** Copying a `failed` row preserves it
  and lets the sweep re-embed it in its new home. This is probably right, and
  is consistent with #1496's position that a failed embedding is a recoverable
  state, not a dead row.
- If some statuses genuinely should not be copied, then `delete_source` must
  refuse rather than delete: leaving the source in place is the only safe
  outcome when the merge is known to be partial.
- Either way `merged_count` should not silently disagree with what the user was
  shown. Report skipped rows if any are skipped.

## Acceptance criteria

- [ ] No memory becomes unreachable as a result of a merge. Pin with a test that
      merges a source containing `success` + `failed` + `pending` rows and
      asserts every one is reachable afterwards.
- [ ] `delete_source=True` cannot remove a context whose contents were not fully
      transferred.
- [ ] The returned count reconciles with what the caller can see, or the
[...]

PR #1498, opened 2026-08-04
fix(api): stop configuration failures stranding memories, and surface the ones already lost (#1496)

### LINKED ISSUE #1496: fix(api): memories that exhaust the embedding retry budget vanish from search with no signal to anyone
## What happened

While investigating an unrelated report on production I queried
`embedding_status` and found **467 memories** permanently at `failed`:

```
status=failed  n=467      -- every one at embedding_retry_count = 3
status=success n=7567
```

`MAX_EMBEDDING_RETRIES = 3` (`config/constants.py:174`), and the sweep's
eligibility clause requires `embedding_retry_count < MAX_EMBEDDING_RETRIES`
(`memory_service.py:embedding_retry_eligible_clause`). **All 467 have exhausted
the budget, so the 30-second sweep will never pick them up again.** They are
stuck permanently.

Because a failed embedding never reaches Qdrant (`memory_service.py:5572`), and
BM25 sparse vectors also live in Qdrant (`db/qdrant.py:697`), these memories are
invisible to `recall` in **both** semantic and keyword mode. They are saved,
they are counted in the usage meter, and they cannot be found.

One affected workspace had 418 failed against 269 successful — **61% of that
workspace's memories are unsearchable.** The owner had no way to know.

## Why nobody noticed

- `remember()` returns success: the row is committed with
  `embedding_status='pending'` and the embed is fired as a background task
  (`memory_service.py:561, 596`). The caller sees a write that worked.
- The dashboard's memory count and the plan usage meter count the rows, not
  their searchability. The affected workspace displays "660 / 100000 memories"
  with no qualification.
- `embedding_status` is returned by `GET /contexts/{id}/memories`
  (`contexts.py:1622, 1710`) but **no frontend code reads it** — `grep` for
  `embedding_status` across `frontend/src` returns nothing.
- There is no alert, no admin dashboard panel, and no log-based signal that
  survives past the individual failure.

The only way to discover this is to query the database directly.

## Not a duplicate of the recovery tooling

`POST /api/v1/admin/embedding/retry-failed` (#93) already exists and does the
right thing — it resets `embedding_status='pending'`, clears the error, and
zeroes `embedding_retry_count` so the sweep can claim them again. That is the
cure. This issue is about the **absence of a diagnosis**: an operator has to
already suspect the problem to go looking for the button.

## Acceptance criteria

- [ ] A workspace owner can see that some of their memories are not searchable —
      at minimum a count, ideally per context. It should be visible where the
      memory count is shown, not buried.
- [ ] Something surfaces the condition to operators without a manual query: an
      admin panel figure, a metric, or a log line emitted on transition to
      permanently-failed (not just per-attempt).
- [ ] The copy distinguishes "saved but not searchable" from "not saved" — the
      row is safe, which matters for user trust.
- [ ] A memory that exhausts its retry budget records *why*. `embedding_error`
      exists on the model; confirm it is populated and expose it to admins.

## Worth deciding separately

Whether exhausting the budget should be terminal at all. A workspace with no
usable embedding credential burns all 3 retries in ~3 minutes and then stays
broken forever, even after the credential is fixed — the failure is not
transient, so retry-with-backoff is the wrong model for it. Recovery currently
requires an admin call that a self-serve user cannot make.

JFK:
Closes #1496.

## What was wrong

467 memories on production sat at `embedding_status='failed'` — **every one at
exactly `retry_count = 3 = MAX_EMBEDDING_RETRIES`**. The sweep only claims a
failed row while `retry_count < MAX`, so nothing would ever touch them again.

A failed embedding never reaches Qdrant, and the BM25 sparse vectors live there
too, so those rows were missing from recall in **both semantic and keyword
mode** — while still being counted by the dashboard and charged against quota.
One workspace: 418 unsearchable against 269 searchable. Nothing anywhere said
so; the only way to find out was to query the database, which is how it was
found.

## Why they were stuck

The retry budget is terminal by design, and that is right for the class it was
written for: poison rows — bad input, oversized text — that more attempts cannot
help.

"No embedding credential is configured" was landing in the same bucket. It is
not that class. It is a fixable state, and it burned all three attempts in about
three minutes; adding the key afterwards changed nothing, because the rows were
already at the ceiling.

## The fix

`is_configuration_failure` refunds the budget for `ConfigurationError` and
`EmbeddingSpendCapExceeded`. Those two are safe to branch on because
`EmbeddingService` re-raises them **unwrapped**, above its blanket
`except Exception -> OpenAIError`, and both are raised **before** the provider
call — so re-probing one costs no tokens and no HTTP request. Two short queries
per sweep tick, globally bounded by the existing `limit(20)`.

**Why a counter reset and not a new `blocked` status.** A status needs a
migration, an enum-drift pin, and an audit of every `== 'failed'` predicate —
and its exit would be event-driven, hooked onto key-create and plan-change.
That structurally misses the two repair paths with **no HTTP handler at all**:
setting the platform `OPENAI_API_KEY`, and a plan change applied out of band.
The counter reset keys on the *failure*, not the *fix*, so the sweep's re-probe
**is** the condition check and every repair path is covered for free. It also
degrades safely under misclassification, which the alternative does not.

## The operator signal

There is no metrics pipeline in this deployment — no `/metrics`, no otel, no
scraper — so the signal goes to structlog:

- `embedding_budget_exhausted` (error) fires **once** when a row enters the
  never-retried-again state. `embedding_failed` drops to warning. The distinct
  name is the point: greppable, alertable, and its absence means something.
- Both carry `error_class`, which turns "read 467 messages" into one grep.
- The sweep reports `embedding_unsearchable_backlog` when non-zero, **before**
  the candidate query — a deployment whose failed rows have all gone terminal
  has nothing to sweep and would otherwise stay completely silent.

## What the user sees

A subtext on the Total Memories card, plus an informational Alert above it.
That card is where it belongs because that is the number that misleads without
it — every figure the user could see agreed with every other one, and all of
them counted rows rather than searchability.

`Alert`, not `ErrorBanner` and not a toast: nothing the user did failed, and the
house rule carves warnings out of the error channels. The copy leads with the
safe fact — saved, not lost, searchable as soon as indexing succeeds. Support is
mentioned only in the `stalled` string, where after this fix it is genuinely
true that automatic recovery will not help.

## Also in here

**A privacy leak, found while designing this and fixed separately in
`3c0a1b3e`.** `GET /workspace/embedding-status` returns up to 50 failed
memories *with their summaries*, scoped by `workspace_id` only — no
accessible-context filter, while every other stats path in the same module has
one. Any workspace member could read the opening 200 characters of another
member's **private** context memories by asking for the embedding queue. It
needs a memory whose embedding actually failed, which is unusual outside exactly
the situation this issue is about — so the leak got materially more reachable at
the same moment.

**A one-time backfill migration.** The code fix is not retroactive; rows already
at the ceiling can never be claimed. Without it this ships pointing at a dead
end and the new counts would name a problem with no way out. It resets the
counter rather than pattern-matching the free-text `embedding_error`, which puts
the rows back through the *fixed* path and lets the corrected logic decide:
configuration failures self-heal, genuine poison rows burn four fresh attempts
and re-stick. Self-limiting.

## Verification

| Gate | Result |
|---|---|
| backend pytest | 1966 passed |
| ruff | clean |
| pyright | 0 errors |
| frontend tsc | 0 errors |
| frontend vitest | no new failures |

Every fix is pinned by a test **verified by mutation**. The decision that
matters lives in `embedding_failure_values` so it is testable without a live
session and a Qdrant client; a narrow AST pin covers the seam those tests cannot
see, and removing either half fails the suite.

## Known gaps, deliberately not fixed here

- **A revoked or wrong key still goes terminal.** It arrives as a plain
  `OpenAIError` because the blanket wrapper erases the provider SDK's type.
  Catching it means reading `openai` status codes — SDK internals this project
  pins no upper bound on. Worth its own issue; it is the second-most-likely
  cause of a stuck workspace after "no key at all".
- `cryptography.fernet.InvalidToken` (wrong `ENCRYPTION_KEY`) has the same
  shape: an operator configuration error reported as a provider error.
- **Steady-state churn is real but small.** A deployment that never fixes its
  credential re-probes those rows forever — bounded at ~40 claims/min globally,
  zero provider calls. If a backlog ever reached six figures the right answer is
  a per-workspace circuit breaker, not a terminal state.
- Separately filed: #1497, where `merge_contexts` copies only `success` rows and
  then deletes the source holding the rest. That one loses data rather than
  hiding it.


JFK:
Two review comments — one landed, one I think is mistaken.

**The attempt count: correct, fixed in 6f081c54.** Eligibility is
`retry_count < MAX_EMBEDDING_RETRIES`, so from 0 a row is claimed at 0, 1 and 2
— three attempts, not four. Corrected in both places in the migration docstring
and pinned by a test so the prose and the constant cannot drift again. It is a
small error but it is the same shape this PR keeps tripping over: text beside
code asserting something the code does not do. Thanks for catching it.

**The `isinstance` union: this one is wrong.** `isinstance(x, A | B)` has been
valid since Python 3.10 — PEP 604 made `types.UnionType` acceptable as the
second argument. This project runs 3.11:

```
$ python -c "
class A(Exception): pass
class B(Exception): pass
print(isinstance(A(), A|B), isinstance(ValueError(), A|B))"
True False
```

It is also exercised directly by the suite in this PR —
`TestWhichFailuresAreConfiguration` calls `is_configuration_failure` with real
`ConfigurationError` and `EmbeddingSpendCapExceeded` instances and asserts both
the True and False branches. If it raised `TypeError` as described, those six
parametrised cases would fail rather than pass, and `backend-unit` is green.

Leaving it as the union form: it reads better than a tuple and matches the type
annotation style used elsewhere in this module. Happy to switch if there is a
project convention I have missed, but not for the stated reason.

PR #1499, opened 2026-08-05
fix(api): merge_contexts left unembedded memories behind, then deleted them (#1497)

### LINKED ISSUE #1497: fix(api): merge_contexts silently drops memories whose embedding failed, then deletes the source holding them

JFK:
Closes #1497. Found while designing #1496 — filed separately because that issue
was about memories being *invisible*, and this one is about *losing* them.

## The bug

`merge_contexts` selected only `embedding_status == "success"` rows, copied
those into the target, and then — with `delete_source=True` — soft-deleted the
source anyway. Anything `pending` or `failed` was neither copied nor kept. A
separate early-return path honoured `delete_source` having copied nothing at
all.

**Live, not theoretical.** The production deployment in #1496 has:

| context | success | failed |
|---|---|---|
| `kagura-desk` | 269 | 383 |
| `wemof-trading` | **0** | **16** |

Merging `wemof-trading` would have copied nothing, reported `merged: 0`, and
deleted the context holding all 16 — against a UI that had just displayed 16.
And #1496 established that `failed` is **recoverable**: the sweep re-embeds
those rows as soon as a credential exists. This was discarding data that was
coming back on its own.

## Why widening the filter would have been worse than the bug

`copy_context_points` needs a vector, and an unembedded row has no Qdrant point,
so the filter was protecting something real. But it protected the *vector* copy;
it never justified excluding the *row*.

The trap: the copy loop hardcoded `embedding_status="success"` on every new row.
Widening the `WHERE` alone would have minted rows claiming to be embedded with
no vector — invisible to search **and** to the sweep, which only claims
pending/processing/failed. Nothing would ever have repaired them. That is a
fresh #1496, manufactured by the merge. The two changes only make sense
together.

## Changes

- Every live row is selected. Unembedded ones copy as `pending` with a fresh
  retry budget, so the #1496 sweep embeds them into their **new** home
  (`process_pending_embedding` resolves the collection from `context_id`).
  `pending` rather than the source's status verbatim — a row that had exhausted
  its budget would otherwise be born terminal and never retried.
- `copy_context_points` returns the ids it upserted instead of a count, and any
  row whose vector did not land is downgraded to `pending` rather than left
  claiming success.
- `merged` counts **rows transferred**. It used to count Qdrant points.
  `pending_embedding` names what is not searchable yet.
- `delete_source` is gated on every row having transferred.
- `is_default` is pre-checked beside `is_locked` — `delete_context` refuses a
  default context, but only *after* rows and vectors were copied, so the merge
  half-completed and then raised.

### Provenance columns the copy was dropping

`source_type` is the serious one. It defaults to `"manual"`, and connector-origin
rows are excluded from the pinned/bootstrap lane precisely because they are not
manual (OWASP LLM01/LLM03, `repositories/memory.py`). A merge promoted connector
content into a lane built to keep it out — and merging into a `trusted` context
cleared the other half of that gate at the same moment. `source_uri` and
`delivery_mode` were dropped alongside it.

`reference_count` is now reset explicitly next to `access_count`; carrying one
while zeroing the other breaks the documented `access_count >= reference_count`
invariant.

## On the tests

**The first version of this regression suite did not catch the bug.** The
harness mocks `db.execute`, so the WHERE clause never runs — restoring the
original `success`-only filter left all nine tests green while the data loss was
fully back. Found by mutation testing, fixed by asserting on the emitted SQL.

Mutation-verified: the selection filter, `source_type` preservation, the
unvectored downgrade, and the copy-loop field set each fail a specific test when
reverted.

**Correction (91b121b4).** An earlier revision of this PR claimed the
`delete_source` gate protected against a future re-narrowing of the selection,
and that it could not be tested. Both were wrong: it compared the copy against
that same selection, so narrowing shrank both sides and the guard passed while
rows were left behind. It now counts live rows independently from the table, and
that version *is* testable — three tests cover it, one mutation-verified against
the old comparison. Caught in review.

## Not addressed here

The frontend merge dialog and its copy. This PR makes the API safe; whether the
UI should warn about deferred indexing is a separate question, and I did not
want to hold the data-loss fix behind it.

### DIFF
--- a/backend/src/db/qdrant.py
+++ b/backend/src/db/qdrant.py
@@ -1046,11 +1046,17 @@ async def copy_context_points(
     memory_id_mapping: dict[str, str],
     collection_name: str = KAGURA_MEMORIES_COLLECTION,
     batch_size: int = 100,
-) -> int:
+) -> set[str]:
     """Copy Qdrant points from source context to target context (Issue #90).
 
     Retrieves points by ID, creates new points with updated context_id and new IDs.
     Processes in batches for memory efficiency.
+
+    #1497: returns the NEW ids actually upserted, not a count. Two branches here
+    skip a memory silently — a source point that does not exist, and one stored
+    without a vector — and a bare count cannot say WHICH. The caller needs that
+    to avoid marking a target row ``success`` when nothing was copied for it,
+    which would manufacture a row no automatic path can ever repair.
 
     Args:
         workspace_id: Workspace ID
@@ -1061,7 +1067,9 @@ async def copy_context_points(
         batch_size: Points per batch
 
     Returns:
-        Number of points copied
+        The set of new memory ids whose vector was upserted. A caller comparing
+        this against ``memory_id_mapping.values()`` learns exactly which rows
+        have no vector in the target.
 
     Raises:
         QdrantError: If copy fails
@@ -1072,7 +1080,7 @@ async def copy_context_points(
         )
 
     client = get_qdrant_client()
-    copied = 0
+    upserted: set[str] = set()
 
     try:
         old_ids = list(memory_id_mapping.keys())
@@ -1124,13 +1132,13 @@ async def copy_context_points(
                     collection_name=collection_name,
                     points=new_points,
                 )
-                copied += len(new_points)
+                upserted.update(str(p.id) for p in new_points)
 
             logger.debug(
                 "context_points_copy_batch",
                 batch=i // batch_size + 1,
                 copied_in_batch=len(new_points),
-                total_copied=copied,
+                total_copied=len(upserted),
             )
 
         logger.info(
@@ -1139,10 +1147,10 @@ async def copy_context_points(
             workspace_id=workspace_id,
             source_context_id=source_context_id,
             target_context_id=target_context_id,
-            count=copied,
+            count=len(upserted),
         )
 
-        return copied
+        return upserted
 
     except Exception as e:
         logger.error(
--- a/backend/src/services/context_service.py
+++ b/backend/src/services/context_service.py
@@ -870,6 +870,14 @@ class ContextService:
                 "Source context is locked. Unlock it first via update_context(is_locked=false)."
             )
 
+        # #1497: delete_context refuses a default context — but it is called AFTER
+        # the rows and vectors have been copied, so the merge half-completed and
+        # then raised. Check it up front, with the other pre-flights.
+        if delete_source and source.is_default:
+            raise ValidationError(
+                "Cannot delete the default context. Merge with delete_source=false."
+            )
+
         # Validate same embedding model (single query for both configs)
         settings = get_settings()
         cfg_result = await self.db.execute(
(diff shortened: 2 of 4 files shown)


JFK:
Good catch — you're right, and it invalidated a claim I made in the PR body.

The guard read `len(rows_by_new_id) != len(source_memories)`, which is a
tautology: narrow the SELECT and both sides shrink together. I wrote that it was
there "so a future re-narrowing refuses the delete instead of quietly
resurrecting this" — it would have done no such thing. It guarded against
exactly the regression it could not see.

Fixed in 91b121b4 by counting live rows straight from the table
(`context_id` + `deleted_at` only), independent of the selection, with both
numbers in the error message.

Your comment also removed a limitation I had stated. I said the gate was "true
by construction, so there is no way to make it fire without also breaking the
selection" and left it deliberately untested. That was only true of the broken
version. The independent count is testable, so it now is — a source holding 16
live rows with 1 selected raises, and reverting to the old comparison fails that
test. Two more pin the other direction: a complete transfer still deletes, and a
row copied as `pending` counts as transferred (gating on vectors rather than
rows would refuse deletes for merges that lost nothing).

PR body updated to match.

Issue #1030, opened 2026-06-16
feat(api): platform-managed embeddings for paid plans (M/L) instead of BYOK


JFK:
## Summary

Paid plans (M = `basic`, L = `pro`) should use **platform-managed embeddings** instead of requiring BYOK (bring-your-own-key). Today every workspace must supply its own embedding API key (or rely on a dev-only env fallback) for memories to become searchable. For paying users this is friction; we want "you pay, we handle embeddings."

The plumbing already exists — this is mostly connecting existing pieces plus one **mandatory** cost-safety addition.

## Current state (verified)

- **Key resolution hierarchy**: `_get_user_api_key()` resolves `context key → workspace key → env(platform) fallback` and tags the source via a `KeySource` enum (`byok` / `platform`) — `backend/src/services/embedding_service.py:95-231`.
- **Cost attribution**: `paid_by: "byok" | "platform"` is already recorded on `LLMCallLog` — `embedding_service.py:315`.
- **Plan tiers**: `free` (S) / `basic` (M) / `pro` (L) with `FEATURE_MIN_PLANS` gating and per-tier `embedding_daily_cap_usd` / `embedding_monthly_cap_usd` (#709) — `backend/src/config/plan_tiers.py`.
- **Plan lookup**: `workspace._plan_tier` — `backend/src/models/auth.py:1434-1438`.

## ⚠ Mandatory requirement — extend the spend cap to the platform path

The embedding spend cap (#709) is **currently BYOK-scoped only** — it deliberately does **not** apply to the env/platform path (`embedding_service.py:249`). Once paid plans embed on **our** key, **we** bear the cost. Without extending the cap to the platform path, a paid workspace's platform embedding cost is **unbounded**. This is not optional:

- [ ] Extend the spend-cap gate to apply to platform-managed embeddings for `basic`/`pro` workspaces (reuse the per-tier cap values; semantics shift from "BYOK drain guard" → "platform budget ceiling").

## Implementation outline

- [ ] Add a plan capability (e.g. `FEATURE_MIN_PLANS["managed_embeddings"] = "basic"`, or a `managed_embeddings` flag on the `PlanTier` dataclass).
- [ ] In `_get_user_api_key()` / `_get_client()`: for `basic`/`pro` workspaces, route to the platform-managed key (skip or de-prioritize BYOK per the decision below).
[...]

PR #1501, opened 2026-08-06
fix(frontend): warn when embedding is unavailable, not when a key list is empty (#1495)

### LINKED ISSUE #1495: fix(frontend): the external-keys warning counts key rows instead of asking whether embedding works
## What

`/workspace/integrations/external-keys` renders a destructive alert —
*"An OpenAI API key is required — memory storage and search require an OpenAI API
key for embedding generation"* — plus an amber ⚠ on the sidebar nav entry,
whenever the workspace has zero OpenAI key rows.

That is the **only** input. `external-keys/page.tsx:184`:

```ts
const hasOpenAIKey = keys.some((k) => k.provider === "openai" && k.enabled);
```

and `Sidebar.tsx:397`:

```ts
listExternalAPIKeys().then((keys) => setHasExternalKeys(keys.length > 0));
```

Neither asks the server whether embedding is actually available.

## Why it is wrong

The server resolves an embedding credential through a chain
(`embedding_service.py:113-223`):

1. context-scoped key → 2. workspace-scoped key → 3. platform `OPENAI_API_KEY`
   → 4. `ConfigurationError`

Step 3 is invisible to the UI. A deployment that sets the platform key — which
is the documented deployment path, and what the `managed_embeddings` feature on
the Pro tier (`plan_tiers.py:190`) exists to describe — serves every workspace
correctly while this banner tells all of them they are broken and must go buy an
API key.

This is the same shape as #1487: the client re-implements a server rule it
cannot see, and the user is told a false reason. #1487 was reported to us as a
plan/quota bug; this one was reported as *"I thought memory-cloud didn't need
external API configuration."*

## The signal already exists

`GET /api/v1/system/embedding/models` (`system.py:485`) computes
`openai_available` from the workspace key **and then falls back to the env var**:

```python
if not openai_available:
    openai_available = bool(os.getenv("OPENAI_API_KEY"))
```

That is the question the banner should be asking. Note `/system/info`'s public
`features` block has no embedding-availability flag, so the authenticated
endpoint above is the correct source — not a new public one.

## Acceptance criteria

- [ ] The alert and the sidebar ⚠ are driven by a server-reported capability,
      not by a count of key rows.
- [ ] With the platform key configured and zero workspace keys, neither appears.
- [ ] With no credential available anywhere, both still appear — this warning is
      genuinely correct in that case and must not be weakened into silence.
- [ ] The copy distinguishes "embedding is unavailable" from "you have not added
      a key", since with managed embeddings the second does not imply the first.
- [ ] A test pins the platform-key-configured case; the current suite cannot
      fail on this because the condition never consults the server.

## Note

On the current production deployment the banner happens to be **correct** — the
platform `OPENAI_API_KEY` is not set there, so BYOK really is required and 466
memories across two keyless workspaces are sitting at `embedding_status=failed`.
That is a separate deployment-configuration problem, tracked apart from this
issue. Fixing the banner must not hide it: with no credential anywhere the
warning has to keep firing.

JFK:
Closes #1495.

## The bug

The external-keys page showed a red **"OpenAI API key required"** alert, and the
sidebar an amber ⚠, whenever the workspace held zero OpenAI key rows:

```ts
const hasOpenAIKey = keys.some((k) => k.provider === "openai" && k.enabled);
```

That count was the entire input. Neither surface asked the server whether
embedding actually worked — and the server does not decide it that way. The
credential chain ends at a platform `OPENAI_API_KEY`, so a workspace with no key
of its own is served fine wherever the deployment sets one.

**Live in production.** After the platform credential was provisioned, every
active workspace embedded at a 0% failure rate while being told it was broken
and had to go buy an API key. Reported from two separate accounts.

| workspace | own key | success | failed |
|---|---|---|---|
| `0e7a406d` | no | 48 | 0 |
| `4081874b` | no | 695 | 0 |
| `61b5cf6f` | yes | 6,851 | 0 |
| `d0687ad6` | yes | 1 | 0 |

Same shape as #1487: a client re-deriving a server rule it cannot see. The
endpoint's own comment had already recorded the hazard — *"would report
has_key=false in deployments where env keys serve embeddings"* — and #1167
handled only the `ENABLE_BYOK=false` half of it.

## The fix

`OpenAIKeyStatusResponse` gains **`embedding_available`**. `has_key` keeps its
meaning ("does this workspace own a key"), because the page still needs that
separate and real question. Three surfaces move to the new field: the sidebar
triangle, the red alert, and the context-creation gate.

**The plan gate is not re-implemented to compute it.** #1030 lets a deployment
restrict the platform fallback to plans carrying `managed_embeddings`, and a
second copy of that condition would drift from the one the embedding path
enforces — rebuilding this defect one layer down. Both callers route through
`platform_fallback_allowed()`, and a test fails if either grows its own copy.

New copy for the case that previously had none: a workspace with no key of its
own, embedding happily on the platform credential, now says so instead of
showing an empty list under a red warning.

## Verification

| Gate | Result |
|---|---|
| backend pytest (services) | 1,889 passed |
| frontend vitest (affected) | 93 passed |
| ruff check / format | clean |
| pyright | 0 errors |
| tsc | clean |

Three mutations verified: dropping the plan gate from the endpoint, giving the
endpoint a private copy of the rule, and making the helper ignore the setting
each fail a specific test.

## Two tests were wrong first

Worth recording, because it happened twice in one change:

- one searched the source **text** for the helper's name and passed against code
  that had dropped the CALL but kept the now-unused import — the exact
  regression it existed to catch;
- another failed against **correct** code, because the comment explaining the
  rule contained the string it was grepping for.

Both now read the AST. A substring in source is not a measurement of behaviour.

## Deployment note

This lands after `EMBEDDING_PLATFORM_FALLBACK_REQUIRES_MANAGED_PLAN=true` was
enabled in production (#1030's intended managed-SaaS configuration, verified
in-process across all 14 workspaces: 7 free DENIED, 7 pro ALLOWED). Without this
PR the Free workspaces would be shown "embedding is handled for you" while the
server refuses them — the same lie, pointing the other way.


JFK:
Correct — fixed in c921cd3c.

`/openai-key-status` is behind `require_byok_enabled` and 404s with BYOK off, so
the probe I added to the external-keys page was a guaranteed-failing request on
every render of that page. Sidebar and the contexts page already guarded it;
this was a third caller copied from neither, and I introduced the inconsistency
in this same PR.

Rather than just fixing the one file, I pinned the property across all three
callers (`byok-probe-guard.test.ts`). The failure mode here is silent — the UI
swallows the 404 and shows no warning, which is indistinguishable from a healthy
workspace — so nothing would have flagged a fourth caller repeating it. Verified
by mutation: removing the guard fails that test.

The guard also covers `systemFeatures === null` (still loading), where the
answer simply is not known yet.

PR #1506, opened 2026-08-09
feat(mcp): surface persistence expectations for scope=working writes (#1505)

### LINKED ISSUE #1505: feat(mcp): surface persistence expectations for scope=working writes
## Problem

`remember()` with high importance returns `scope: "working"`, and nothing in the response or docstring tells the writing agent what that implies for durability. Dogfooding moment (2026-08-08): a memory written with `importance: 0.95` (a key experiment result) came back `scope: "working"` — causing genuine "will this survive until sleep consolidation promotes it?" uncertainty mid-session. Meanwhile `delivery_mode="always"` pins to persistent immediately, an asymmetry that makes the normal-write behavior feel riskier than it (presumably) is.

This is primarily a transparency/UX gap, not (necessarily) a behavior bug.

## Proposal

One or more of:

1. **Response transparency**: `"persistence": {"scope": "working", "promotes_via": "sleep_consolidation", "eviction_risk": "none_before_next_run"}` (or equivalent wording) in the `remember`/`update_memory` response.
2. **Docstring**: state the working→persistent lifecycle and what, if anything, can be lost before promotion.
3. Optional: auto-persist above an importance threshold (e.g. ≥0.9), mirroring the `delivery_mode="always"` pin-to-persistent behavior.

(1)+(2) alone would fully resolve the dogfooding anxiety.

_Origin: dogfooding feedback — Claude Code session on wemof-agent (2026-08-08/09)._

JFK:
Closes #1505

## Problem

`remember()` returned `scope: "working"` and nothing else about what that means. Next to `delivery_mode="always"` — which pins straight to persistent — a high-importance write reads as provisional, and the caller had no way to answer "does this survive until consolidation promotes it?"

## What this adds

A `persistence` block on the `remember()` / `update_memory()` responses, plus the lifecycle in the tool descriptions. Every statement is read off the code that enforces it:

- **`committed`** — the row is committed to PostgreSQL before the call returns. `scope` does not gate durability; it selects a consolidation lifecycle.
- **`promotes_via`** — which consolidation pass *this deployment* runs, resolved from the same flags the task modules read. The three configurations are genuinely different, and one runs nothing:

  | `SLEEP_ENABLED` | `ENABLE_NEURAL_MEMORY` | pass |
  |---|---|---|
  | true | true | `sleep_consolidation` |
  | true | false | **neither** — `sleep_maintenance_task` returns at its neural guard, and `schedule_neural_tasks` skips the legacy cron whenever sleep is on |
  | false | — | `legacy_consolidation` |

  `promotes_via` is `null` in the middle case rather than naming a pass that never fires. Promotion criteria differ too: the legacy pass gates on `access_count` (surfacing) and has no centrality rule, so the advertised criteria follow the running pass.
- **`consolidation_archive_min_age_days`** — the running pass's own archival floor.

## What it deliberately does not claim

Not a retention SLA. An earlier revision of this PR said "nothing you write today can be dropped out from under you" — **that was false.** `DedupMergePhase._fetch_active_memories` selects on user/workspace/context and `deleted_at IS NULL` only: no age gate, no scope gate, no adoption gate. A memory written minutes ago can lose a ≥0.98-cosine merge the same night (tags and edges transfer to the winner; its id stops resolving), and that applies to persistent-scope memories too.

So the age field is **named** for its scope rather than relying on prose, and both `detail` strings name the lifecycles the floor does not bind (near-duplicate merge, `forget()`).

## Anti-drift

- The advertised floor is bound to the sleep path's own `_archival_eligible` predicate, asserted at `days-1` (ineligible) and `days` (eligible) — retuning either side without the other fails.
- The legacy path's literal `30` is now the named `LEGACY_ARCHIVE_MIN_AGE_DAYS`, with an AST test that `consolidation_task` still compares against it (so the constant can't become decorative).
- Two AST tests pin the dedup selection columns. If an age or scope gate is ever added there, they fail — prompting the wording to be *widened*, rather than staying silently over-conservative.

## Testing

24 tests. All three configurations, both scopes, the service wiring (real `MemoryService.remember()` with a mocked DB), and the MCP JSON both handlers actually emit. Mutation-checked: dropping the field at a construction site, freezing the env read behind a cache, hardcoding the floor, dropping the scoping clause, and reintroducing a bare issue ID are each caught.

One test asserts the agent-facing schema text carries no bare issue IDs — a Pydantic model docstring and `Field(description=)` ship to OpenAPI, so the `#NNNN` convention applies there too.

## Not implemented

Proposal (3) from the issue, auto-persist above an importance threshold. The issue notes (1)+(2) resolve the gap; (3) changes consolidation behaviour and deserves its own decision.

### DIFF
--- a/backend/src/api/routes/memory.py
+++ b/backend/src/api/routes/memory.py
@@ -97,8 +97,21 @@ async def remember(
         {
             "status": "success",
             "memory_id": "uuid",
-            "scope": "working"
+            "scope": "working",
+            "persistence": {
+                "scope": "working",
+                "committed": true,
+                "promotes_via": "sleep_consolidation",
+                "consolidation_archive_min_age_days": 30,
+                "detail": "..."
+            }
         }
+
+    The memory is committed before this returns; ``scope`` selects the
+    consolidation lifecycle, not whether it was stored. ``persistence``
+    describes that lifecycle for this deployment and is null when the scope
+    cannot be classified. Its age floor is scoped to consolidation and is not a
+    retention guarantee — see ``services.persistence``.
 
     Example:
         POST /api/v1/memory/remember
--- a/backend/src/mcp_server/tools/_definitions.py
+++ b/backend/src/mcp_server/tools/_definitions.py
@@ -141,9 +141,26 @@ Multi-context management (personal/work/contexts in one collection):
 
 The more semantic metadata you provide, the better the search relevance.
 
+DURABILITY — what scope="working" means:
+A new memory is committed to the database BEFORE this call returns. It is saved;
+you do not need to re-write it or wait for anything. `scope` selects which
+consolidation lifecycle the memory is on, NOT whether it was stored:
+• "working" (the default for a normal write) — a nightly consolidation pass can
+  promote it to "persistent"; `persistence.promotes_via` names the pass this
+  server actually runs (null if none is enabled). That pass will not archive a
+  working memory before `persistence.consolidation_archive_min_age_days` old,
+  and only if it has never been adopted.
+• "persistent" — outside consolidation's reach. delivery_mode="always" pins
+  straight here on write; that is a delivery guarantee, not a stronger
+  durability guarantee than a working-scope write already has.
+That age floor is scoped to consolidation and is NOT a retention SLA: separate
+near-duplicate merge maintenance can retire a memory at any age (its tags and
+edges move to the memory it merged into), and forget() removes one on demand.
+The response carries a `persistence` block for the scope you actually got back.
+
 IMPORTANT: Always specify context_id to ensure you're using the intended context. Use list_contexts() to discover available context IDs.
 
-Returns: {status, memory_id, scope, context_id, context_name, context_display_name, context_is_private, context_is_locked}. NOTE: the embedding is generated asynchronously AFTER this returns, so the new memory is not findable via recall() for a brief moment.""",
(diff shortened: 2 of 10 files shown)


JFK:
## Review follow-up: a live escape, now fixed

A multi-perspective review of this diff produced 23 findings; 22 were refuted on inspection, and **one was confirmed and was real**:

**Both `UpdateMemoryResponse` construction sites were built without the `persistence` kwarg.** `update_memory` therefore never emitted the block, while `models/schemas.py` still declared the field, the handler still splatted it, and the tool description still advertised `persistence?: {...}`. An advertised contract was unmet.

The whole suite stayed green through it. The update-path test replaced the service wholesale:

```python
service.update_memory = AsyncMock(
    return_value=SimpleNamespace(..., persistence=persistence_info("persistent"))
)
```

— hand-feeding the value it then asserted on. It exercised the handler's dict splat and never the service. This module's docstring claimed "a construction site that forgets the field fails here rather than shipping an invisible feature"; that was only ever true of `remember()`.

Fixed in f2af3f57:
- populate `persistence` at both sites
- drive `_update_in_place` and `_upsert_by_external_id` for real, so the field's absence fails at the service instead of being papered over by a mock
- an AST guard over every `RememberResponse`/`UpdateMemoryResponse` construction, so a new site added later cannot omit it either

Deleting both kwargs now fails 3 tests; before, it passed 2340.

Worth recording separately: mutation testing had already validated 4 mutations against this feature and did not surface this. Mutation testing measures whether the tests you wrote bite — not whether you covered the right paths.

Issue #1503, opened 2026-08-08
feat(api): tolerant tag matching in recall filters (drift-resistant reads)


JFK:
## Problem

`recall` tag filters are exact-match only. Writer-side tag drift (`dev-env` vs `dev-environment`, `troubleshoot` vs `troubleshooting`) silently empties or thins results — the reader gets few/no hits with no indication that near-miss tags exist.

Dogfooding example (2026-08-08): a bootstrap recall with `filters={"type": "troubleshooting", "tags": ["dev-environment"]}` returned 2 results with `confidence: low`. Any past memory tagged with a drifted spelling is invisible to this query, and the caller cannot tell "nothing exists" apart from "spelling mismatch".

## Proposal

Read-time tolerance, opt-in to preserve exact semantics:

- `tags_match: "normalized"` — case/hyphen/underscore/plural normalization before matching, or
- fuzzy resolution against the `list_tags` vocabulary (edit distance ≤ 2), or
- minimal version: when a tag filter matches 0 memories but a near-duplicate tag exists, include a hint in the response: `"tag_suggestions": {"dev-environment": ["dev-env (3)"]}`.

## References
- #614 `list_tags` already computes near-duplicate spellings (surfacing half)
- #746 hygiene skill (bulk rename, remediation half) — this issue is the read-time third leg

_Origin: dogfooding feedback — Claude Code session on wemof-agent (2026-08-08/09)._

PR #1508, opened 2026-08-09
feat(mcp): write-time recall-ability lint in remember() response (#1502)

### LINKED ISSUE #1502: feat(mcp): write-time recall-ability lint in remember() response
## Problem

Recall quality silently depends on how carefully the *writer* crafted `summary`/`tags` at `remember()` time. During a heavy dogfooding session (wemof-agent development, 2026-08-08/09, ~10 writes + 8 recalls), every write required deliberate effort — conclusion-first summary, synonym inclusion, tag-convention discipline — and there is no feedback loop: a sloppy write degrades future recall **silently**, discovered only when a later session gets low-confidence results.

The docstring teaches good summary style (✅/❌ examples), but nothing checks compliance at write time.

## Proposal

Optional, non-blocking write-time lint in the `remember()` (and upsert `update_memory()`) response, e.g.:

```json
"lint": [
  {"code": "summary_narrative", "hint": "summary reads as narrative; state the reusable conclusion"},
  {"code": "tag_near_duplicate", "hint": "tag 'dev-env' is near-duplicate of existing 'dev-environment' (12 uses)"},
  {"code": "summary_no_domain_terms", "hint": "no searchable domain terms; consider synonyms"}
]
```

Cheap server-side heuristics are enough (length/style check, existing-tag edit distance via the `list_tags` index). Could be opt-in (`lint: true`) or always-on since it is advisory only.

## References
- #746 tag & type hygiene (client-side, after-the-fact remediation — this is write-time prevention)
- #614 `list_tags` near-duplicate surfacing (the data source this lint needs)

_Origin: dogfooding feedback — Claude Code session on wemof-agent, heavy MCP usage (recall/remember/update_memory/load_pinned/recall_upcoming)._

JFK:
Closes #1502

> **Stacked on [#1507](https://github.com/kagura-ai/memory-cloud/pull/1507) (#1503).** Base is `1503-feat/tolerant-tag-matching`, not `main` — the `tag_near_duplicate` rule reuses that PR's tag-vocabulary read and near-duplicate relation rather than duplicating them. **Merge #1507 first**, then this retargets to `main` automatically.

## Problem

Recall quality silently depends on how carefully the *writer* crafted `summary` and `tags`. A sloppy write degrades future recall with no feedback at the time — it surfaces later as a low-confidence result in a different session. The `remember()` docstring teaches good style with explicit ✅/❌ examples, but nothing checked compliance.

## What this adds

An advisory `lint` array on the `remember()` / `update_memory()` responses:

| code | fires when |
|---|---|
| `summary_short` / `summary_long` | outside the thresholds the request schema already logs about |
| `summary_narrative` | the summary opens as a record of an event, not the reusable conclusion |
| `no_tags` | the memory is unreachable by any tag filter |
| `tag_near_duplicate` | a new tag resembles one already established here, with the dominant spelling and its count |

```json
"lint": [{"code": "tag_near_duplicate",
          "hint": "tag 'dev-env' is new but resembles 'dev-environment' (12 memories). Reusing the established spelling keeps tag filters working.",
          "subject": "dev-env"}]
```

The length thresholds are **imported** from the request schema (now named `SUMMARY_SHORT_THRESHOLD` / `SUMMARY_LONG_THRESHOLD`) rather than restated, so the advice a writer receives cannot drift from the guidance the schema already logs server-side.

## Three constraints shaped every rule

**Never blocking.** It runs *after* the commit and the audit row, and swallows its own errors. Nothing here can turn a successful write into a failure. A test asserts the *ordering* — that `_lint_write` appears after `db.commit()` — not merely that a `try/except` exists.

**Only mechanical, checkable claims.** No rule guesses at meaning. `summary_narrative` is anchored to the **opening** of the summary, because a conclusion-first summary may legitimately mention a meeting later in the sentence:

> "Retry budget must reset on configuration failures, not count against the cap — agreed in the platform meeting after the outage."

That must not fire, and there's a test for it.

**Silence is the common case.** A well-formed write returns no hints, and the MCP response carries no `lint` key at all — presence is itself the signal. Every rule is tested from both sides: an established tag isn't flagged, a genuinely new topic isn't flagged, a good summary produces nothing. A rule that fires on ordinary good writes would just teach agents to ignore the field.

## Cost

The vocabulary read is skipped entirely when there are no tags to compare against it, and otherwise reuses #1503's bounded, context-scoped aggregate.

## Testing

31 tests. Mutation-checked: dropping `lint` from the response construction sites, moving the lint before the commit, letting it raise instead of swallowing, and flagging every new tag rather than only near-duplicates are each caught.


JFK:
### CI note

`.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, so a PR based on another branch gets **no checks at all** — that is why this one shows none. CI will run automatically once #1507 merges and GitHub retargets this PR to `main`.

Verified locally in the meantime, on this branch:

- `pytest tests/services/ tests/mcp_server/ tests/utils/` → **2832 passed**, 298 skipped
- `ruff check` → clean · `ruff format --check` → clean · `pyright src/` → 0 errors

Merge order for the milestone: **#1507 → this**.

